PR #5, opened 2025-08-14
Correct namespace for WP\MCP\Core\McpAdapter

jonathanbossenger:
Corrects the examples in the README to point to the correct namespace for the McpAdapter class.


jonathanbossenger:
@galatanovidiu during testing I found a small bug in the readme examples.

galatanovidiu:
## Great catch! 👍 

Thanks for identifying and fixing this namespace issue, @jonathanbossenger! This is definitely a bug that would cause confusion for developers following the README examples.

This is an important correction since developers copying the examples from the README would encounter "Class not found" errors without this fix.

Thanks for the thorough testing and for catching this! 🚀

PR #8, opened 2025-08-17
refactor exploration: as a wordpress plugin [WIP]

justlevine:
## What

This **WIP** PR explores what this repo would look like as a WordPress Canonical instead of as a Composer library.

I'm also using it to explore and keep track of some of the questions I had during code review, unrelated to that primary change.

And I'll probably use this to cherry-pick some things back to core in the interim that don't conflict with the premise here or require a decision (like missing editorconfig, or the phpcs CI)

CI can be seen downstream at: https://github.com/justlevine/mcp-adapter/pull/1

### What's Included

- [x] Minimum PHP 7.4 
- [x] Canonical plugin w. no deps instead of Jetpack Autoloader
- [x] CI/CD, PHPStan, WPEnv, and other tooling configured.
- [x] Installation + Development Instructions 
- [x] Rector/PHPCBFed the `src`

### Todo
- [x] Add first batch of questions/comments on diff.
- [ ] Review the `src` results from the rector. (e.g. the named args should be an array with a closed shape).
- [ ] Review the failing lints/tests (for anything major)

### Not now

This only makes sense if this approach gets approved in which case, _all_ of these changes, not just the below should be done in a few separate PRs so they can be adequately reviewed.
- [ ] renaming `src` to `inc(ludes)` - didn't want to confuse the GitHub diff.
- [ ] using a local psr-4 (or wp-flavor classmap) autoloader) instead of composer.
- [ ] Reviewing the src `internals` - I might ask some questions as they come up in the comments, but i think "how we consume an API" needs to be answered before we can weigh in on what the contract/internal behavior of an API should be.


.wp-env.json
```diff
@@ -0,0 +1,20 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/wp-env.json",
+	"core": null,
+	"plugins": ["WordPress/abilities-api", "./."],
```
justlevine, line 4: Including trunk by default but we can version this if for some reason Abilities API isnt merged in time for core.

CONTRIBUTING.md
justlevine, line 1: I didn't bother to rewrite the `/docs` for this exploration. 
These steps are the dev docs.
They also have the benefit of being identical to those used in the Abilities API, so contributors dont need to juggle different project flows.

README.md
```diff
@@ -123,54 +123,77 @@ patterns for various use cases.
 ### Component Overview
 
 ```
-├── Core/                      # Core system components
-│   ├── McpAdapter.php        # Main registry and server management
-│   └── McpServer.php         # Individual server configuration
-├── Domain/                    # Business logic and MCP components
-│   ├── Tools/                # MCP Tools implementation
-│   │   ├── McpTool.php       # Base tool class
+./includes (@todo: currently src to keep the diff)
```
justlevine, line 126: https://github.com/WordPress/abilities-api/pull/3#discussion_r2263969976

leaves `src` for assets that need to be built that arent packages.

README.md
```diff
@@ -198,9 +221,8 @@ Individual server management with comprehensive configuration:
 
 ### Required Dependencies
 
-- **PHP**: >= 8.1
+- **PHP**: >= 7.4
```
justlevine, line 224: As a plugin and not a composer library, it makes sense not to go higher than we absolutely need to.
And since it's no longer a composer dependency, keeping at `7.4` doesn't get in the way of any plugins/themes that need to go higher either.

Also makes it easier for the eventual merge into WordPress core (assuming that hasn't changed as a long-term goal once this matures).

Cost: we need to use `switch` or `$map[ $key ]` instead of `match`.

README.md
```diff
@@ -212,45 +234,107 @@ This adapter requires the WordPress Abilities API, which provides:
...
 
-### Via Composer (Recommended)
+### From WordPress.org (Recommended)
 
-The preferred way to install the MCP Adapter is through Composer for enhanced dependency management:
+@todo coming soon
+
+### Via GitHub Releases
+
+The preferred way to install the MCP Adapter is by downloading the latest release from the [GitHub Releases page](https://github.com/WordPress/mcp-adapter/releases/latest).
```
justlevine, line 245: Only preferred until we're released on the plugin directory. Then, if a different plugin needs it, you'll be prompted to install it.

README.md
```diff
@@ -212,45 +234,107 @@ This adapter requires the WordPress Abilities API, which provides:
...
+
+```jsonc
+{
+  "repositories": [
+    {
+      "type": "vcs",
+      "url": "https://github.com/WordPress/mcp-adapter.git"
+    }
+    // ... other repositories.
+  ],
```
justlevine, line 265: Once released, this is just a link to wpackagist.org and wpackagist-plugin/mcp-adapter

README.md
```diff
@@ -212,45 +234,107 @@ This adapter requires the WordPress Abilities API, which provides:
...
+## Basic Usage
 
-**Composer Benefits:**
+### Requiring the plugin as a dependency
 
-- Automatic dependency resolution and updates
-- Version constraint management across your project
-- Integration with existing Composer-based workflows
-- Simplified dependency tracking in `composer.json`
+#### `Requires Plugins` Plugin Header (Recommended - coming soon)
```
justlevine, line 288: Needs release, but our use case is literally what this feature was built for.

README.md
```diff
@@ -212,45 +234,107 @@ This adapter requires the WordPress Abilities API, which provides:
...
-   }
-   ```
-3. Ensure the WordPress Abilities API is loaded before initializing the adapter
+While this is enough (@todo once on WordPress.org) to ensure the MCP Adapter is loaded before your plugin, if you need to ensure specific version requirements you can use the techniques described in the next section.
 
-### Example Implementation
+#### Checking availability with code
 
-For a complete working example of MCP Adapter integration, see the [MCP Adapter Implementation Example](https://github.com/galatanovidiu/mcp-adapter-implementation-example) - a WordPress plugin demonstrating best practices for implementing MCP servers with the adapter.
+If the above method is not suitable for your use case, you can manually check if the MCP Adapter is loaded in your plugin:
```
justlevine, line 308: If we're talking dx/adoption, more WordPress plugin/themese check for the existence of another plugin than use composer 🤷

README.md
```diff
@@ -212,45 +234,107 @@ This adapter requires the WordPress Abilities API, which provides:
...
+    );
+  } );
+  return;
+}
+```
+
+You can also check for specific versions of the MCP Adapter using the `WP_MCP_VERSION` constant:
+
+```php
+if ( ! defined( 'WP_MCP_VERSION' ) || version_compare( WP_MCP_VERSION, '1.0.0', '<' ) ) {
```
justlevine, line 326: The whole reason we're doing this outside of core is so we can ~move fast and break things~ iterate quickly without dragging the technical debt with us. 

jetpack autoloader doesn't solve this, it solves conflicts caused by making this a library and using Composer to version-manage it.

README.md
```diff
@@ -394,50 +478,25 @@ The MCP Adapter has been designed with enterprise production use in mind, suppor
...
 
-The MCP Adapter is designed as a **Composer package**, not a WordPress plugin, to provide maximum flexibility and
-integration capabilities. This architectural choice leverages
-the [Jetpack Autoloader](https://github.com/Automattic/jetpack-autoloader) to solve version conflicts and enable
-seamless integration across multiple WordPress projects.
+The MCP Adapter is designed as a **Canonical plugin**, not a Composer package, to provide maximum flexibility and
+integration capabilities enable seamless integration across multiple WordPress projects without external dependencies.
 
-### Package Benefits
+### Plugin Benefits
```
justlevine, line 486: I didnt mention that this makes it easier for us to get into core because im not sure if that's the goal anymore (otherwise I've get even _more_ questions about the choices in this repo), and my point here was more to contrast the library benefits . 
(Once the discussion is over, this section doesn't belong in the root readme)

composer.json
```diff
@@ -1,86 +1,96 @@
...
+	"repositories": [
+		{
+			"type": "composer",
+			"url": "https://wpackagist.org",
+			"only": [
+				"wpackagist-plugin/*",
+				"wpackagist-theme/*"
+			]
+		}
+	],
```
justlevine, line 31: Used for PCP, not in the build

composer.json
```diff
@@ -1,86 +1,96 @@
...
+		}
+	},
+	"config": {
+		"platform": {
+			"php": "7.4"
+		},
+		"sort-packages": true,
+		"optimize-autoloader": true,
+		"classmap-authoritative": false,
+		"apcu-autoloader": false,
```
justlevine, line 58: In a real merge candidate I'd pay more attention to detail.

```suggestion
```

composer.json
```diff
@@ -1,86 +1,96 @@
...
+		"wp-coding-standards/wpcs": "^3.1",
+		"wp-phpunit/wp-phpunit": "^6.5",
+		"wpackagist-plugin/plugin-check": "^1.6",
+		"yoast/phpunit-polyfills": "^4.0"
+	},
+	"scripts": {
+		"lint:php": "phpcs",
+		"lint:php:fix": "phpcbf",
+		"lint:php:stan": "phpstan analyse --memory-limit=1G",
+		"test": "phpunit --strict-coverage"
```
justlevine, line 94: In the Abilities API we just call them `lint`, `format`, and `analyse` since they only make sense in a php context, and then they're wrapped in the namespace for the `npm run *` call . Using the same names for both composer/npm like here probably makes more sense - or we don't care bc people should use npm 🤷

mcp-adapter.php
```diff
@@ -0,0 +1,57 @@
...
+	/**
+	 * Shortcut constant to the path of this file.
+	 */
+	define( 'WP_MCP_DIR', plugin_dir_path( __FILE__ ) );
+
+	/**
+	 * Version of the plugin.
+	 */
+	define( 'WP_MCP_VERSION', '0.0.1' );
+}
```
justlevine, line 44: yes yes I know I'm verbose, both in my comments and my code 😅. abilities-api doesnt bother to wrap these

package.json
```diff
@@ -0,0 +1,35 @@
...
+	],
+	"scripts": {
+		"plugin-zip": "wp-scripts plugin-zip",
+		"format": "wp-scripts format",
+		"lint:php": "wp-env run tests-cli --env-cwd=wp-content/plugins/$(basename \"$(pwd)\")/ composer run-script lint:php",
+		"lint:php:fix": "wp-env run tests-cli --env-cwd=wp-content/plugins/$(basename \"$(pwd)\")/ composer run-script lint:php:fix",
+		"lint:php:stan": "wp-env run tests-cli --env-cwd=wp-content/plugins/$(basename \"$(pwd)\")/ composer run-script lint:php:stan",
+		"test:php": "wp-env run tests-cli --env-cwd=wp-content/plugins/$(basename \"$(pwd)\")/ vendor/bin/phpunit -c phpunit.xml.dist",
+		"wp-env": "wp-env"
+	}
```
justlevine, line 34: fully self-contained linting/testing environment using wp-env 🚀

package.json
```diff
@@ -0,0 +1,35 @@
...
+		"node": ">=20.10.0",
+		"npm": ">=10.2.3"
+	},
+	"devDependencies": {
+		"@wordpress/env": "^10.28.0",
+		"@wordpress/scripts": "^30.21.0",
+		"@wordpress/prettier-config": "^4.28.0"
+	},
+	"files": [
+		"src",
```
justlevine, line 19: should be `includes` when `src` is `includes`

phpcs.xml.dist
justlevine, line 1: Stricter, WPCS-compatible, sourced from existing WordPress/* plugins before being reviewed/approved for Abilities API.

(can be toned tone to our needs, but a shared starting point)

phpcs.xml.dist
```diff
@@ -1,64 +1,208 @@
...
+<exclude name="WordPress.Files.FileName" />
+<!-- Allow modern PHP syntax -->
+<exclude name="Universal.Operators.DisallowShortTernary" />
+<!-- Needed to typehint, see: https://github.com/WordPress/WordPress-Coding-Standards/issues/403 -->
+<exclude name="Generic.Commenting.DocComment.MissingShort" />
+<!-- Allow internal hook names without prefixes -->
+<exclude name="WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound" />
+<!-- Allow unused parameters in interface implementations -->
+<exclude name="Generic.CodeAnalysis.UnusedFunctionParameter" />
+</rule>
```
justlevine: glitch in the matrix. maybe I'll add husky, but most people (myself too) hate it?

phpunit.xml.dist
justlevine, line 1: if something in my formatting didnt accidently squash these XMLs, the highlight here is that coverage works with the same `npm run lint:php` command, you just need to start `wp-env` with XDebug first.

bin/install-wp-tests.sh
justlevine, line 1: no boilerplate because we use wp-env

phpstan.neon.dist
```diff
@@ -0,0 +1,39 @@
...
+		phpVersion:
+			min: 70400
+			max: 80400
+		bootstrapFiles:
+				- mcp-adapter.php
+		paths:
+				- mcp-adapter.php
+				- src/
+		scanDirectories:
+			- ../abilities-api
```
justlevine, line 26: This works because `wp-env` installs the abilities api _and_ symlinks us to the correct place. If someone is calling phpstan via composer, they probably need a `phpstan.neon` that points to the real abilities-api location.

....but this is only a problem until (in theory) WP 6.9

src/Autoloader.php
justlevine, line 1: Just a BS wrapper that admin_notices if the composer autoloader is missing. If we go full WPCS or even just an internal PSR-4 we wouldn't even need this file?

(but I cant advocate an approach w.o knowing if core merge is still a 7.x goal)

src/Autoloader.php
```diff
@@ -0,0 +1,90 @@
...
+			add_action(
+				$hook,
+				static function () {
+						$error_message = __( 'MCP Adapter: The Composer autoloader was not found. If you installed the plugin from the GitHub source code, make sure to run `composer install`.', 'mcp-adapter' );
+
+					if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+						error_log( esc_html( $error_message ) ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log -- This is a development notice.
+					}
+					?>
+					<div class="error notice">
```
justlevine, line 80: IRL we're WP 6.8+ We can use `admin_notices()`

src/Core/McpAdapter.php
justlevine, line 1: All of these were just `phpcbfed` and `rector`ed. They probably look like crap. On my todo to review before bringing this up for real discussion

src/Plugin.php
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		return self::$instance;
+	}
+
+	/**
+	 * Setup the plugin.
+	 */
+	private function setup(): void {
+		McpAdapter::instance();
```
justlevine, line 50: This is more illustrative, since im pretty sure `McpAdapter` is a singleton and doesn't need to be instantiated before use.
If in theory there was future MCP-cetric stuff that did need centralized loading, this would be the place.

tests/_output/.gitkeep
justlevine, line 1: a nice clean folder for our tests, coverage, and tooling caches to make cleanup easy.

tests/bootstrap.php
justlevine, line 1: This is now the standard WP PHPUnit boostrapper

tests/bootstrap.php
```diff
@@ -1,165 +1,46 @@
...
-require "{$_tests_dir}/includes/bootstrap.php";
+// Give access to tests_add_filter() function.
+require_once $_test_root . '/includes/functions.php';
+
+// Activate the plugin.
+tests_add_filter(
+	'muplugins_loaded',
+	static function (): void {
+		// Load the abilities API, next to the MCP adapter.
+		require_once dirname( __DIR__, 2 ) . '/abilities-api/abilities-api.php';
```
justlevine, line 39: Installed by wp-env

justlevine (commented):
notes to self, probably not in a state where its worth other people chiming in yet, but if you do look at this and have thoughts _please_ feel free to share.

galatanovidiu:
Thanks for exploring this approach, @justlevine! I appreciate the thorough refactoring work here.

I have some concerns about positioning this as a standalone WordPress plugin that I'd like to discuss:

**Core concern**: In its current form, this package provides foundational infrastructure rather than end-user functionality. Making it a plugin could create confusion about its intended use case, since users installing it directly wouldn't see immediate value without additional implementation work.

**Potential issues I see:**
- **User expectations**: Plugin directory users typically expect immediate functionality after activation
- **Discoverability challenges**: Without clear server implementations, users might struggle to understand how to utilize the adapter
- **Scope creep risk**: Pressure to add user-facing features to justify plugin status

**Alternative approach**: What if we keep this repository focused as a pure library/package, and create a separate `mcp-adapter-plugin` repository when we need plugin functionality? This would:

- **Maintain clear separation**: Library stays focused on core MCP infrastructure
- **Enable multiple plugin types**: Different plugins could consume the same foundational package
- **Preserve flexibility**: Developers can use the library directly or via the plugin wrapper

This approach would let us optimize each repository for its specific purpose: this one for library consumers, and a future plugin repo for end-users who need turnkey functionality.

I'd love to hear your thoughts on this separation-of-concerns approach versus the unified repository model you're exploring here.

justlevine:
@galatanovidiu thanks for diving in, even while my thoughts are still so messy 😅 .

I think the broader question that needs to be answered is "what is the goal and expected use-case of the MCP Adapter?" 

If the plan is (still?) for this to be merged to Core - just on a longer timetable than MCP Adapter - the best practices/use cases are going to be different than if we don't have any plans for core merge (a la Two Factor, WPGraphQL, or ActionScheduler if it were .org), and even more different if we're doing upstream syncs piecemeal (like `gutenberg` or `performance`).

> **Core concern**: In its current form, this package provides foundational infrastructure rather than end-user functionality. Making it a plugin could create confusion about its intended use case, since users installing it directly wouldn't see immediate value without additional implementation work.

I think the questions that need alignment are:

- "foundational infrastructure" _for what_?
- _What is_ "its intended use case"? and
- _Who is_ the intended audience (we agree it's not "end-users")?

My understanding (per the [Building Blocks post](https://make.wordpress.org/ai/2025/07/17/mcp-adapter/) and subsequent discussions) is that we are building core MCP support for WordPress, starting "on building the server side", then possibly expanding to an MCP Client.

This seems similar enough to both the WP REST API (that started as a plugin and not a Composer library) and (the canonical) WPGraphQL plugin, so even if we weren't recommending `Requires Plugins` and users were "installing it directly", I think they'd get exactly what they expect from a canonical plugin titled `MCP Adapter` with a jargony description referencing wrappers, adapters, and a nascent `abilities_api`, even though these features are not user facing in the traditional sense.

Using WPGraphQL to strengthen the example, there's also the possibility that we _will_ need MCP-specific Admin UX down the line, even if I don't know what yet (e.g. permissions, approved clients, logs... but something that actually justifies a UI for config 😅).

> **Potential issues I see:**
> [...]

(I think above answers why I don't share the `User expectations` and `Discoverability challenges` concerns, but happy to discuss more directly if you're unswayed.)

> * **Scope creep risk**: Pressure to add user-facing features to justify plugin status

Scope creep is a risk for any project, plugin _or_ library. The solution is defining a clear scope and adhering to it, not to avoid plugin status. AFAIK `performance`, `plugin-check`, `two-factor` haven't scope creeped, and I don't think this repo is as overarching as `gutenberg`.

Also,this only really becomes a problem if the plan is for the MCP Adapter - plugin or library - to remain outside of Core in the very long term (2+ years) or coexist with upstream syncs, but also remember we already have a "dumping-ground" for scope creep in the form of AI Experiments.

---

> **Alternative approach**: What if we keep this repository focused as a pure library/package, and create a separate `mcp-adapter-plugin` repository when we need plugin functionality?
> [...]
> I'd love to hear your thoughts on this separation-of-concerns approach versus the unified repository model you're exploring here.

I'm not opposed to having separate library-only and "full-fat" plugin releases, if there's a real use case for them coexisting (which to my gut seems like a weaker case than even just releasing as a library).

But if so, I'd recommend we develop and release from **this repo**. should be just as easy to exclude the plugin wrapper files from Composer/Packagist while keeping them .zip to GH Release + .org pipleine and matching `wpackagist-plugin` Comperser version, without any of the unnecessary overhead.

As for the reasons:

> This would:
> 
> * **Maintain clear separation**: Library stays focused on core MCP infrastructure

(I think this was already addressed above, if not happy to discuss more but tl;dr:)

PSR-4 + namespaces should be handling all internal separation of concerns (in theory - I havn't yet gone deep into `/src`, and if not it's early enough for us to refactor to ensure our quasi DRY/SOLID isolates well enough), and then we have the Abilities API on one side of us and the AI Experiments on the other.

> * **Enable multiple plugin types**: Different plugins could consume the same foundational package

Not entire sure I understand what this means.
- What does "plugin types" mean in this context, and which types are limited if we go the plugin route?
- More broadly, what's the nuance between "consum[ing] the same foundational package" and relying on the same plugin?

> * **Preserve flexibility**: Developers can use the library directly or via the plugin wrapper

As long as we follow the aforementioned "internal separation of concerns", users can still use this as a library, more or less the same way this PR (and IRL abilities api and performance) uses `plugin-check` as a library instead of a plugin. Something along the lines of (theorycraft):

```php
// If the plugin isn't installed and we dont want to use the Requires Plugins header to make them install it.
if ( ! class_exists( 'WP\MCP\Core\McpAdapter' ) {
  // if we ditch Composer's autoloader for internal this wouldn't even need to plugin chrome..
  // ... if that's something we want???
  require_once MY_PLUGIN_PATH . '/vendor/wordpress/mcp-adapter/includes/Core/Autoload.php';
  \WP\MCP\Core\Autoloader::autoload();
}

// same usage... 
$adapter = WP\MCP\Core\McpAdapter::instance();
```

But it'd be easier to propose a specific pattern for consuming this plugin as a library, if I had some insight or examples for justified use cases as to why somebody would want to go that route (and/or why we'd want them to)

jeffpaul:
My opinion/understanding so far was to structure this as a composer package so others could leverage it (e.g. without needing to worry about plugin dependencies), including the [AI Experiments plugin](https://github.com/wordPress/ai) (as well as things like [ClassifAI](https://github.com/10up/classifai) and other AI tools in the space).

justlevine:
> My opinion/understanding so far was to structure this as a composer package so others could leverage it (e.g. without needing to worry about plugin dependencies), including the [AI Experiments plugin](https://github.com/wordPress/ai) (as well as things like [ClassifAI](https://github.com/10up/classifai) and other AI tools in the space).

As a reminder canonical plugins can still be used as a composer packages, but composer packages require Composer and cannot be autoinstalled by WordPress. Until it's merged in core, the Abilities API is available as both a package and a plugin.

@jeffpaul can you clarify 
- what issues you see with people being able to "leverage" this if it was available as canonical plugin _and_ a Composer library?
- why you think it'd be better for the AI Experiments (or really any plugin that wants to integrate with core's MCP Adapter) to have this as a library instead of a Required Plugin?

As I wrote in my last reply to @galatanovidiu , I'm hoping having a justifiable "pro-library" (versus a "yes-and") use case will help clarify the code review and other [MCP Adapter objectives](https://make.wordpress.org/ai/2025/07/17/mcp-adapter/).

.wp-env.json
justlevine: Done in #18

README.md
justlevine: Done in #16

bin/install-wp-tests.sh
justlevine: Done in #18

src/Autoloader.php
justlevine: Handled in #18

src/Plugin.php
justlevine: Done in #18

tests/bootstrap.php
justlevine: Done in https://wordpress.slack.com/archives/C08TJ8BPULS/p1756056964263679

package.json
justlevine: Done in #18

phpcs.xml.dist
justlevine: Done in #14

phpstan.neon.dist
justlevine: Done in #18

phpunit.xml.dist
justlevine: Done in #14

README.md
justlevine: Done in #18

README.md
justlevine: Removed in #18 (it doesn't belong in a root readme anyway)

justlevine:
The qualified feedback from this has been backported, and further iteration/feedback can happen independently. Closing.

Issue #19, opened 2025-08-25
`testing.md` should use wp-env steps


justlevine:
As of #18 , we now run and test this plugin using `wp-env`, as described in https://github.com/WordPress/mcp-adapter/blob/trunk/CONTRIBUTING.md#running-tests .

As such, we need to update https://github.com/WordPress/mcp-adapter/blob/trunk/docs/guides/testing.md accordingly (and/or decide to deduplicate).

justlevine:
Didn't realize how many docs there already were, seems there's other docs that need updating, as well as the two readmes to dedupe...

PR #21, opened 2025-08-25
Update MCP server creation instructions in README

jonathanbossenger:
Added details on the hook to use, and the parameter it can accept.


justlevine (approved):
This is accurate.

Any conversations about updating/deduping docs  (#19) or whether we want to cement/publicize custom server registration into v0.1.0 (https://wordpress.slack.com/archives/C08TJ8BPULS/p1755935323664639?thread_ts=1755849165.480119&cid=C08TJ8BPULS) can happen later/independently of this PR.

Issue #22, opened 2025-08-27
Reduce and lock down the API footprint for initial release


justlevine:
## What

For the initial (working) release, I recommend we lock down the public-facing API to the absolute minimum required. E.g.

- Marking classes `final`
- Auditing the `public` methods for either removal or `@internal` annotations.

## Why

The current codebase is a fully-implemented DRY/SOLID abstraction, that didn't have a chance to undergo public review before merge in #3 . Limiting the scope of what's exposed publicly in the initial release will allow us to rely on and holistically iterate on the codebase without worrying about prescriptive patterns or tech debt.

## Alternatives considered

- Auditing 8k loc for usage patterns/antipatterns, DX simplification and WP core alignment. This is slow, and doesn't matter if the initial release goal is consumption/usage and not end-to-end extensibility or core WP merge candidacy. Opinionated code isn't inherently problematic, and we have tests and tools to confirm it works.
- Removing "beyond the scope" functionality, and re-adding it in the future. This seems silly because again, the code itself is fine in and of itself.

PR #26, opened 2025-08-29
dev!: refactor non-registry functionality from `Core/McpAdapter` 

justlevine:
## What

This PR cleans up the `Core/McpAdapter` server registry, by (re)moving functionality that is unused or unrelated to server registry handling.

## Why

SRP - now that we have a `Plugin` wrapper we don't need the extra complexity/boilerplate.

## How

See diff comments.

### DIFF
--- a/includes/Core/McpAdapter.php
+++ b/includes/Core/McpAdapter.php
@@ -2,7 +2,7 @@
 /**
  * WordPress MCP Registry - Main class for managing multiple MCP servers.
  *
- * @package McpAdapter
+ * @package WP\MCP\Core
  */
 
 declare( strict_types=1 );
@@ -17,34 +17,13 @@ use WP\MCP\Infrastructure\Observability\NullMcpObservabilityHandler;
 /**
  * WordPress MCP Registry - Main class for managing multiple MCP servers.
  */
-class McpAdapter {
+final class McpAdapter {
 	/**
 	 * Registry instance
 	 *
-	 * @var \WP\MCP\Core\McpAdapter|null
+	 * @var \WP\MCP\Core\McpAdapter
 	 */
-	private static ?self $instance = null;
-
-	/**
-	 * The initialized flag.
-	 *
-	 * @var bool
-	 */
-	private static bool $initialized = false;
-
-	/**
-	 * Flag to track if initialization failed due to missing dependencies.
-	 *
-	 * @var bool
-	 */
-	private static bool $initialization_failed = false;
-
-	/**
-	 * Stores the reason for initialization failure.
-	 *
-	 * @var string[]
-	 */
-	private static array $initialization_errors = array();
+	private static self $instance;
 
 	/**
 	 * Registered servers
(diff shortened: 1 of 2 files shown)


includes/Core/McpAdapter.php
```diff
@@ -17,34 +17,13 @@
 /**
  * WordPress MCP Registry - Main class for managing multiple MCP servers.
  */
-class McpAdapter {
+final class McpAdapter {
```
justlevine, line 20: 1. `final` unless/until we have a specific use case for extending this class means we don't need to worry about the shape of the public methods. See #22 
2. Personally, I'd rename this class `Registry` but since the outdated docs all still reference this class I didn't want to deal with that right now.

includes/Core/McpAdapter.php
```diff
@@ -54,146 +33,51 @@ class McpAdapter {
...
-			return;
-		}
-
-		if ( ! $this->check_dependencies() ) {
-			self::$initialization_failed = true;
-			return;
-		}
-		add_action( 'rest_api_init', array( $this, 'mcp_adapter_init' ), 20000 );
-		self::$initialized = true;
-	}
```
justlevine, line 77: Moving the `add_action()` to `::instance()` allows us to clean out the constructor entirely, meaning we can test and use without needing additional class properties for lifecycle state.

includes/Core/McpAdapter.php
```diff
@@ -54,146 +33,51 @@ class McpAdapter {
...
+	 * @return \WP\MCP\Core\McpAdapter
 	 */
-	public static function instance(): ?self {
-		if ( null === self::$instance && ! self::$initialization_failed ) {
+	public static function instance(): self {
+		if ( ! isset( self::$instance ) ) {
 			self::$instance = new self();
+
+			// Defer initialization until after the REST API.
+			add_action( 'rest_api_init', array( self::$instance, 'init' ), 20000 );
```
justlevine, line 54: I'm not sure _why_ we're deferring, I'm just preserving what was.
cc @galatanovidiu

includes/Plugin.php
```diff
@@ -47,9 +47,43 @@ public static function instance(): self {
 	 * Setup the plugin.
 	 */
 	private function setup(): void {
+		// Bail if dependencies are not met.
+		if ( ! $this->has_dependencies() ) {
+			return;
+		}
+
 		McpAdapter::instance();
```
justlevine, line 55: By only hooking into WordPress if dependencies are met, we don't need the registry itself to handle state/availability checks

includes/Core/McpAdapter.php
```diff
@@ -54,146 +33,51 @@ class McpAdapter {
...
 	 * @param callable|null $transport_permission_callback Optional custom permission callback for transport-level authentication. If null, defaults to is_user_logged_in().
 	 *
-	 * @return \WP\MCP\Core\McpAdapter
+	 * @return true|\WP_Error True on success, WP_Error on failure.
 	 * @throws \Exception If the server already exists or if called outside of the mcp_adapter_init action.
 	 */
-	public function create_server( string $server_id, string $server_route_namespace, string $server_route, string $server_name, string $server_description, string $server_version, array $mcp_transports, ?string $error_handler, ?string $observability_handler = null, array $tools = array(), array $resources = array(), array $prompts = array(), ?callable $transport_permission_callback = null ): self {
+	public function create_server( string $server_id, string $server_route_namespace, string $server_route, string $server_name, string $server_description, string $server_version, array $mcp_transports, ?string $error_handler, ?string $observability_handler = null, array $tools = array(), array $resources = array(), array $prompts = array(), ?callable $transport_permission_callback = null ) {
```
justlevine: This leaked for a different PR.

```suggestion
	 * @return \WP\MCP\Core\McpAdapter
	 * @throws \Exception If the server already exists or if called outside of the mcp_adapter_init action.
	 */
	public function create_server( string $server_id, string $server_route_namespace, string $server_route, string $server_name, string $server_description, string $server_version, array $mcp_transports, ?string $error_handler, ?string $observability_handler = null, array $tools = array(), array $resources = array(), array $prompts = array(), ?callable $transport_permission_callback = null ): self {
```

includes/Plugin.php
```diff
@@ -20,9 +20,9 @@ final class Plugin {
 	/**
 	 * The one true plugin.
 	 *
-	 * @var ?static
+	 * @var static
 	 */
-	private static $instance;
+	private static self $instance;
```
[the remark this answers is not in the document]
justlevine: Its a `final` class smh

includes/Core/McpAdapter.php
```diff
@@ -17,34 +17,13 @@
...
-	 */
-	private static bool $initialization_failed = false;
-
-	/**
-	 * Stores the reason for initialization failure.
-	 *
-	 * @var string[]
-	 */
-	private static array $initialization_errors = array();
+	private static self $instance;
```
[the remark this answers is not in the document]
justlevine: There is no inheritance support 🤦

includes/Core/McpAdapter.php
galatanovidiu: You're right to preserve it. The rest_api_init hook is actually required here - it's not an arbitrary deferral.
The MCP adapter creates servers that instantiate transports through the REST API.
The codebase even has a guard that enforces server creation only happens during the mcp_adapter_init action. So the current timing is architecturally necessary, not just a preference.

This will also allow all other plugins to register abilities that can be exposed through the MCP adapter rest transports.

justlevine: That much I figured, which is why I left it out of this PR😅

What I'm unclear about:
- This defers _everything_, not _just_ the transports/execution/whatever other relevant bits (I'm still going through it all).
- Why do the transports need their own separate hooks e.g. https://github.com/justlevine/mcp-adapter/blob/b8e595343e68278e56a6a872db36058a1ed075db/includes/Transport/Http/StreamableTransport.php#L49
- Why the `20000`, `20001` and `20002` priorities?

includes/Core/McpAdapter.php
galatanovidiu: This might be the best time to introduce breaking changes.
For the name, I would go with McpAdapterRegistry (I like to be a bit more verbose when naming things :) )

includes/Core/McpAdapter.php
galatanovidiu: These are some hacks I had to use while testing with an early version of abilities-api.
I think we can use the default priority here. But we need to test this with different scenarios.

galatanovidiu (approved):
LGTM

includes/Core/McpAdapter.php
justlevine: > This might be the best time to introduce breaking changes.

Great I'll create a new PR once this is merged.

> For the name, I would go with McpAdapterRegistry 

Per https://github.com/WordPress/mcp-adapter/pull/32/files#r2313545529 , I'm now leaning more to `WP/MCP/Core/ServerRegistry` to offset the `WP/MCP/Core/ClientRegistry`. (or Mcp{Client|Server}Registry if you don't think it's redundant). Either way, we'll figure it out on that PR 🤷

includes/Core/McpAdapter.php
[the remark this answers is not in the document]
justlevine: That's _where_ it's instantiated smh

includes/Plugin.php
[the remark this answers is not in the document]
justlevine: That's _where_ it's instantiated smh

PR #43, opened 2025-09-09
Relax parameter validation to allow some falsy values as parameters.

aaronfc:
While using the adapter I found that `0` could not be sent as a parameter to the tools.

I am relaxing the check here to allow `0` and any other falsy value except from: `'null'`, `''` and `null`.

I can relax that even more but I am missing context on why those were added in first place.

### DIFF
--- a/includes/Handlers/Tools/ToolsHandler.php
+++ b/includes/Handlers/Tools/ToolsHandler.php
@@ -110,9 +110,11 @@ class ToolsHandler {
 		if ( ! empty( $request_params['arguments'] ) ) {
 			foreach ( $request_params['arguments'] as $key => $value ) {
 				// Only remove truly empty values (null, empty strings) but preserve valid falsy values like 0
-				if ( null === $value || '' === $value || 'null' === $value ) {
-					unset( $request_params['arguments'][ $key ] );
+				if ( null !== $value && '' !== $value && 'null' !== $value ) {
+					continue;
 				}
+
+				unset( $request_params['arguments'][ $key ] );
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


galatanovidiu (approved):
LGTM with a note.
We might remove required parameters here.
But for now, we can keep it simple.

galatanovidiu:
> While using the adapter, I found that `0` could not be sent as a parameter to the tools.
> 
> I am relaxing the check here to allow `0` and any other falsy value except from: `'null'`, `''` and `null`.
> 
> I can relax that even more but I am missing context on why those were added in first place.

They have been added based on previous experience with MCP clients sending empty arguments, even when they are not necessary/required, which can cause the tool's execution to fail.

Any thoughts on this @justlevine @felixarntz

justlevine (approved):
> Any thoughts on this

@galatanovidiu I definitely hear the value in wanting to be as forgiving as possible of MCP-generated `$inputs`, and I also much prefer making those accommodations here instead of compromising the schema upstream in `abilities-api`.

At the same time, `$request_parameters['arguments']['some_arg'] = 0;`  (or `''` et al in some cases, this PR is a bit of a compromise already) has real semantic value, and we need a way for an MCP to be able to pass that to an Ability.

I think at some point we need to hope the information provided by schema validation is enough to correct the MCP if they provide a fasley that's not empty.

galatanovidiu:
@justlevine, you're absolutely right. Looking at this more carefully, I realize the "Clean parameters arguments" functionality was unnecessary and potentially problematic.

This stems from my experience with previous MCP projects, where I used REST API endpoints as MCP tools and encountered issues with parameter handling. However, in this context, adding parameter cleaning was the wrong approach - the schema validation should handle invalid inputs.

So, I think we should remove the "Clean parameters arguments" functionality altogether.

What do you think @justlevine @aaronfc ?

aaronfc:
Sounds good to me @galatanovidiu . I have fone ahead and removed the cleanup in 
f0d99f1.

Let's see what @justlevine thinks.

PR #46, opened 2025-09-11
Implement Unified HttpTransport for MCP 2025-06-18 Compliance

galatanovidiu:
Closes #45

## Overview

This PR implements a new unified `HttpTransport` class that replaces the fragmented transport architecture with a single, MCP 2025-06-18 compliant transport layer. The new implementation provides enhanced session management, CORS support, and comprehensive JSON-RPC 2.0 compliance.

### New Transport Implementation
- **Added `includes/Transport/HttpTransport.php`** - Unified transport implementing MCP 2025-06-18 Streamable HTTP specification
- **Enhanced notification handling** - Proper 202 responses for JSON-RPC notifications
- **Session management** - Complete lifecycle with `Mcp-Session-Id` headers
- **CORS compliance** - Full preflight and header support
- **Batch processing** - Handle single messages and batch arrays

### Deprecation Updates
- **Updated `RestTransport.php`** - Added deprecation notices and `_deprecated_class()` calls
- **Updated `StreamableTransport.php`** - Added deprecation notices and `_deprecated_class()` calls
- Both classes now point users to `HttpTransport` as replacement

### Comprehensive Testing
- **Added `tests/Integration/HttpTransportTest.php`** - Complete MCP compliance test suite
- **19 integration tests** covering all specification requirements
- **HTTP method compliance** - GET, POST, DELETE, OPTIONS testing
- **Session lifecycle testing** - Creation, validation, termination
- **Error response validation** - JSON-RPC 2.0 format compliance
- **CORS header verification** - Preflight and response header testing

### Documentation Updates
- **Updated all code examples** - Changed from `RestTransport` to `HttpTransport`
- **Enhanced testing guide** - Added MCP compliance testing section
- **Updated architecture docs** - Reflect new unified transport approach
- **Added deprecation warnings** - Clear migration guidance throughout

### MCP 2025-06-18 Compliance
- HTTP methods: GET, POST, DELETE, OPTIONS
- JSON-RPC 2.0 message validation
- Session management with secure IDs
- CORS preflight handling
- Notification processing (202 responses)
- Batch request support
- Protocol version header handling

### Unified Client Support
- Works with mcp-wordpress-remote proxy
- Supports direct MCP client connections
- Intelligent request type detection
- Consistent behavior across connection types

### Enhanced Security
- Origin header validation framework
- Session-based request validation
- Configurable permission callbacks
- Secure session storage using WordPress transients

## Backward Compatibility

[...]

PR #48, opened 2025-09-17
Major Refactor: Enhanced Transport Infrastructure, Default Server, and CLI Support

galatanovidiu:
## Summary

This PR represents a comprehensive architectural restructuring of the WordPress MCP Adapter, addressing significant technical debt while introducing three major feature sets.

While I understand the importance of small, atomic PRs for easier review, the scope and interconnected nature of the issues made it impractical and time-consuming.

I hope this work unblocks a cleaner, more modular workflow moving forward. 

### 1. HttpTransport - Modern Transport Infrastructure
The new `HttpTransport` class replaces the existing `RestTransport` and `StreamableTransport` implementations, providing a unified, more robust transport layer.

**Key Features:**
- **Unified HTTP handling** with dedicated request processing services (`HttpRequestHandler`, `RequestRouter`)
- **Session management** with user-based authentication (`SessionManager`, `HttpSessionValidator`)
- **Standardized error responses** via `JsonRpcResponseBuilder` with proper HTTP status code mapping
- **Improved request context** handling through `HttpRequestContext` and `McpTransportContext`
- **Better separation of concerns** with specialized infrastructure components

**Technical Benefits:**
- Provides consistent error handling across all HTTP-based transports
- Implements proper session lifecycle management for stateful operations
- Improves security through built-in authentication validation

### 2. Default Server with Layered Tools Architecture
A new default MCP server that automatically registers with the adapter, featuring a layered approach to exposing WordPress abilities as MCP tools.

**Components:**
- **DefaultServerFactory**: Automated server creation and registration
- **Layered Tools System**: Three specialized tools for ability discovery and execution:
  - `DiscoverAbilitiesTool`: Discovers available WordPress abilities
  - `GetAbilityInfoTool`: Retrieves detailed information about specific abilities
  - `ExecuteAbilityTool`: Executes abilities with proper parameter handling

**Technical Benefits:**
- Implements dynamic ability discovery without hard-coding tool definitions
- Provides a flexible execution model that adapts to WordPress capabilities
- Establishes a standardized interface for WordPress functionality access

### 3. STDIO Server Bridge for CLI Integration
Complete WP-CLI integration enabling MCP servers to be exposed through STDIO transport, making them accessible from command-line tools and external applications.

**Features:**
- **WP-CLI Command**: New `wp mcp-adapter` command for server management
- **StdioServerBridge**: Bridges MCP servers to STDIO for command-line communication
- **Universal Server Access**: Any registered MCP server can be accessed via CLI

**Usage:**
```bash
# Start an MCP server via STDIO
wp mcp-adapter serve --server=mcp-adapter-default-server --user=admin

# List available servers
wp mcp-adapter list
```

**Technical Benefits:**
- Enables STDIO transport protocol

### 4. MCP Component Type System & Auto-Discovery
Introduction of a type-based system for categorizing and automatically discovering MCP components (tools, resources, and prompts).

**Key Features:**
- **Type Classification**: Abilities can be explicitly marked as `tool`, `resource`, or `prompt` via `meta.mcp.type`
- **Public Access Control**: Fine-grained control over which abilities are exposed through MCP via `meta.mcp.public`
- **Auto-Discovery**: The default server automatically discovers resources and prompts based on their `mcp.type` without manual registration
- **Default Behavior**: Abilities default to type `tool` if not specified

**Example:**
```php
wp_register_ability('my-plugin/post-resource', [
    'label' => 'Post Resource',
    'description' => 'Access WordPress posts',
    'meta' => [
        'mcp' => [
            'public' => true,     // Expose via MCP
            'type'   => 'resource' // Auto-discovered as resource
        ],
        'uri' => 'site://posts/{id}' // Resource-specific metadata
    ]
]);
```

**Technical Benefits:**
- Eliminates the need for manual registration of resources and prompts in the default server
- Provides clear component type distinction for better organization
- Enables the default server to automatically configure itself based on ability metadata

### Enhanced Core Infrastructure

- **Component Registry** (`McpComponentRegistry`): Centralized management of MCP components with improved lifecycle handling
- **Transport Factory** (`McpTransportFactory`): Standardized transport creation with dependency injection
- **Handler Standardization** (`HandlerHelperTrait`): Consistent implementation patterns across all handlers
- **Improved Separation of Concerns**: Clear boundaries between transport, server, and handler layers
- **Better Dependency Management**: Reduced coupling between components through interface-based design

### Error Handling Improvements

- **Standardized Error Codes**: Consistent JSON-RPC error code mapping to HTTP status codes
- **Structured Error Responses**: `JsonRpcResponseBuilder` ensures uniform error format across all endpoints
- **Enhanced Validation**: Comprehensive validation for tool execution, request processing, and parameter handling
- **Better Error Context**: More informative error messages with proper error chain propagation

### Code Quality Enhancements

- **Type Safety**: Improved type hints and generic specifications throughout the codebase
- **Static Analysis**: PHPStan compliance improvements with proper type annotations
- **Code Standards**: Full WordPress coding standards compliance
- **Dependency Management**: Updated to the latest versions of development dependencies
- **Configuration Consistency**: All hooks and filters use standardized `mcp_adapter_` prefix

## Testing & Quality Assurance

**Comprehensive Test Coverage** (26 test files added/modified):
- Unit tests for all new infrastructure components
- Integration tests for HTTP transport layer and request routing
- Session management and authentication flow tests
- Error handling consistency validation tests
- Handler trait behavior tests
- Transport infrastructure validation tests

### Removed Components
- **`RestTransport`**: Removed and replaced by `HttpTransport` - all functionality preserved in the new implementation
- **`StreamableTransport`**: Removed and replaced by `HttpTransport` - enhanced streaming capabilities included
- Migration guide provided below for updating existing implementations

### Filter Name Standardization
All WordPress filter names now use the `mcp_adapter_` prefix for consistency:
- `mcp_rest_transport_*` → `mcp_adapter_rest_transport_*`
- `mcp_streamable_transport_*` → `mcp_adapter_streamable_transport_*`
- Custom filter implementations require prefix updates

### Transport Migration

See the [migration guide](docs/migration/v0.3.0.md) for detailed migration instructions.

### Filter Migration
```php
// Old filter names
add_filter('mcp_validation_enabled', $callback);

// New standardized names
add_filter('mcp_adapter_validation_enabled', $callback);
```

### New Infrastructure Components
- `HttpTransport` - Primary transport implementation with HTTP/1.1 and HTTP/2 support
- `HttpRequestHandler` - Request processing with proper header management
- `HttpSessionValidator` - Session validation with WordPress user integration
- `SessionManager` - Stateful session management with cleanup procedures
- `RequestRouter` - Intelligent request routing based on HTTP method and content type
- `JsonRpcResponseBuilder` - JSON-RPC compliant response formatting
- `McpTransportContext` - Enhanced transport context with better state management
- `McpComponentRegistry` - Component lifecycle management with proper initialization
- `McpTransportFactory` - Transport instantiation with configuration injection

### CLI Implementation Components
- `McpCommand` - WP-CLI command implementation with subcommand support
- `StdioServerBridge` - STDIO transport bridge with proper stream handling

### Default Server Tools
- `DiscoverAbilitiesTool` - WordPress ability discovery with metadata extraction
- `ExecuteAbilityTool` - Ability execution with parameter validation and error handling
- `GetAbilityInfoTool` - Detailed ability information retrieval
- `DefaultServerFactory` - Automated server creation with configuration management

### Resolved Issues
- Eliminated circular dependencies in the transport layer
- Fixed inconsistent error response formats
- Resolved session state management problems
- Addressed missing type hints and annotations

### Architecture Improvements
- Implemented proper dependency injection patterns
- Established clear interface contracts
- Improved separation of concerns across layers
- Standardized configuration management
- Enhanced error propagation mechanisms

## Future Technical Considerations

- Enhanced CLI commands for advanced server management scenarios
- Performance optimizations for high-concurrency environments
- Additional metadata-driven features for component discovery and management


includes/Abilities/DiscoverAbilitiesAbility.php
```diff
@@ -0,0 +1,108 @@
...
+	 */
+	public static function execute( $input = array() ): array {
+		$abilities = wp_get_abilities();
+
+		$ability_list = array();
+		foreach ( $abilities as $ability ) {
+			$ability_name = $ability->get_name();
+
+			// Exclude abilities that start with 'mcp-adapter/' to prevent self-referencing
+			if ( str_starts_with( $ability_name, 'mcp-adapter/' ) ) {
```
gziolo: There are only 3 `mcp-adapter` abilities. Do you think they should be exposed as abilities through the REST API or for internal usage in WordPress on the server outside of MCP? I'm trying to better understand your vision here and what filtering capabilities we would have to include, as discussed in:
- https://github.com/WordPress/abilities-api/issues/38

galatanovidiu: I don’t have strong opinions here, just some ideas and context to share:

The main reason for introducing this particular filter is to exclude the internal mcp-adapter abilities from the list exposed to the MCP Client, to avoid self-referencing and potential infinite recursion.

I don’t think those three internal abilities should be exposed via the REST API by default. However, it would be helpful to provide a WordPress filter hook that developers can use to enable it if they choose to. To support this kind of flexibility, I think we’ll need to introduce a `show_in_rest` meta (or something similar).

Additionally, an abilities filter by namespace would also be beneficial.

gziolo: I'm getting a bit better understanding why it was convenient to register these meta abilities as usual. There is underlying logic in the MCP server configuraiton that allows to list tools registered as abilities.

Screenshot 2025-09-23 at 08 56 04

I agree it's better to focus on filtering capabilities next, as we need all that you described anyway. In fact, it would also allow us to have more flexibility when listing these tools or resources, because instead of providing an allow list, we could also support filters as an alternative.

swissspidy:
Does this supersede #46 and incorporate feedback from there?

I'm really not a fan of massive PRs like this that make huge architectural decisions. Were there any discussions about them that I missed? Can the PR be split up?

For instance, the WP-CLI changes can easily be extracted, or better yet reverted. Why a stdio server when there's http? Not to mention that WP-CLI is user-less, things should work without simulating a specific user.

galatanovidiu:
> Does this supersede #46 and incorporate feedback from there?

Yes, it does. I'm keeping #46 open for now until we reach consensus on the direction forward.

> I'm really not a fan of massive PRs like this that make huge architectural decisions. Were there any discussions about them that I missed? Can the PR be split up?

I'm also not a fan of these massive PRs, especially this one was really intensive and time-consuming.
More details here:
https://wordpress.slack.com/archives/C08TJ8BPULS/p1758214390610409 


> For instance, the WP-CLI changes can easily be extracted, or better yet reverted. Why a stdio server when there's http? Not to mention that WP-CLI is user-less, things should work without simulating a specific user.

I'm not pro or against exposing a server through STDIO transport. Personally, I will most likely never use this one, but this was stipulated on https://make.wordpress.org/ai/2025/07/17/mcp-adapter/.

docs/architecture/overview.md
```diff
@@ -1,708 +1,350 @@
...
-- Error rates and categorization
-- Tool execution success/failure rates
-- Permission denials and authentication events
+### Built-in Transports
 
-#### MCP Components
+- **HttpTransport**: Recommended (MCP 2025-06-18 compliant)
+- **STDIO Transport**: Via WP-CLI commands
+- **RestTransport**: Deprecated
+- **StreamableTransport**: Deprecated
```
JasonTheAdams: I'd be in favor of just removing these. It feels strange to include deprecated classes in the first release in WP 6.9.

docs/architecture/overview.md
```diff
@@ -1,708 +1,350 @@
...
-- **McpErrorHandlerInterface**: Interface for logging errors to monitoring systems  
-- **ErrorLogMcpErrorHandler**: Default WordPress-compatible error logging
-- **NullMcpErrorHandler**: Null object pattern for disabled error handling
-- **Transport Level**: Catches exceptions and coordinates error handling
-- **Handler Level**: Uses factory methods for consistent error responses
-- **Ability Level**: WordPress abilities handle validation and execution errors
+### Event Emission Pattern
 
-#### Error Response vs Error Logging
+The system emits events rather than storing counters:
```
JasonTheAdams, line 170: What do you mean by "storing counters" here?

docs/architecture/overview.md
```diff
@@ -1,708 +1,350 @@
...
 class McpAdapter {
-    private static ?McpAdapter $instance = null;
+    private static self $instance;
     
-    public static function instance(): McpAdapter {
-        if ( self::$instance === null ) {
+    public static function instance(): self {
+        if (!isset(self::$instance)) {
             self::$instance = new self();
+            add_action('rest_api_init', [self::$instance, 'init'], 15);
```
JasonTheAdams, line 197: I wish we could use the Service Provider pattern to reduce instantiation side-effects like this.

What would you think about adding a separate `initialize()` method, so it's:

```php
McpAdapter::instance()->initialize()
```

That way we can at least instantiate without side-effects.

docs/getting-started/basic-examples.md
```diff
@@ -92,491 +91,183 @@ add_action( 'abilities_api_init', function() {
...
+- **Tools**: Execute actions (like `tools/call`)
+- **Resources**: Provide data access (like `resources/read`) - require `meta.uri`
+- **Prompts**: Generate messages (like `prompts/get`) - return `messages` array
 
-## Error Handling Examples
+### Annotations
+All components support annotations in `meta.annotations`:
+- **Universal**: `priority`, `readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`
+- **Resources**: `audience`, `lastModified` (MCP specification)
+- **Prompts**: Support both template-level and message content annotations
```
JasonTheAdams: It could be useful to link out to the MCP docs here, such as https://modelcontextprotocol.io/specification/2025-06-18/server/resources#annotations

docs/getting-started/installation.md
```diff
@@ -69,39 +63,71 @@ class MyMcpPlugin {
...
+            'my-plugin',
+            'mcp',
+            'My Plugin MCP Server',
+            'Custom MCP server for my plugin',
+            '1.0.0',
+            [ \WP\MCP\Transport\HttpTransport::class ],
+            \WP\MCP\Infrastructure\ErrorHandling\ErrorLogMcpErrorHandler::class,
+            [ 'my-plugin/get-posts' ]
+        );
+    }
```
JasonTheAdams, line 125: The presence of this here makes it seem like creating a per-plugin server is recommended, but my understanding is that it's really not. I'm not personally aware of any benefits of having many MCP servers running over just one, especially with tool layering.

docs/guides/creating-abilities.md
```diff
@@ -1,711 +1,392 @@
...
+]);
 ```
 
-## Error Handling Strategies
-
-### Structured Error Responses
+## Creating Resources
 
-Create consistent error responses that AI agents can understand:
+Resources provide access to data or content. They require a `uri` in the meta field:
```
JasonTheAdams: I'm mildly concerned that the way a tool/resource/prompt is determined is by inferring it from the parameters such as `uri`. I wonder if it would be better to have an `mcp_component_type` meta parameter that explicitly sets the type. This would be more clearly polymorphic and provide better validation for the intended component type.

But maybe I'm missing something. I'm just not seeing how the component type is clearly determined in the examples below.

docs/guides/custom-transports.md
```diff
@@ -1,151 +1,119 @@
...
-1. [When to Create Custom Transports](#when-to-create-custom-transports)
-2. [Transport Architecture](#transport-architecture)
-3. [Basic Custom Transport](#basic-custom-transport)
-4. [Advanced Transport Features](#advanced-transport-features)
-5. [Real-World Examples](#real-world-examples)
-6. [Testing and Debugging](#testing-and-debugging)
+- ✅ **`HttpTransport`** - Recommended (implements MCP 2025-06-18 specification)
+- ✅ **`STDIO Transport`** - Available via WP-CLI commands
+- ❌ **`RestTransport`** - Deprecated, use `HttpTransport`
+- ❌ **`StreamableTransport`** - Deprecated, use `HttpTransport`
```
JasonTheAdams: As said above, I think these should be outright removed, not deprecated.

docs/guides/default-server.md
```diff
@@ -0,0 +1,343 @@
+# Default MCP Server
+
+The MCP Adapter automatically creates a default server that provides core MCP functionality for WordPress abilities. This server acts as a bridge between AI agents and WordPress, allowing them to discover and execute WordPress abilities through the Model Context Protocol.
```
JasonTheAdams, line 3: I wonder if it would be helpful here to include a short paragraph explaining the Layered Tooling approach being taken here, as it may not be obvious to some why abilities aren't a 1-to-1 with MCP components.

docs/guides/error-handling.md
```diff
@@ -124,99 +86,70 @@ $error = McpErrorFactory::tool_not_found(123, 'missing-tool');
 
 ### Error Codes
 
-The system uses standard JSON-RPC error codes:
+Standard JSON-RPC and MCP-specific error codes:
 
 ```php
-// Standard JSON-RPC codes (-32768 to -32000)
+// Standard JSON-RPC codes
 const PARSE_ERROR      = -32700;
```
JasonTheAdams, line 93: Where are these constants defined to be referenced? `McpErrorFactory`?

docs/guides/observability.md
JasonTheAdams, line 1: Am I missing where we explain how to register a handler with the server? I suspect it's documented elsewhere, but it seems worth including in here or at least linking out to.

JasonTheAdams (changes requested):
Amazing documentation here, @galatanovidiu! I haven't gone through the code yet, but the docs helped a lot in inferring how things work. Hopefully this helps with some conversation at a high level.

docs/architecture/overview.md
galatanovidiu: The mcp-adapter will not be included in WordPress core, and there are no current plans to do so.
For now, it can be used either as a standalone plugin or as a package, depending on the use case.

docs/architecture/overview.md
galatanovidiu: Good catch, this is a leftover from a previous version of the docs.
There used to be an observer that counted events, but I removed it since it wasn’t particularly useful, and it was not working properly anyway.
I’ll revise this section to avoid confusion.

docs/getting-started/basic-examples.md
galatanovidiu: I've redone this section.
I've removed the list as it will be best to get the details on what annotations are available and how to use them directly from the MCP specifications.

docs/getting-started/installation.md
galatanovidiu: I think we should keep this functionality available, but we shouldn’t encourage creating a server per plugin as the default approach. In most cases, a single MCP server is sufficient and avoids unnecessary complexity. That said, there may still be edge cases where bootstrapping your own server makes sense.

@Jameslepage @swissspidy @gziolo @justlevine — curious to hear your thoughts.

docs/guides/creating-abilities.md
galatanovidiu: Yes, you’re absolutely right.

Currently, the default server exposes all capabilities as MCP tools, which is not correct. We definitely need a way to differentiate between component types (tools, resources, prompts, etc.).

While implementing this isn’t a huge technical task, there are a few ongoing discussions around the abilities API and filtering/built-in metas, so I’ve been holding off until those are settled.

The only MCP-related meta I'm currently checking is the `public_mcp`; However, I think we need a more structured approach for MCP-related metadata. Example:  

```
{
...
"meta": {
    "show_in_rest": true,
    "mcp": {
      "public": true,
      "type": "tool",
      ...
    }
  }
}

File: docs/getting-started/installation.md (hunk 6 above)
Comment (gziolo) [reply docs/getting-started/installation.md:125]: Is there a way to instantiate a shared global instance of the default MCP server? This way we would promote a simplified usage that is going to be sufficient to most cases. The existing example for a plugin-specific server could be included in the `Advanced Usage` section.

File: docs/guides/creating-abilities.md (hunk 7 above)
Comment (gziolo) [reply docs/guides/creating-abilities.md]: It feels like we need to resolve the existing issue in the Abilities API that will help to reason about resources vs tools in a more general way:
- https://github.com/WordPress/abilities-api/issues/62

For the rest of potential meta annotations, it would be helpful to finish a filtering capability for ability registration so developers can inject some of these settings:
- https://github.com/WordPress/abilities-api/pull/74

File: docs/getting-started/installation.md (hunk 6 above)
Comment (galatanovidiu) [reply docs/getting-started/installation.md:125]: Yes, we do that automatically
https://github.com/WordPress/mcp-adapter/pull/48/files#diff-c7fb45d5a3e3afff7d0d365aa4b6793f1dc32c5a3e34f8e07062b3c42d6ca031R248

I like the idea of having an `Advanced Usage` section

File: docs/guides/creating-abilities.md (hunk 7 above)
Comment (galatanovidiu) [reply docs/guides/creating-abilities.md]: Another solution will be to remove the concept of resources for now (both on abilities-api and on the MCP adaptor) and implement it in a future version. 

The resources, in my opinion, are a nice-to-have feature, but not a deal breaker.

In MCP, the resources are not things that an agent will query/execute, or choose when to use them. They are static content that a user can attach to the prompt if and when they want. At least this is how it is implemented on all the MCP clients I've tried.

Comment (gziolo) [reply docs/guides/creating-abilities.md]: Yes, starting only with tools is also a viable strategy to start with 👍🏻

File: docs/guides/error-handling.md (hunk 10 above)
Comment (galatanovidiu) [reply docs/guides/error-handling.md:93]: Yes, they are defined on `McpErrorFactory`

File: docs/guides/observability.md
Comment (galatanovidiu) [reply docs/guides/observability.md:1]: I've added the missing implementation examples

File: docs/architecture/overview.md (hunk 3 above)
Comment (galatanovidiu) [reply docs/architecture/overview.md:170]: I've removed that description as it was unclear and unnecessary.

File: docs/architecture/overview.md (hunk 2 above)
Comment (Jameswlepage) [reply docs/architecture/overview.md]: Noting - It will be [publicized alongside](https://make.wordpress.org/core/2025/07/28/roadmap-to-6-9/) the 6. 9 release as a Core package offered by this team. So I would agree in removing these as well.

File: docs/guides/creating-abilities.md (hunk 7 above)
Comment (Jameswlepage) [reply docs/guides/creating-abilities.md]: @gziolo @galatanovidiu What's the most current thinking here? 
> The resources, in my opinion, are a nice-to-have feature, but not a deal breaker.

I'd agree with this.

File: docs/getting-started/installation.md (hunk 6 above)
Comment (Jameswlepage) [reply docs/getting-started/installation.md:125]: I agree that the default implementation and basic usage should Have one MCP server, especially because they will solve the grouping question at some point with future protocol versions. 

But at the same time, I don't think we should bury the fact that there can be multiple MCP implementations. I'm thinking specifically in the WooCommerce context. 

It makes a lot of sense for Woo to have its own server, specifically aimed at agentic commerce, as an example. In fact, that's almost a necessity, and I could see that being the case with other plugins as well.

Comment (Jameswlepage) [reply docs/getting-started/installation.md:125]: I like the approach of a global out of the box server. I mentioned using the layered approach as that fallback server somewhere, and still think that's a great thing to pursue.

File: tests/Unit/Handlers/ToolsHandlerCallTest.php [hunk 11]
```diff
@@ -19,21 +17,6 @@ public static function set_up_before_class(): void {
...
-			'0.0.1',
-			array(),
-			DummyErrorHandler::class,
-			DummyObservabilityHandler::class,
-			$tools,
-		);
-	}
-
 	public function test_missing_name_returns_missing_parameter_error(): void {
 		$server  = $this->makeServer( array( 'test/always-allowed' ) );
```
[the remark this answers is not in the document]
galatanovidiu: Problem resolved

tests/Unit/ErrorHandlers/ErrorEnvelopeTest.php
```diff
@@ -22,7 +22,7 @@ public function test_missing_parameter_error(): void {
 		$err = McpErrorFactory::missing_parameter( 123, 'test_param' );
 
 		$this->assertSame( 123, $err['id'] );
-		$this->assertSame( McpErrorFactory::MISSING_PARAMETER, $err['error']['code'] );
+		$this->assertSame( McpErrorFactory::INVALID_PARAMS, $err['error']['code'] );
```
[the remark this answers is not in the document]
galatanovidiu: Problem resolved

docs/guides/creating-abilities.md
galatanovidiu: My current thinking here is to add the `meta.mcp` parameters, even if we introduce breaking changes.

For those who already use mcp-adapter v0.1.0, we can create migration documentation covering:
- How to update existing abilities with the new meta.mcp.type structure
- Changes needed for custom server definitions
- Migration path from inference-based type determination to explicit type declaration

For abilities-api default abilities, this is also very important if we want them exposed as MCP features.

If these abilities will be included in WP 6.9, it's critical to build them the right way from the beginning. Once they're in core, changing the metadata structure becomes much more difficult due to backward compatibility requirements.

Since we're still at v0.1.0, now is the right time to make this architectural improvement rather than being stuck with parameter inference forever. The explicit type declaration will make the codebase much clearer and more maintainable going forward.

docs/architecture/overview.md
galatanovidiu: Sure, I can do this.
It will be a breaking change, but we might have more breaking changes. See my other comment [here](https://github.com/WordPress/mcp-adapter/pull/48#discussion_r2394131211)

JasonTheAdams:
@galatanovidiu Given all the testing that's been done here and since we want to view 0.x as an iterative alpha, I'm going to suggest that we remove the deprecated code here to remove technical debt and merge this. Let me know once you've cleaned up the deprecated things and I'll give it a list quick review and approve. 👍

docs/guides/default-server.md
galatanovidiu: Done

docs/guides/custom-transports.md
galatanovidiu: Done

docs/architecture/overview.md
galatanovidiu: Done

galatanovidiu:
> @galatanovidiu Given all the testing that's been done here and since we want to view 0.x as an iterative alpha, I'm going to suggest that we remove the deprecated code here to remove technical debt and merge this. Let me know once you've cleaned up the deprecated things and I'll give it a list quick review and approve. 👍

@JasonTheAdams  Done

docs/getting-started/basic-examples.md
```diff
@@ -92,491 +91,189 @@ add_action( 'abilities_api_init', function() {
...
+                ]
+            ],
+            'annotations' => [
+                'readOnlyHint' => true,
+                'idempotentHint' => true
+            ],
+            'mcp' => [
+                'public' => true,   // Expose this ability via MCP
+                'type'   => 'prompt' // Mark as prompt for auto-discovery
+            ]
```
JasonTheAdams, line 225: Are we relying on this structure in this PR? Or just documenting the intended structure?

docs/guides/default-server.md
```diff
@@ -0,0 +1,371 @@
...
+    'resources'              => array(),
+    'prompts'                => array(),
+);
+```
+
+## Core Abilities
+
+The default server includes three core abilities that provide MCP functionality:
+
+### Layered Tooling Architecture
```
JasonTheAdams, line 40: As a note, we'll want to update this doc in a subsequent PR that adds support for categories into layering.

includes/Abilities/DiscoverAbilitiesAbility.php
JasonTheAdams, line 1: Huh. Do we actually want to register the layering tools as abilities? It feels like we shouldn't touch the abilities with something that's exclusively for how we structure the MCP server. I figured we'd manually register these tools apart from the Abilities API.

includes/Cli/McpCommand.php
```diff
@@ -0,0 +1,189 @@
...
+	 * @when after_wp_load
+	 * @synopsis [--server=<server-id>] [--user=<id|login|email>]
+	 */
+	public function serve( array $args, array $assoc_args ): void {
+
+		$enable_serve = apply_filters( 'mcp_adapter_enable_stdio_transport', true );
+
+		if ( ! $enable_serve ) {
+			\WP_CLI::error( 'The STDIO transport is disabled. Enable it by setting the "mcp_adapter_enable_stdio_transport" filter to true.' );
+		}
```
JasonTheAdams: Should this be examined within the transport itself? Or at another layer? Feels like this check shouldn't be exclusive to WP CLI.

includes/Core/McpComponentRegistry.php
```diff
@@ -0,0 +1,397 @@
...
+	public function __construct(
+		McpServer $mcp_server,
+		McpErrorHandlerInterface $error_handler,
+		string $observability_handler,
+		bool $mcp_validation_enabled
+	) {
+		$this->mcp_server    = $mcp_server;
+		$this->error_handler = $error_handler;
+		/** @var class-string<\WP\MCP\Infrastructure\Observability\Contracts\McpObservabilityHandlerInterface> $observability_handler */
+		$this->observability_handler  = $observability_handler;
```
JasonTheAdams, line 98: I'm curious why we're passing the FQCN instead of an instance of the handler?

JasonTheAdams (commented):
I glanced through a bunch of the code with the mind that we'll iterate more later. I have a few questions for you before proceeding, @galatanovidiu!

docs/getting-started/basic-examples.md
galatanovidiu: Yes, I'm relying on this structure. This resolves the "_tool/resource/prompt is determined by inferring it from the parameters_" problem.

- The `public` parameter will determine if the ability will be used on the default MCP server, or not.
- The `type` parameter defines the kind of ability being registered, in this case, 'prompt' enables auto-discovery as a prompt-type ability.

This makes the behavior explicit instead of relying on implicit inference.

@gziolo what do you think about this

includes/Abilities/DiscoverAbilitiesAbility.php
galatanovidiu: I had this discussion with @Jameswlepage 
My first implementation utilized custom tools, but I've since modified the functionality to use the abilities.
This makes the implementation cleaner.

includes/Core/McpComponentRegistry.php
galatanovidiu: The observability handler is passed as an FQCN because it implements only static methods `record_event()`, `record_timing()`.

We can review both observability and error handlers and modify them in subsequent PR's.

includes/Cli/McpCommand.php
galatanovidiu: You're absolutely right! The check should be in StdioServerBridge itself, not exclusive to WP-CLI. This would match the pattern we use in HttpTransport, where permission checks are part of the transport layer.

I will look into this

docs/getting-started/basic-examples.md
gziolo: For MCP specific behaviors, this structure makes perfect sense 👍🏻 

In the Abilities API, there is already a basic handling that helps to distinguish `resource` from `tool` based on `meta.type`:

https://github.com/WordPress/abilities-api/blob/80f47ee41bc8c92bde11b59941c001c617513cf6/packages/client/src/api.ts#L179-L180

https://github.com/WordPress/abilities-api/blob/80f47ee41bc8c92bde11b59941c001c617513cf6/includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php#L95

It's mostly to choose between `GET` vs `POST` request method. My current thinking is that we should generalize that differently using hints as discussed in:
- https://github.com/WordPress/abilities-api/issues/62

This way, a prompt would be marked as `hints.readOnly`, which would resolve to `GET` method internally for handling abilities with REST API, and `mpc.type` set to `prompt` would provide additional details for the MCP adapter. That ensures there is no duplication between Abilities API and MCP for deciding on `tool` vs `resource` vs `prompt` which matters mostly for MCP today.

includes/Abilities/DiscoverAbilitiesAbility.php
gziolo: I left a comment https://github.com/WordPress/mcp-adapter/pull/48/files#r2368119081 suggesting that we consider creating a local abilities registry for these 3 top-layer tools, and keep it referenced locally in MCP. Nearly the same way as in unit tests:

```php
$registry = new WP_Abilities_Registry();
$registry->register( 'mcp-adapter/discover-abilities', $args );
// etc..
```

However, I'm still not entirely sure how it fits in the current architecture.

docs/getting-started/basic-examples.md
gziolo: I opened a PR to better illustrate my feedback:
- https://github.com/WordPress/abilities-api/pull/99

includes/Abilities/DiscoverAbilitiesAbility.php
galatanovidiu: I will give this a try

docs/getting-started/basic-examples.md
galatanovidiu: It would be great to leverage the same metadata in MCP, but we need to address a fundamental terminology mismatch first.

The core issue: WordPress Abilities API and MCP use "tools" and "resources" to mean completely different things:

**WordPress Abilities API:**

Tools = callable functions that modify data
Resources = callable functions that read data

**MCP:**

Tools = callable functions that the AI invokes automatically when needed
Resources = static context that users manually attach to conversations

So WordPress's "tool vs resource" distinction doesn't map to MCP's "tool vs resource"

**Concrete example:** A `site-info` ability could be implemented as:

MCP tool: AI calls it when it needs current site data during conversation
MCP resource: User pre-attaches site info as reference context

The choice depends on the usage pattern (AI-driven vs. user-driven), not the type of action.

gziolo: That's yet another reason to stop using `tools` and `resources` nomenclature inside the Abilities API and leave it for MCP. In fact, it's what the PR I shared is trying to achieve using annotations instead.

includes/Cli/McpCommand.php
galatanovidiu: Changes done. Please check

includes/Core/McpComponentRegistry.php
galatanovidiu: OK, so I've updated the observability handler to be used as an instance, along with other improvements related to observability. Please review https://github.com/WordPress/mcp-adapter/pull/48/commits/03d3b87c1636289a2da2fcec7ef1df33ac9c9b89

includes/Servers/DefaultServerFactory.php
```diff
@@ -0,0 +1,131 @@
...
+			$ability_name = $ability->get_name();
+			$meta         = $ability->get_meta();
+
+			// Skip if not publicly exposed
+			if ( ! ( $meta['mcp']['public'] ?? false ) ) {
+				continue;
+			}
+
+			// Skip mcp-adapter internal abilities
+			if ( str_starts_with( $ability_name, 'mcp-adapter/' ) ) {
```
galatanovidiu: `f ( str_starts_with( $ability_name, 'mcp-adapter/' ) ) {` can be removed by ensuring mcp-adapter internal abilities don't set `mcp.public = true`. `if ( ! ( $meta['mcp']['public'] ?? false ) ) {` would then filter them out automatically, eliminating the need for an explicit namespace-based exclusion.

includes/Abilities/DiscoverAbilitiesAbility.php
```diff
@@ -0,0 +1,160 @@
...
+		wp_register_ability(
+			'mcp-adapter/discover-abilities',
+			array(
+				'label'               => 'Discover Abilities',
+				'description'         => 'Discover all available WordPress abilities in the system. Returns a list of all registered abilities with their basic information.',
+				'category'            => 'mcp-adapter',
+				'input_schema'        => array(
+					'type'                 => 'object',
+					'additionalProperties' => false,
+				),
```
JasonTheAdams: Wait, so the input is an object with no properties? 😕 

Doesn't that just mean there is no input? Why have an `input_schema` at all?

includes/Abilities/DiscoverAbilitiesAbility.php
```diff
@@ -0,0 +1,160 @@
...
+					'properties' => array(
+						'abilities' => array(
+							'type'  => 'array',
+							'items' => array(
+								'type'       => 'object',
+								'properties' => array(
+									'name'        => array( 'type' => 'string' ),
+									'label'       => array( 'type' => 'string' ),
+									'description' => array( 'type' => 'string' ),
+								),
```
JasonTheAdams, line 49: I don't necessarily think we need to if it's working, but something we can explore doing more of in things like properties is adding `description` and `examples` to the individual properties' schema to better inform the AI.

includes/Abilities/DiscoverAbilitiesAbility.php
```diff
@@ -0,0 +1,160 @@
...
+	 * @return array Array containing public MCP abilities.
+	 */
+	public static function execute( $input = array() ): array {
+		// Enforce security checks before execution
+		$permission_check = self::check_permission( $input );
+		if ( is_wp_error( $permission_check ) ) {
+			return array(
+				'error' => $permission_check->get_error_message(),
+			);
+		}
```
JasonTheAdams, line 126: Doesn't the Abilities API already run the authentication check when executing?

includes/Abilities/ExecuteAbilityAbility.php
```diff
@@ -0,0 +1,222 @@
...
+	 *
+	 * @return bool|\WP_Error True if the user has permission to execute the specified ability.
+	 * @phpstan-return bool|\WP_Error
+	 */
+	public static function check_permission( $input = array() ) {
+		$ability_name = $input['ability_name'] ?? '';
+
+		if ( empty( $ability_name ) ) {
+			return new \WP_Error( 'missing_ability_name', 'Ability name is required' );
+		}
```
JasonTheAdams, line 96: Why do we need this if the schema says this is required? Same question below in execution.

includes/Abilities/ExecuteAbilityAbility.php
```diff
@@ -0,0 +1,222 @@
...
+				'error'   => $permission_check->get_error_message(),
+			);
+		}
+
+		if ( ! $permission_check ) {
+			return array(
+				'success' => false,
+				'error'   => 'Permission denied for ability execution',
+			);
+		}
```
JasonTheAdams, line 188: Hmm, I don't love the idea that abilities have to check their own permissions in special circumstances. If someone programmatically bypasses the permission check to run an ability, then that should just work. They can obviously otherwise bypass it via a filter.

includes/Abilities/GetAbilityInfoAbility.php
```diff
@@ -0,0 +1,191 @@
...
+				'label'               => 'Get Ability Info',
+				'description'         => 'Get detailed information about a specific WordPress ability including its input/output schema, description, and usage examples.',
+				'category'            => 'mcp-adapter',
+				'input_schema'        => array(
+					'type'                 => 'object',
+					'properties'           => array(
+						'ability_name' => array(
+							'type'        => 'string',
+							'description' => 'The full name of the ability to get information about',
+						),
```
JasonTheAdams, line 44: We call this the Ability ID, internally, not the name. Probably a good idea to be consistent with this as not to confuse the AI.

includes/Core/McpComponentRegistry.php
```diff
@@ -0,0 +1,438 @@
...
+	}
+
+	/**
+	 * Register resources to the server.
+	 *
+	 * @param array $abilities Array of ability names to convert to MCP resources.
+	 *
+	 * @return void
+	 */
+	public function register_resources( array $abilities ): void {
```
JasonTheAdams, line 221: I don't think it's worth stopping this PR to refactor, but later I'd like to decouple this registry from the Abilities API. That is, let things such as tools and resources be independently registered, and have another class that takes the abilities, converts, and registers them as tools

includes/Core/McpComponentRegistry.php
JasonTheAdams, line 1: One thing I'm a bit confused about with this registry is how it's different from the MCP Server itself? Isn't the MCP Server a registry in its own right for the resources, tools, and prompts? What makes it worth having a distinct registry?

JasonTheAdams: Oh, it looks like each MCP Server _has_ a distinct registry, is that right?

includes/Core/McpTransportFactory.php
```diff
@@ -0,0 +1,117 @@
...
+						/* translators: %s: Transport class name */
+						esc_html__( 'MCP transport class "%s" must implement the McpTransportInterface.', 'mcp-adapter' ),
+						esc_html( $mcp_transport )
+					)
+				);
+			}
+
+			// Interface-based instantiation with dependency injection
+			$context = $this->create_transport_context();
+			new $mcp_transport( $context );
```
JasonTheAdams, line 85: So the transport instances aren't stored anywhere? If this is a factory it should return the transport.

includes/Domain/Resources/McpResource.php
```diff
@@ -344,7 +344,8 @@ public function to_array(): array {
 	 * @return string
 	 */
 	public function to_json(): string {
-		return wp_json_encode( $this->to_array() );
+		$json = wp_json_encode( $this->to_array() );
+		return false !== $json ? $json : '{}';
```
JasonTheAdams, line 348: Do we _want_ to fail silently, here?

includes/Handlers/Initialize/InitializeHandler.php
```diff
@@ -45,27 +45,13 @@ public function handle( int $request_id = 0 ): array {
...
-			'completion' => new stdClass(),
-			'roots'      => array(
-				'list'        => true,
-				'listChanged' => true,
-			),
+			'tools'       => new stdClass(), // Empty object indicates support
+			'resources'   => new stdClass(), // Basic resources support without listChanged/subscribe
+			'prompts'     => new stdClass(), // Basic prompts support without listChanged
+			'logging'     => new stdClass(), // Server supports sending log messages to client
+			'completions' => new stdClass(), // Server supports argument autocompletion (note: plural!)
```
JasonTheAdams, line 54: I'm curious why we're using `stdClass` instead of arrays?

includes/Infrastructure/ErrorHandling/McpErrorFactory.php
```diff
@@ -281,6 +322,75 @@ public static function unauthorized( int $id, string $details = '' ): array {
...
+	 */
+	public static function mcp_error_to_http_status( $mcp_error_code ): int {
+		// Handle integer error codes (existing logic)
+		switch ( $mcp_error_code ) {
+			// Transport-level errors - these indicate malformed requests
+			case self::PARSE_ERROR:      // Invalid JSON - syntactic error
+				return 400;
+
+			case self::INVALID_REQUEST:  // Invalid JSON-RPC structure - syntactic error
+				return 400;
```
JasonTheAdams, line 344:
```suggestion
			case self::PARSE_ERROR:      // Invalid JSON - syntactic error
			case self::INVALID_REQUEST:  // Invalid JSON-RPC structure - syntactic error
				return 400;
```

includes/Servers/DefaultServerFactory.php
```diff
@@ -0,0 +1,126 @@
...
+
+		// Ensure config is an array and merge with defaults
+		if ( ! is_array( $config ) ) {
+			$config = $wordpress_defaults;
+		}
+		$config = wp_parse_args( $config, $wordpress_defaults );
+
+		// Use McpAdapter to create the server with full validation
+		$adapter = McpAdapter::instance();
+		$adapter->create_server(
```
JasonTheAdams, line 74: Another place where the factory should be returning the instance, not just losing it in memory where it's only reference is in hooks.

includes/Transport/Infrastructure/HttpRequestHandler.php
```diff
@@ -0,0 +1,280 @@
...
+
+		// Handle GET requests (listening for messages from server via SSE)
+		if ( 'GET' === $context->method ) {
+			return $this->handle_sse_request( $context );
+		}
+
+		// Handle DELETE requests (session termination)
+		if ( 'DELETE' === $context->method ) {
+			return $this->handle_session_termination( $context );
+		}
```
JasonTheAdams, line 60: I recommend using either `HttpRequestContext` constants for comparing the method and/or providing declarative methods such as `$context->is_post_method()`.

This way we're not using `'GET'` and such here which assumes normalization in the context.

includes/Transport/Infrastructure/RequestRouter.php
```diff
@@ -0,0 +1,245 @@
...
+			);
+			$this->context->observability_handler->record_event( 'mcp.request', $tags, $duration );
+
+			// Create error response from exception.
+			return array( 'error' => McpErrorFactory::internal_error( $request_id, 'Handler error occurred' )['error'] );
+		}
+	}
+
+	/**
+	 * Add nextCursor for backward compatibility with existing API.
```
JasonTheAdams, line 133: I'm not quite sure what this is referring to, but can we just leave it since we're not worrying about BC at this point?

includes/Transport/Infrastructure/SessionManager.php
```diff
@@ -0,0 +1,294 @@
...
+	 * @var int
+	 */
+	private const MAX_SESSIONS = 32;
+
+	/**
+	 * Session inactivity timeout in seconds (24 hours).
+	 *
+	 * @var int
+	 */
+	private const INACTIVITY_TIMEOUT = DAY_IN_SECONDS;
```
JasonTheAdams: Let's prefix these with `DEFAULT_` to make it clearer.

includes/Transport/Infrastructure/SessionManager.php
JasonTheAdams, line 1: I think this class could benefit from a `Session` DTO class. Things such as checking inactivity timeouts, comparing creation times, and so forth can become declarative and independently testable.

includes/Transport/Infrastructure/SessionManager.php
```diff
@@ -0,0 +1,294 @@
...
+		// Get current sessions
+		$sessions = self::get_all_user_sessions( $user_id );
+
+		// Check session limit - remove oldest if over limit
+		$config       = self::get_config();
+		$max_sessions = $config['max_sessions'];
+		if ( count( $sessions ) >= $max_sessions ) {
+			// Remove oldest session (FIFO)
+			$oldest_session_id = null;
+			$oldest_time       = PHP_INT_MAX;
```
JasonTheAdams: Why are we setting this? How would a session's `created_at` ever be greater than the greatest integer?

includes/Transport/HttpTransport.php
```diff
@@ -0,0 +1,131 @@
...
+	protected HttpRequestHandler $request_handler;
+
+	/**
+	 * Initialize the class and register routes
+	 *
+	 * @param \WP\MCP\Transport\Infrastructure\McpTransportContext $transport_context The transport context.
+	 */
+	public function __construct( McpTransportContext $transport_context ) {
+		$this->request_handler = new HttpRequestHandler( $transport_context );
+		add_action( 'rest_api_init', array( $this, 'register_routes' ), 16 );
```
JasonTheAdams, line 43: Can we move this to an `initialize()` method or somewhere else? I think it's better to not have hooks as a side-effect of constructors.

JasonTheAdams (changes requested):
Ok, @galatanovidiu! So I just went through every non-test file and left a bunch of questions/comments. I want to be very clear on my intent:
* I tried to keep most suggestions minor
* **Feel free** to defer feedback to subsequent PRs as you see fit
* This is a lot of great work!

One other, broader observation, is that some of these classes have a massive amount of parameters. Since we're not in 8.x land and can't use named parameters, it may be worth using a `class::from_array()` type method instead, so we can pass an associative array.

includes/Abilities/DiscoverAbilitiesAbility.php
galatanovidiu: Let's address this in a follow-up issue/PR

includes/Abilities/ExecuteAbilityAbility.php
galatanovidiu: Let's address this in a follow-up issue/PR

includes/Abilities/GetAbilityInfoAbility.php
galatanovidiu: Let's address this in a follow-up issue/PR

includes/Core/McpComponentRegistry.php
galatanovidiu: Let's address this in a follow-up issue/PR

includes/Core/McpComponentRegistry.php
galatanovidiu: yes, that is right

includes/Core/McpTransportFactory.php
galatanovidiu: Yes, you are right.
This is just an Initializer.
Let's address this in a follow-up issue/PR and see if we want to keep the name and refactor this into a factory or just rename the class

includes/Handlers/Initialize/InitializeHandler.php
galatanovidiu: The MCP clients are expecting this to be objects, not arrays.
We can keep them like this or use something like `(object)array()`

includes/Domain/Resources/McpResource.php
galatanovidiu: Probably not. We can address this in a follow-up issue/PR

includes/Servers/DefaultServerFactory.php
galatanovidiu: Yes, you are right.
This is just an Initializer. 
Let's address this in a follow-up issue/PR and see if we want to keep the name and refactor this into a factory, or just rename the class

includes/Transport/Infrastructure/HttpRequestHandler.php
galatanovidiu: Let's address this in a follow-up issue/PR

includes/Abilities/DiscoverAbilitiesAbility.php
galatanovidiu: Done [4837b59](https://github.com/WordPress/mcp-adapter/pull/48/commits/4837b5931c5f64d51a79c6fdabffdf0291841b46)

includes/Transport/Infrastructure/SessionManager.php
galatanovidiu: Done [d5f8c32](https://github.com/WordPress/mcp-adapter/pull/48/commits/d5f8c3233c3c9c59cb72552c661357b7e2de5fad)

includes/Transport/Infrastructure/SessionManager.php
galatanovidiu: Let's address this in a follow-up issue/PR

JasonTheAdams (approved):
Given our plan to merge this and iterate further with much of the feedback, I'll go ahead and approve this. I haven't had a chance, yet, to thoroughly test this, so please do so before we mark a 0.3.0 release.

includes/Transport/Infrastructure/SessionManager.php
galatanovidiu: We initialize $oldest_time to PHP_INT_MAX as a sentinel value. This ensures that any session’s created_at timestamp will be smaller, allowing us to correctly identify the oldest session during iteration.

Anyway, I’ve refactored the code to use a sorting-based approach, which simplifies the logic.

[d2abe84](https://github.com/WordPress/mcp-adapter/pull/48/commits/d2abe840ea45e23f8fd2cbf7f19315f2843b6ba8)

includes/Transport/HttpTransport.php
galatanovidiu: Let's address this in a follow-up issue/PR

includes/Transport/Infrastructure/SessionManager.php
JasonTheAdams: I'm glad you did! The sorting method is much cleaner. 😄

PR #71, opened 2025-10-21
Refactor error handling: Replace exceptions with WP_Error pattern

### LINKED ISSUE #24: Reduce use of `throw {ExceptionType}` in favor of WordPress error handling
## What

We should review the current usage of `throw *` and (when appropriate) handle the error or return `WP_Error` objects instead.

## Why

To butcher the whole "Exceptions should be exceptional" argument, using Exceptions for control flow is an antipattern (in most places but especially ) in WordPress-land.

While in some cases, throwing may be the optimal way to centralize and "abstract" our error handling, more often it leads to `try {} catch {}` pollution additional doc requirements, and overall more dx complication - especially in the WordPress crowd.

## How

Starting top level (`WP/MCP/Core`) and working inwards, we review where we can directly return a `new WP_Error()` (and what try-catches we can remove as a result ). This should allow us to review and remediate things incrementally without worrying about disrupting the existing lifecycle while still help prevent us leaking new exceptions in future commits.

galatanovidiu:
## Why

The plugin was using PHP exceptions for error handling, which contradicts WordPress conventions where `WP_Error` is the standard pattern.

## What

Refactored all error handling from exceptions to WordPress's native `WP_Error` pattern across the entire plugin.

### Changes Made

**Core Layer** (`includes/Core/`)
- `McpAdapter::create_server()` now returns `self|WP_Error` (removed 5 exception throws)
- `McpComponentRegistry` removed 4 try-catch blocks, now uses `is_wp_error()` checks
- `McpTransportFactory` logs errors using `_doing_it_wrong` and continues instead of throwing

**Domain Layer** (`includes/Domain/`)
- All `get_ability()` methods return `WP_Ability|WP_Error` instead of throwing
- All validators (`McpToolValidator`, `McpResourceValidator`, `McpPromptValidator`) return `bool|WP_Error`
- All `validate()` methods return `self|WP_Error`
- All registration classes propagate `WP_Error` instead of throwing exceptions

**Handlers Layer** (`includes/Handlers/`)
- Added `is_wp_error()` checks after `get_ability()` calls
- Preserved try-catch for `\Throwable` to catch unexpected PHP errors

**Testing**
- Added `assertWPError()` helper to TestCase
- Updated 37 tests to expect `WP_Error` instead of exceptions

### Benefits

- Follows WordPress core conventions
- Removed 14+ try-catch blocks
- Better error recovery with graceful degradation
- Enhanced observability with specific error codes
- Zero breaking changes

Closes #24

### DIFF
--- a/includes/Core/McpAdapter.php
+++ b/includes/Core/McpAdapter.php
@@ -102,10 +102,9 @@ final class McpAdapter {
 	 * @param array $prompts Prompts to register.
 	 * @param callable|null $transport_permission_callback Optional custom permission callback for transport-level authentication. If null, defaults to is_user_logged_in().
 	 *
-	 * @return \WP\MCP\Core\McpAdapter
-	 * @throws \Exception If the server already exists or if called outside of the mcp_adapter_init action.
+	 * @return \WP\MCP\Core\McpAdapter|\WP_Error McpAdapter instance on success, WP_Error on failure.
 	 */
-	public function create_server( string $server_id, string $server_route_namespace, string $server_route, string $server_name, string $server_description, string $server_version, array $mcp_transports, ?string $error_handler, ?string $observability_handler = null, array $tools = array(), array $resources = array(), array $prompts = array(), ?callable $transport_permission_callback = null ): self {
+	public function create_server( string $server_id, string $server_route_namespace, string $server_route, string $server_name, string $server_description, string $server_version, array $mcp_transports, ?string $error_handler, ?string $observability_handler = null, array $tools = array(), array $resources = array(), array $prompts = array(), ?callable $transport_permission_callback = null ) {
 		// Use NullMcpErrorHandler if no error handler is provided.
 		if ( ! $error_handler ) {
 			$error_handler = NullMcpErrorHandler::class;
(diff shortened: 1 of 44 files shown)


JasonTheAdams:
I'm _really_ curious about this. Is throwing exceptions truly _never_ the solution in WordPress? I searched the WP code, and found there are [166 PHP files](https://github.com/search?q=repo:WordPress/wordpress-develop+throw+new+language:PHP&type=code&l=PHP&p=2) that contains `throw new` at least once.

I can't find a document on this, so I asked ChatGPT, and it seems to have given [sensible criteria](https://chatgpt.com/share/68f98506-6c50-800c-aa09-41d28c0226b5). Funnily enough, it actually quotes the Abilities API work for some of its reasoning. Hahah!

galatanovidiu:
> I'm _really_ curious about this. Is throwing exceptions truly _never_ the solution in WordPress? I searched the WP code, and found there are [166 PHP files](https://github.com/search?q=repo:WordPress/wordpress-develop+throw+new+language:PHP&type=code&l=PHP&p=2) that contains `throw new` at least once.
> 
> I can't find a document on this, so I asked ChatGPT, and it seems to have given [sensible criteria](https://chatgpt.com/share/68f98506-6c50-800c-aa09-41d28c0226b5). Funnily enough, it actually quotes the Abilities API work for some of its reasoning. Hahah!


@JasonTheAdams  I’m usually in the “throw exceptions” camp, but in this codebase we were mostly throwing, catching immediately, and then logging or returning an MCP error. This refactor is a win: using WP_Error aligns with WP conventions, removes boilerplate try/catch (e.g., in McpComponentRegistry/handlers), and gives us consistent error codes while keeping behavior MCP-spec compliant.

JasonTheAdams (approved):
Thanks for explaining! LGTM!

Issue #53, opened 2025-10-20
Follow-up #48: I think this class could benefit from a `Session` DTO class. Things such as checking inactivity timeouts, comparing creation times, and so forth can become declarative and independently testable.


galatanovidiu:
I think this class could benefit from a `Session` DTO class. Things such as checking inactivity timeouts, comparing creation times, and so forth can become declarative and independently testable.

_Originally posted by @JasonTheAdams in https://github.com/WordPress/mcp-adapter/pull/48#discussion_r2445394124_

JasonTheAdams:
Another one that would be useful here is for handling the MCP responses. Right now we construct the response shape and include `_metadata` along the way. At the last minute, in #72, it's unset from the array. It would be better to have a DTO which distinguishes between the body of the response and any additional information we want to track internally.

PR #72, opened 2025-10-21
Fix metadata appearing in tool response content

galatanovidiu:
## Why
  Internal `_metadata` fields were appearing in the client-facing response, showing up in both the `structuredContent` object and the JSON-encoded `content[0].text` field.
   This metadata is intended only for internal observability tracking and should not be exposed to MCP clients.

  ## What
  - Extract `_metadata` from the result **before** adding it to response content/structuredContent
  - Add `unset($result['_metadata'])` to ensure metadata is removed from the result
  - Reorder operations in `ToolsHandler::call_tool()` to:
    1. Extract metadata first
    2. Remove it from result
    3. Then add cleaned result to response

  This ensures metadata is only present at the top level of the response (where `RequestRouter` removes it) and never reaches the client.

### DIFF
--- a/includes/Handlers/Tools/ToolsHandler.php
+++ b/includes/Handlers/Tools/ToolsHandler.php
@@ -133,6 +133,15 @@ class ToolsHandler {
 				),
 			);
 
+			// Extract and store metadata before adding result to response content.
+			$response['_metadata'] = $result['_metadata'] ?? array(
+				'component_type' => 'tool',
+				'tool_name'      => $request_params['name'],
+			);
+
+			// Remove metadata from result so it doesn't appear in content or structuredContent.
+			unset( $result['_metadata'] );
+
 			// @todo: add support for EmbeddedResource schema.ts:619.
 			if ( isset( $result['type'] ) && 'image' === $result['type'] ) {
 				$response['content'][0]['type'] = 'image';
@@ -144,12 +153,6 @@ class ToolsHandler {
 				$response['content'][0]['text'] = wp_json_encode( $result );
 				$response['structuredContent']  = $result;
 			}
-
-			// Add metadata from result if present, or create basic metadata.
-			$response['_metadata'] = $result['_metadata'] ?? array(
-				'component_type' => 'tool',
-				'tool_name'      => $request_params['name'],
-			);
 
 			return $response;
 		} catch ( \Throwable $exception ) {


includes/Handlers/Tools/ToolsHandler.php
```diff
@@ -133,6 +133,15 @@ public function call_tool( array $message, int $request_id = 0 ): array {
...
 			);
 
+			// Extract and store metadata before adding result to response content.
+			$response['_metadata'] = $result['_metadata'] ?? array(
+				'component_type' => 'tool',
+				'tool_name'      => $request_params['name'],
+			);
+
+			// Remove metadata from result so it doesn't appear in content or structuredContent.
+			unset( $result['_metadata'] );
```
JasonTheAdams, line 143: I'm really confused by this. We're storing the metadata in the response, removing it from the result, and then returning the response (that now has the metadata)? Doesn't that mean we _are_ putting the metadata in the response? 😕

galatanovidiu: Good catch on the confusion - let me clarify the flow:

The code **is** functional, but the metadata handling is indeed tangled. Here's what's happening:

### The Flow

  1. **Deep in the handler** (ability execution at line 310): `$result` comes back from `$ability->execute()` - this is raw data that will become response content
  2. **Metadata extraction** (lines 136-143): We extract any metadata that came from the result, attach it to the response, then **remove it from the result**
  3. **Why remove it?** The `$result` data gets embedded in `response['content']` and `response['structuredContent']` (lines 153-154). We don't want `_metadata` appearing in those fields since they're sent to the MCP client.
  4. **RequestRouter** (RequestRouter.php:87-88): Extracts `response['_metadata']` for observability tracking, then removes it before sending to client

### Why This Way?

Metadata needs to flow from handler → RequestRouter for observability, but must NOT leak to the MCP client.  The current array-based approach requires this manual extraction/removal dance.

### Known Issue

This is tangled code. The proper fix is **Response DTOs**, but that's a larger refactor planned separately on #53 . For now, this works and prevents metadata leakage.

JasonTheAdams: Oh my. Well, I'm glad it works for now! Looking forward to some DTOs! 😅 👍

JasonTheAdams (approved):
Thanks for explaining! This is good to go, then!

PR #78, opened 2025-10-21
Fix prompt parameter validation by converting input_schema to MCP arguments

galatanovidiu:
## Why

Prompts that accept input parameters were failing validation because they lacked the proper MCP arguments format. While abilities use JSON Schema input_schema for validation, MCP prompts require a specific arguments array format for their interface. This mismatch caused validation failures when prompts received input parameters.

## What

  - Add automatic conversion from JSON Schema input_schema to MCP prompt arguments format in RegisterAbilityAsMcpPrompt
  - Implement the convert_input_schema_to_arguments() method that transforms schema properties to MCP arguments
  - Update documentation with comprehensive examples showing the new input_schema approach

This is a breaking change.

### DIFF
--- a/docs/guides/creating-abilities.md
+++ b/docs/guides/creating-abilities.md
@@ -252,19 +252,54 @@ wp_register_ability('my-plugin/site-config', [
 
 ## Creating Prompts
 
-Prompts generate structured messages for language models and support two types of annotations. They should set `type: 'prompt'` in the MCP configuration:
+Prompts generate structured messages for language models. They use `input_schema` to define parameters, which are automatically converted to MCP prompt arguments format. Prompts should set `type: 'prompt'` in the MCP configuration.
 
-### 1. Template-Level Annotations
-Describe the prompt template's behavior characteristics:
+### Input Schema for Prompts
+
+Prompts use standard JSON Schema `input_schema` to define their parameters. The MCP Adapter automatically converts this to the MCP prompt `arguments` format:
+
+```php
+// Your definition (JSON Schema):
+'input_schema' => [
+    'type' => 'object',
+    'properties' => [
+        'code' => ['type' => 'string', 'description' => 'Code to review']
+    ],
+    'required' => ['code']
+]
+
+// Automatically converted to MCP format:
+'arguments' => [
+    ['name' => 'code', 'description' => 'Code to review', 'required' => true]
+]
+```
+
+### Complete Prompt Example
 
 ```php
 wp_register_ability('my-plugin/code-review', [
     'label' => 'Code Review Prompt',
     'description' => 'Generate a code review prompt with specific focus areas',
+    'input_schema' => [
+        'type' => 'object',
+        'properties' => [
+            'code' => [
+                'type' => 'string',
+                'description' => 'Code to review'
+            ],
+            'focus' => [
+                'type' => 'array',
+                'description' => 'Areas to focus on during review',
+                'items' => ['type' => 'string'],
+                'default' => ['security', 'performance']
+            ]
+        ],
+        'required' => ['code']
+    ],
     'execute_callback' => function($input) {
-        $code = $input['code'] ?? '';
+        $code = $input['code'];
         $focus = $input['focus'] ?? ['security', 'performance'];
-        
+
         return [
             'messages' => [
                 [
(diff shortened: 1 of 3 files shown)


docs/guides/creating-abilities.md
```diff
@@ -252,19 +252,54 @@ wp_register_ability('my-plugin/site-config', [
...
+        'code' => ['type' => 'string', 'description' => 'Code to review']
+    ],
+    'required' => ['code']
+]
+
+// Automatically converted to MCP format:
+'arguments' => [
+    ['name' => 'code', 'description' => 'Code to review', 'required' => true]
+]
+```
```
JasonTheAdams, line 275: Is this something we actually should document? Or is it just an internal detail? I could go either way on this.

includes/Domain/Prompts/RegisterAbilityAsMcpPrompt.php
```diff
@@ -89,20 +104,84 @@ private function get_data(): array {
...
+			// Add description if available
+			if ( ! empty( $property_schema['description'] ) ) {
+				$argument['description'] = $property_schema['description'];
+			}
+
+			$arguments[] = $argument;
+		}
+
+		return $arguments;
+	}
```
JasonTheAdams, line 183: Maybe I'm missing it, but where are we setting the "type" and tentative "items"? I'm seeing this true even in the examples. is there a reason for this?

JasonTheAdams (commented):
Good fix! Couple of questions for you, @galatanovidiu!

docs/guides/creating-abilities.md
galatanovidiu: This needs to be documented because **MCP Prompts do not have an input schema**; they have a simpler `arguments` array format per the [MCP specification](https://modelcontextprotocol.io/specification/2025-06-18/server/prompts#protocol-messages).

However, on `prompts/get` calls, the parameters are sent exactly the same way as `tools/call` parameters. Because the Abilities API performs schema validation on these incoming parameters, we must define prompt arguments using the full `inputSchema` format (just like tools) when registering the ability, then the MCP Adapter automatically transforms it to the simpler `arguments` array format in the `prompts/list` response.

**This documentation section should be expanded and clarified, not removed.** Developers need to understand this design decision because it's not immediately obvious why prompt arguments require `inputSchema` in the Abilities API when the MCP spec shows them as simple `arguments`. Without this explanation, developers might incorrectly assume prompts and tools should be defined differently, or miss the validation benefits that the full schema provides during prompt execution.

includes/Domain/Prompts/RegisterAbilityAsMcpPrompt.php
galatanovidiu: Good question! As mentioned in my earlier comment, this conversion intentionally omits type and items because the MCP specification for prompt arguments (used in prompts/list responses) only includes three fields:
  - name: the argument identifier
  - description: explanation of the argument's purpose
  - required: boolean indicating if it's mandatory

  This code converts explicitly the JSON Schema input_schema to a simpler MCP prompt arguments format for the prompts/list response.

However, the full input_schema (including type, items, etc.) is still preserved in the ability registration and is used for validation when the prompt is actually executed via the prompts/get method. That's the key design decision here: we register prompts with full JSON Schema in the Abilities API (for validation during execution), but expose them with the simpler arguments array format in the MCP prompts/list response (per the MCP spec).

So this method is doing exactly what it should: stripping the type information for the MCP client-facing API while keeping it available for server-side validation. The type and items are never set in the arguments array because MCP clients don't expect them; they only expect the name, description, and required fields.

JasonTheAdams: You're absolutely right. I was looking at an example that was for prompt templating, not the prompt spec. 👍

docs/guides/creating-abilities.md
JasonTheAdams: I see what you mean. It could be confusing for someone using this adapter to understands MCP. I don't know how often that will be the case, I but I think it's fair. Thanks!

JasonTheAdams (approved):
Thank you for answering my questions and clearing things up for me, @galatanovidiu!

PR #86, opened 2025-11-06
Ensures init action fires before initializing in WP-CLI

galatanovidiu:
Ensures that the `init` action has fired before initializing the adapter in WP-CLI context.

This prevents potential issues where commands might not have access to servers if the adapter initializes too early. It now hooks into the `init` action if it hasn't already fired.


JasonTheAdams (approved):
Good catch! LGTM! 🎉

Issue #87, opened 2025-11-06
Getting Fatal error for check_permission()


Ananya-Cedcoss:
This is a method signature incompatibility error in plugin. Here's what's happening:
The Issue
You have two interfaces with conflicting check_permission() method signatures:

McpTransportInterface::check_permission() — expects no parameters
McpRestTransportInterface::check_permission(WP_REST_Request $request) — expects a WP_REST_Request parameter

When McpRestTransportInterface tries to implement/extend McpTransportInterface, PHP requires that the child method have a compatible signature with the parent. The child can't add required parameters that the parent doesn't have.

here is the exact error i am getting
Fatal error: Declaration of WP\MCP\Transport\Contracts\McpRestTransportInterface::check_permission(WP_REST_Request $request) must be compatible with WP\MCP\Transport\Contracts\McpTransportInterface::check_permission() in /home/cedcoss/Local Sites/new-update-of-woocommerce/app/public/wp-content/plugins/mcp-adapter-trunk/includes/Transport/Contracts/McpRestTransportInterface.php on line 29

stefanodipietro:
Same problem here. Following.

mindctrl:
The files in this repo look correct. `McpTransportInterface::check_permission()` no longer exists, so won't cause the error. Are you running another plugin that's consuming this as a dependency?

nikolas4175-godaddy:
I noticed the same and on my site it seems to be a conflict with latest versions of WooCommerce (10.3.x), which [include mcp-adapter](https://github.com/woocommerce/woocommerce/blob/4674fce99ce8227c5b99ed02398a26f1c8de3d48/plugins/woocommerce/composer.json#L60) as a composer dependency, specifically pinned to 7a2d22cff92328bc94f5b1648a66ae4273e949c5. The errors are not present on WC v10.2.2.

JasonTheAdams:
@galatanovidiu Is this resolved by #91?

galatanovidiu:
The issue was addressed in #48. 

However, as noted above, the mcp‑adapter (in the version used by WooCommerce v10.3) still ships an older adapter version that does not include the updated contract/method signature changes. Because of that mismatch, the method-signature incompatibility you’re seeing occurs (i.e., check_permission() with and without the WP_REST_Request parameter).  ￼

The new WooCommerce version, to be released on December 9th, will include an updated mcp-adapter that resolves this mismatch.

[...]

PR #91, opened 2025-11-10
Fix/tool annotations mapping

### LINKED ISSUE #70: Adapt abilities-api annotations format to comply with MCP ToolAnnotations specification
## Summary

The mcp-adapter needs to properly convert abilities-api annotations to the https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations format.

Currently, the conversion in RegisterAbilityAsMcpTool passes through annotations unchanged, which doesn't match the MCP field naming requirements.

### Current State

 The abilities-api currently defines annotations in WP_Ability::$default_annotations ([class-wp-ability.php:38-53](https://github.com/WordPress/abilities-api/blob/trunk/includes/abilities-api/class-wp-ability.php#L38-L53)):

```php
  array(
      'instructions' => '',      // string: Instructions on how to use the ability
      'readonly'     => false,   // bool: Ability does not modify environment
      'destructive'  => true,    // bool: May perform destructive updates
      'idempotent'   => false,   // bool: Same arguments have no additional effect
  )
```

###  MCP ToolAnnotations Specification

According to the MCP specification, ToolAnnotations should be:

```ts
  interface ToolAnnotations {
    destructiveHint?: boolean;  // Default: true
    idempotentHint?: boolean;   // Default: false
    readOnlyHint?: boolean;     // Default: false
  }
```

## Questions to Resolve

1. Instructions field: The instructions field has no equivalent in MCP ToolAnnotations. MCP uses the tool description field for instructions. Should we:
    - Prepend the instructions to the MCP tool description?
    - Ignore the instructions field entirely?
    - Pass it as a non-standard annotation field (may cause validation issues)?
2. I presume we want all other annotations to be passed to MCP unchanged?

@JasonTheAdams, @justlevine @swissspidy ...

galatanovidiu:
## Why

The adapter was passing annotations through unchanged, causing field name mismatches between WordPress Abilities API format and MCP specification. WordPress uses `readonly`, `destructive`, and `idempotent`, while MCP requires `readOnlyHint`, `destructiveHint`, and `idempotentHint`. This caused protocol compliance issues and could break MCP clients expecting the correct field names.

Additionally null annotation values should not be included in MCP responses.

## What

Adapts abilities-api annotations format to comply with the MCP ToolAnnotations specification. Adds annotation mapping functions that convert WordPress annotation field names to MCP-compliant format:

- **Tools**: Maps `readonly` → `readOnlyHint`, `destructive` → `destructiveHint`, `idempotent` → `idempotentHint`
- **Resources & Prompts**: Validates and normalizes `audience`, `lastModified`, and `priority` fields according to MCP specification

The implementation includes:

- Annotation mapping methods in `RegisterAbilityAsMcpTool`, `RegisterAbilityAsMcpResource`, and `RegisterAbilityAsMcpPrompt`
- Annotation validation in `McpToolValidator`, `McpResourceValidator`, and `McpPromptValidator` to ensure MCP compliance
- Filtering of null values
- Support for MCP-native fields (`openWorldHint`, `title` for tools)
- Comprehensive test coverage for annotation mapping and validation across all three component types
- Documentation updates explaining the annotation format conversion

This change ensures full compliance with the MCP ToolAnnotations specification while maintaining backward compatibility by accepting both WordPress-format and MCP-native annotation fields.

Closes #70


includes/Domain/Resources/RegisterAbilityAsMcpResource.php
```diff
@@ -88,6 +88,117 @@ public function get_uri() {
...
+				continue;
+			}
+			$priority = (float) $value;
+			// Clamp priority between 0 and 1 per MCP spec.
+			$priority                  = max( 0.0, min( 1.0, $priority ) );
+			$mcp_annotations[ $field ] = $priority;
+		}
+
+		return $mcp_annotations;
+	}
```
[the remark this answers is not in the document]
galatanovidiu: resolved on [06b44eb](https://github.com/WordPress/mcp-adapter/pull/91/commits/06b44eb19045711fdd4bb08f9e83fd002f534b2d)

includes/Domain/Prompts/RegisterAbilityAsMcpPrompt.php
```diff
@@ -79,6 +79,117 @@ private function __construct( WP_Ability $ability, McpServer $mcp_server ) {
...
+				continue;
+			}
+			$priority = (float) $value;
+			// Clamp priority between 0 and 1 per MCP spec.
+			$priority                  = max( 0.0, min( 1.0, $priority ) );
+			$mcp_annotations[ $field ] = $priority;
+		}
+
+		return $mcp_annotations;
+	}
```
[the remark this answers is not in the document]
galatanovidiu: resolved on [06b44eb](https://github.com/WordPress/mcp-adapter/pull/91/commits/06b44eb19045711fdd4bb08f9e83fd002f534b2d)

includes/Domain/Resources/RegisterAbilityAsMcpResource.php
```diff
@@ -88,6 +88,117 @@ public function get_uri() {
...
+
+		foreach ( $formats as $format ) {
+			$datetime = \DateTime::createFromFormat( $format, $timestamp );
+			if ( $datetime && $datetime->format( $format ) === $timestamp ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
```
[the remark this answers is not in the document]
galatanovidiu: resolved on [26dff6f](https://github.com/WordPress/mcp-adapter/pull/91/commits/26dff6fc423b65dba6b447f5d3b2249f7f49d119)

includes/Domain/Prompts/RegisterAbilityAsMcpPrompt.php
```diff
@@ -79,6 +79,117 @@ private function __construct( WP_Ability $ability, McpServer $mcp_server ) {
...
+
+		foreach ( $formats as $format ) {
+			$datetime = \DateTime::createFromFormat( $format, $timestamp );
+			if ( $datetime && $datetime->format( $format ) === $timestamp ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
```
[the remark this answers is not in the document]
galatanovidiu: resolved on [26dff6f](https://github.com/WordPress/mcp-adapter/pull/91/commits/26dff6fc423b65dba6b447f5d3b2249f7f49d119)

includes/Domain/Resources/McpResourceValidator.php
```diff
@@ -153,13 +153,136 @@ public static function get_validation_errors( array $resource_data ): array {
...
+
+		foreach ( $formats as $format ) {
+			$datetime = \DateTime::createFromFormat( $format, $timestamp );
+			if ( $datetime && $datetime->format( $format ) === $timestamp ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
```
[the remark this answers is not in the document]
galatanovidiu: resolved on [26dff6f](https://github.com/WordPress/mcp-adapter/pull/91/commits/26dff6fc423b65dba6b447f5d3b2249f7f49d119)

includes/Domain/Tools/RegisterAbilityAsMcpTool.php
```diff
@@ -59,6 +59,76 @@ private function __construct( WP_Ability $ability, McpServer $mcp_server ) {
...
+		// Convert WordPress-format fields to MCP format and remove old fields.
+		foreach ( $field_mapping as $wp_field => $mcp_field ) {
+			// If WordPress-format field exists convert it.
+			if ( ! isset( $mcp_annotations[ $wp_field ] ) ) {
+				continue;
+			}
+			$mcp_annotations[ $mcp_field ] = (bool) $mcp_annotations[ $wp_field ];
+			// Remove the WordPress-format field from the result.
+			unset( $mcp_annotations[ $wp_field ] );
+		}
```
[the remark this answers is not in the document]
galatanovidiu: resolved on [a324e79](https://github.com/WordPress/mcp-adapter/pull/91/commits/a324e79fd1befbb753fd04b101c81c42e227331d)

docs/guides/creating-abilities.md
```diff
@@ -58,32 +58,116 @@ wp_register_ability('my-plugin/my-ability', [
...
+```
+
+**Tool Annotation Mapping Table:**
+
+| WordPress Format | MCP Format | Description |
+|-----------------|------------|-------------|
+| `readonly` | `readOnlyHint` | Tool doesn't modify data |
+| `destructive` | `destructiveHint` | Tool may delete/destroy data |
+| `idempotent` | `idempotentHint` | Same input → same output |
+| *(no equivalent)* | `openWorldHint` | Can work with arbitrary data |
```
JasonTheAdams, line 102: I'm inclined to make this `open-world` as I could see it being a more broadly supported Ability annotation in the future. Providing the indication that an Ability make be accessing things beyond the website itself is useful.

JasonTheAdams: Upon more reflection, I guess the point is that the user can either provide the MCP version of the annotation, or it will support the WP version if there is one. So if an `open-world` WP version is added in the future, we can add support in a backwards compatible way.

If I'm right here,  @galatanovidiu, then you can disregard. 😄

includes/Domain/Prompts/McpPromptValidator.php
```diff
@@ -129,9 +131,98 @@ public static function get_validation_errors( array $prompt_data ): array {
...
+				continue;
+			}
+
+			// Validate lastModified field.
+			if ( 'lastModified' === $field ) {
+				if ( ! is_string( $value ) || empty( trim( $value ) ) ) {
+					$errors[] = __( 'Annotation field lastModified must be a non-empty string', 'mcp-adapter' );
+					continue;
+				}
+				if ( ! McpValidator::validate_iso8601_timestamp( trim( $value ) ) ) {
```
JasonTheAdams: Should we be trimming at this point, or assuming `$value` has been trimmed upstream?

includes/Domain/Prompts/McpPromptValidator.php
```diff
@@ -129,9 +131,98 @@ public static function get_validation_errors( array $prompt_data ): array {
...
+		foreach ( $annotations as $field => $value ) {
+			// Check if field is a valid MCP annotation field.
+			if ( ! in_array( $field, $valid_fields, true ) ) {
+				$errors[] = sprintf(
+					/* translators: %s: annotation field name */
+					__( 'Unknown annotation field: %s. Valid MCP annotation fields are: audience, lastModified, priority', 'mcp-adapter' ),
+					$field
+				);
+				continue;
+			}
```
JasonTheAdams: I think we can actually remove `$valid_fields` and simply handle this as a last condition after `'priority'`. We're explicitly checking for each type and validating it, effectively making an `if...elseif...else`, where the last condition is an invalid field.

includes/Domain/Prompts/McpPromptValidator.php
```diff
@@ -476,7 +567,7 @@ private static function get_content_annotation_validation_errors( $annotations,
 					__( 'Message %d content annotation \'lastModified\' must be a string', 'mcp-adapter' ),
 					$message_index
 				);
-			} elseif ( ! self::validate_iso8601_timestamp( $annotations['lastModified'] ) ) {
+			} elseif ( ! McpValidator::validate_iso8601_timestamp( $annotations['lastModified'] ) ) {
```
JasonTheAdams: Yeah, interesting that we're not trimming here. What's the difference?

includes/Domain/Utils/McpAnnotationMapper.php
```diff
@@ -0,0 +1,98 @@
...
+namespace WP\MCP\Domain\Utils;
+
+/**
+ * Utility class for mapping WordPress ability annotations to MCP Annotations format.
+ *
+ * Provides shared annotation mapping and transformation logic used across multiple
+ * MCP component registration classes. Handles conversion of WordPress-format annotations
+ * to MCP-compliant annotation structures.
+ */
+class McpAnnotationMapper {
```
JasonTheAdams, line 19: Hmm, I think this class is doing too much. It's both mapping the WP version to the MCP version _and_ providing validation. I think we should limit this to simply knowing the map, and then move the specific annotation validations to `McpValidator`, which is then used by `McpPromptValidator`. We can use it here, if necessary, but the main point is the validation itself should be elsewhere.

includes/Domain/Utils/McpValidator.php
```diff
@@ -0,0 +1,188 @@
...
+	public static function validate_image_mime_type( string $mime_type ): bool {
+		$valid_image_types = array(
+			'image/jpeg',
+			'image/jpg',
+			'image/png',
+			'image/gif',
+			'image/webp',
+			'image/bmp',
+			'image/svg+xml',
+		);
```
JasonTheAdams: Does the MCP specification define a fixed set like this? If not, we should simply check for the `image/` prefix.

includes/Domain/Utils/McpValidator.php
```diff
@@ -0,0 +1,188 @@
...
+	public static function validate_audio_mime_type( string $mime_type ): bool {
+		$valid_audio_types = array(
+			'audio/wav',
+			'audio/mp3',
+			'audio/mpeg',
+			'audio/ogg',
+			'audio/webm',
+			'audio/aac',
+			'audio/flac',
+		);
```
JasonTheAdams: Same question as above regarding a fixed list in MCP.

includes/Domain/Tools/RegisterAbilityAsMcpTool.php
```diff
@@ -59,6 +59,76 @@ private function __construct( WP_Ability $ability, McpServer $mcp_server ) {
...
+			'destructive' => 'destructiveHint',
+			'idempotent'  => 'idempotentHint',
+		);
+
+		$valid_mcp_fields = array(
+			'readOnlyHint'    => 'boolean',
+			'destructiveHint' => 'boolean',
+			'idempotentHint'  => 'boolean',
+			'openWorldHint'   => 'boolean',
+			'title'           => 'string',
```
JasonTheAdams: Should the title annotation fall back on the Ability label, if not provided?

JasonTheAdams (changes requested):
Great work, @galatanovidiu! Left some comments and questions.

docs/guides/creating-abilities.md
galatanovidiu: `open-world` doesn’t currently exist in the Abilities API. If it’s added in the future, we can easily add the mapping at that point. At this stage, we also don’t know how exactly it might be defined (`open-world`, `openWorld`, etc.), so I’d prefer not to have the MCP adapter dictate how future annotations are introduced.

Currenty a `openWorldHint` anotations can be added to abilities anotations as the registry is not forcing you to use only the documented ones (readonly, destructive, idempotent).

On the mcp-adpater side I'm more restrictive and only allow the ones that are documented on the MCP specifications. I’m not strongly attached to keeping this restrictive though, and I’m happy to relax it if we decide so.

JasonTheAdams: I completely agree. We can add support for an `open-world` WP annotation in the future. I think what you have here is good. 😄

includes/Domain/Tools/RegisterAbilityAsMcpTool.php
galatanovidiu: Based on the MCP specifications, a title can be added both at the top level AND as part of annotations.
I was ignoring the annotations title.

The chnages made on [94fec28](https://github.com/WordPress/mcp-adapter/pull/91/commits/94fec282daaeb168816187cad1dafffea166e480) will ensure that `annotations.title` is added and they will default to the ability lablel

Ref: https://modelcontextprotocol.io/specification/2025-06-18/schema#tool-title

includes/Domain/Utils/McpAnnotationMapper.php
galatanovidiu: Yes, you are right. Doing to much and being to restrictive. 
I've simplified the [mapper](https://github.com/WordPress/mcp-adapter/pull/91/files#diff-53793054abace4261e57fc20d822c743ef20475ec034bf126b11f2309f165d9bR89).
I've moved the validation into the [validation functionality](https://github.com/WordPress/mcp-adapter/pull/91/files#diff-cde034d6f33a32867d15679c55a9f7221ac05f4865a20384b2c64dfeb0a0c301R316).

includes/Domain/Prompts/McpPromptValidator.php
galatanovidiu: I've made some extensive modifications related to data normalization. Please recheck the code

includes/Domain/Utils/McpValidator.php
galatanovidiu: You are right. I've modified it to relaxed validation

includes/Domain/Utils/McpValidator.php
galatanovidiu: Same as above. I've modified it to a more relaxed validation

includes/Domain/Prompts/McpPrompt.php
```diff
@@ -66,9 +66,9 @@ class McpPrompt {
 	/**
 	 * The MCP server instance this prompt belongs to.
 	 *
-	 * @var \WP\MCP\Core\McpServer
+	 * @var \WP\MCP\Core\McpServer|null
 	 */
-	private McpServer $mcp_server;
+	private ?McpServer $mcp_server = null;
```
JasonTheAdams, line 71: I'm confused why it's possible for the McpServer to be nullable. Is there a use case for this? Or should it be a required parameter in the constructor?

includes/Domain/Utils/McpValidator.php
```diff
@@ -0,0 +1,370 @@
...
+				if ( ! is_numeric( $value ) ) {
+					$errors[] = __( 'Annotation field priority must be a number', 'mcp-adapter' );
+					continue;
+				}
+				if ( self::validate_priority( $value ) ) {
+					continue;
+				}
+				$errors[] = __( 'Annotation field priority must be between 0.0 and 1.0', 'mcp-adapter' );
+				continue;
+			}
```
JasonTheAdams: Let's use a `switch...case` here to make our cases clear and that we're comparing against `$field`.

includes/Domain/Utils/McpValidator.php
```diff
@@ -0,0 +1,370 @@
...
+			if ( ! empty( trim( $value ) ) ) {
+				continue;
+			}
+
+			$errors[] = sprintf(
+				/* translators: %s: annotation field name */
+				__( 'Tool annotation field %s must be a non-empty string', 'mcp-adapter' ),
+				$field
+			);
```
JasonTheAdams: Same here for the use of `switch...case`. The logic is a bit confusing to follow since we're really working conditionally on `$field`, but using early returns.
```php
switch( $field ) {
    case 'readOnlyHint':
    case 'destructiveHint':
    	// etc., and validation is here
		break;

    case 'title':
       // title validation here
}
```

includes/Domain/Utils/McpAnnotationMapper.php
```diff
@@ -0,0 +1,180 @@
...
+	 * @return mixed|null Normalized value or null if invalid.
+	 */
+	private static function normalize_annotation_value( string $field_type, $value ) {
+		switch ( $field_type ) {
+			case 'boolean':
+				return (bool) $value;
+
+			case 'string':
+				if ( ! is_scalar( $value ) ) {
+					return null;
```
JasonTheAdams, line 156: I wonder if it would make more sense to return a `WP_Error` when invalid, instead?

includes/Domain/Utils/McpAnnotationMapper.php
```diff
@@ -0,0 +1,180 @@
...
+		),
+		'idempotentHint'  => array(
+			'type'             => 'boolean',
+			'features'         => array( 'tool' ),
+			'ability_property' => 'idempotent',
+		),
+		'openWorldHint'   => array(
+			'type'             => 'boolean',
+			'features'         => array( 'tool' ),
+			'ability_property' => '',
```
JasonTheAdams: Nit: I prefer `null` over `''`, as an empty string is not truly nothing.

includes/Domain/Utils/McpAnnotationMapper.php
```diff
@@ -0,0 +1,180 @@
...
+	 *
+	 * @param array  $annotations     Raw annotations from the ability.
+	 * @param string $mcp_field       The MCP field name.
+	 * @param string $ability_property Optional WordPress-format field name.
+	 *
+	 * @return array{has_value:bool,value:mixed}
+	 */
+	private static function resolve_annotation_value( array $annotations, string $mcp_field, string $ability_property ): array {
+		// WordPress-format overrides take precedence when present.
+		if ( '' !== $ability_property && array_key_exists( $ability_property, $annotations ) && ! is_null( $annotations[ $ability_property ] ) ) {
```
JasonTheAdams: Should we be checking `! is_null( $annotations[ $ability_property ] )`? If the overriding annotation is null, then shouldn't we still consider that intentional?

includes/Domain/Utils/McpAnnotationMapper.php
```diff
@@ -0,0 +1,180 @@
...
+			return array(
+				'has_value' => true,
+				'value'     => $annotations[ $mcp_field ],
+			);
+		}
+
+		return array(
+			'has_value' => false,
+			'value'     => null,
+		);
```
JasonTheAdams: I think we should simply return the value or else return a `WP_Error` here to clean things up.

includes/Domain/Utils/McpAnnotationMapper.php
```diff
@@ -0,0 +1,180 @@
...
+			);
+
+			if ( ! $value_info['has_value'] ) {
+				continue;
+			}
+
+			$normalized = self::normalize_annotation_value( $config['type'], $value_info['value'] );
+			if ( null === $normalized ) {
+				continue;
+			}
```
JasonTheAdams: If either getting the value or normalizing returns a `WP_Error`, let's use `_doing_it_wrong()` before continuing.

includes/Domain/Prompts/McpPrompt.php
JasonTheAdams: Same question for Tool and Resource.

JasonTheAdams (changes requested):
Good improvements, @galatanovidiu! I left some more feedback. 😄

includes/Domain/Utils/McpAnnotationMapper.php
galatanovidiu: Done on [0b99a5b](https://github.com/WordPress/mcp-adapter/pull/91/commits/0b99a5bf22f332433d40154d043aa53279a51dbf)

includes/Domain/Prompts/McpPrompt.php
galatanovidiu: This allows flexible component registration, components can be created independently and then associated with a server during registration.

I can modify this and make it more strict if you prefer.

includes/Domain/Utils/McpValidator.php
galatanovidiu: Done on [0bb6289](https://github.com/WordPress/mcp-adapter/pull/91/commits/0bb62897e35c6565f05f81e66f0096b33c2dfb25)

includes/Domain/Prompts/McpPrompt.php
JasonTheAdams: I see. This actually raises a different question, then: Why are the prompts, tools, and resources coupled to a specific server instance, rather than something ingested by the server?

includes/Domain/Utils/McpValidator.php
galatanovidiu: Done on [b0df6c6](https://github.com/WordPress/mcp-adapter/pull/91/commits/b0df6c69ad221e39997a2a2bcf43927315bec486)

This looks much cleaner BTW

includes/Domain/Utils/McpAnnotationMapper.php
galatanovidiu: We can do that, but I do not wnat to validate at this stage. Here I'm just normalizing the data included. 
We have the `mcp_adapter_validation_enabled` setting, which enables component validation if the user wants it.

includes/Domain/Utils/McpAnnotationMapper.php
galatanovidiu: I do not think so. By default thise values are null https://github.com/WordPress/WordPress/blob/master/wp-includes/abilities-api/class-wp-ability.php#L38

So null cannot be intentional as it is the default value, also null annotations have no meaning, they shoul have a value. In MCP we remove them if they are null

includes/Domain/Utils/McpAnnotationMapper.php
galatanovidiu: I've simplified the `resolve_annotation_value` method but it will return null instead of WP_Error. As I was saying, at this point I want to normalize the annotations not validating

[08f972d](https://github.com/WordPress/mcp-adapter/pull/91/commits/08f972dc91a9c7545efdf0f6e8bce29da9fd4006)

includes/Domain/Utils/McpAnnotationMapper.php
galatanovidiu: Same as above

includes/Domain/Prompts/McpPrompt.php
galatanovidiu: hhhh, yes, this creates a circular dependency:
- Components depend on the server (via `get_mcp_server()`)
- The server contains/registers components

Probably, we can check for uniqueness and validate it inside McpComponentRegistry (or something similar). I need to think about this.

includes/Domain/Utils/McpAnnotationMapper.php
JasonTheAdams: Hmm... then shouldn't we be including whatever the value is in the returned results? The fact that we're omitting things from the results means that we are validating — i.e. if it's not valid then we omit it.

includes/Domain/Utils/McpAnnotationMapper.php
JasonTheAdams: Hmm, so this would be covering the scenario wherein someone ignores the Ability annotation, leaving it null, and uses the MCP alternative instead? That seems like a pretty niche scenario. It also assumes that all built-in annotations in the future will have a default.

includes/Domain/Prompts/McpPrompt.php
JasonTheAdams: Yeah, I would think the dependency chain is Server > Registry > Tools/Resources/Prompts. Ideally, the dependency is uni-directional, so the components aren't aware of the registry or server, and the registry isn't aware of the server.

includes/Domain/Utils/McpAnnotationMapper.php
galatanovidiu: If someone ignores the ability annotation and relies on the MCP alternative instead, then the MCP value will be used. This approach also maintains backward compatibility.

I’m not sure what future ability annotations might look like, but we can address that when the time comes. That said, I do expect them to follow the same format, with the default value remaining null.

As a side note, we should be more careful with backward compatibility moving forward. The previous breaking change caused issues because WooCommerce includes an older version of the mcp-adapter. If someone includes a newer version of the adapter, it could break their site. See issue #87 for details.

includes/Domain/Utils/McpAnnotationMapper.php
galatanovidiu: I see what you mean, returning null does effectively exclude invalid values, which could be seen as a form of validation. My intent here, though, is to keep this step focused purely on normalization and avoid introducing validation logic at this stage.

I also don’t want to throw errors or risk breaking the entire MCP functionality if a tool annotation happens to be defined incorrectly. Returning null here ensures the system remains resilient and continues working even when a single annotation is malformed.

The explicit validation path is already available through the `mcp_adapter_validation_enabled` filter, which users can enable if they prefer stricter behavior.

Anyways, I'm open to discuss a  refactor related to normalization/validation functionality, but this should be done on another issue/PR as this one is getting really big :)

includes/Domain/Prompts/McpPrompt.php
galatanovidiu: Should we address this on another issue/PR?

JasonTheAdams: Yeah, let's resolve this in a subsequent PR. 👍

includes/Domain/Prompts/McpPromptValidator.php
galatanovidiu: This has been addresed on b0df6c6 and 0bb6289

includes/Domain/Utils/McpAnnotationMapper.php
JasonTheAdams: I honestly wish we didn't have those default annotations set as null. That muddies the difference between the user having not set the annotation at all versus it being a default.

That said, I think the chances of someone setting that as null, then setting an MCP alternative as something, but actually wanting the nulled WP version to be used it really unlikely. So let's move on. 👍

includes/Domain/Utils/McpAnnotationMapper.php
JasonTheAdams: Agreed, this isn't hurting anything for now, so we can move on.

Just to get my thought out, I'm actually open to things effectively working as they do now. My only thought was to use `_doing_it_wrong()` or something like that to communicate _something_ so it doesn't simply silently fail and move along without any clue as to what happened.

JasonTheAdams (approved):
Great work, @galatanovidiu! We have a couple discussions going on, but to your point I believe anything remaining can/should be resolved in a subsequent PR.

LGTM! 🚀

PR #93, opened 2025-11-17
Handle Flattened Input/Output Schemas from Abilities API

### LINKED ISSUE #88: Handle flattened input/output schemas from Abilities API
The Abilities API supports both object-based and flattened (simple type) schemas for inputs and outputs. However, MCP requires `input_schema.type` to always be `"object"`.

This means the MCP adapter needs to transform flattened schemas into object structures when converting Abilities to MCP tools.

### Context

Related PR: https://github.com/WordPress/abilities-api/pull/61#pullrequestreview-3197110787

### Current Behavior

The Abilities API allows schemas like:
```php
'input_schema' => [
    'type' => 'string',
    'description' => 'Post type to count'
]
```

### Required Behavior

The MCP adapter must transform these into:
```json
{
    "type": "object",
    "properties": {
        "input": {
            "type": "string",
            "description": "Post type to count"
        }
    }
}
```

### Implementation Notes

1. **Property naming:** Based on discussion with @gziolo, use `"input"` for input schemas and `"output"` for output schemas
2. **When to transform:** Apply transformation when `type !== "object"`
3. **Description:** Copy the description from the original schema to the wrapper property if available
4. **Arrays:** This should also handle `type: "array"` cases

### Additional Considerations

- Document this transformation behavior for future adapter implementations
- Consider adding tests for various simple types (string, number, boolean, array)
- Ensure the transformation preserves all schema properties (enum, format, etc.)

### Acceptance Criteria

- [ ] Adapter successfully transforms flattened input schemas to MCP-compatible object format
- [ ] Adapter successfully transforms flattened output schemas to MCP-compatible object format  
- [ ] Tests cover string, number, boolean, and array types
- [ ] Documentation updated to explain the transformation
- [ ] Existing object-based schemas continue to work without modification

galatanovidiu:
## WHY

The Abilities API supports both object-based and flattened schemas, but the MCP specification requires all tool input schemas to be of type `"object"`. Without this transformation, abilities using flattened schemas would fail MCP validation. This change ensures full MCP specification compliance while maintaining backward compatibility with existing object-based schemas.

## WHAT

This PR adds automatic transformation of flattened schemas (string, number, boolean, array) to MCP-compatible object schemas for both input and output schemas. The adapter now detects when an ability uses a flattened schema and wraps it in an object structure with a single property (`"input"` for inputs, `"result"` for outputs).

The implementation uses a `SchemaTransformer` utility class that automatically wraps flattened schemas during tool registration and unwraps them during execution, ensuring abilities receive and return values in their original format while MCP clients see compliant object schemas.

## Changes

- Added `SchemaTransformer` utility class for schema transformation
- Modified `RegisterAbilityAsMcpTool` to transform input and output schemas
- Added metadata support to `McpTool` to track transformation state
- Updated `ToolsHandler` to unwrap input arguments and wrap output results
- Added comprehensive test coverage for all schema types
- Updated documentation with examples of both schema formats

Closes #88

### DIFF
--- a/docs/guides/creating-abilities.md
+++ b/docs/guides/creating-abilities.md
@@ -40,7 +40,7 @@ If not specified, abilities default to `type: 'tool'`.
 wp_register_ability('my-plugin/my-ability', [
     'label' => 'My Ability',
     'description' => 'What this ability does',
-    'input_schema' => [...],      // For tools
+    'input_schema' => [...],      // For tools (supports both object and flattened schemas)
     'output_schema' => [...],     // Optional for tools
     'execute_callback' => 'my_callback',
     'permission_callback' => 'my_permission_check',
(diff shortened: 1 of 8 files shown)


includes/Handlers/Tools/ToolsHandler.php
```diff
@@ -383,6 +392,14 @@ public function handle_tool_call( array $params, int $request_id = 0 ): array {
...
 
+			// Wrap result if output schema was transformed from flattened to object format
+			$output_wrapped = ! empty( $tool_metadata['_output_schema_transformed'] );
+			if ( $output_wrapped ) {
+				// Wrap: "value" → {result: "value"}
+				$output_wrapper = $tool_metadata['_output_schema_wrapper'] ?? 'result';
+				$result         = array( $output_wrapper => $result );
+			}
+
 			// Successful execution - add metadata.
```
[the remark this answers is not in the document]
galatanovidiu: resolved on [8ab1dbf](https://github.com/WordPress/mcp-adapter/pull/93/commits/8ab1dbfb61a19c794073aceeab4fb0ab6d65f407)

includes/Domain/Tools/McpTool.php
JasonTheAdams, line 1: Is this metadata stuff related to the schema flattening in some way? Or is this just being tucked in? 😁

docs/guides/creating-abilities.md
```diff
@@ -56,6 +56,193 @@ wp_register_ability('my-plugin/my-ability', [
...
+- You want descriptive parameter names
+- Your output contains multiple related values (e.g., `{post_id, url, status}`)
+
+**Use Flattened Schemas when:**
+- Your ability accepts/returns a single, simple value
+- The input/output is straightforward (e.g., a string, number, boolean, or array)
+- You want to simplify the API for basic operations
+- Your output is a single primitive value (e.g., a count, a slug, a boolean flag)
+
+**Note**: All schema metadata (descriptions, constraints, enums, etc.) is preserved during the automatic transformation from flattened to object format. Input schemas use `"input"` as the wrapper property, while output schemas use `"result"`.
```
JasonTheAdams, line 244: Can you help me understand this wrapper stuff? Digging through the code, I'm not grasping what this is about.

JasonTheAdams (commented):
Nice work, @galatanovidiu! Working my way through a review and have a couple clarifying questions.

docs/guides/creating-abilities.md
galatanovidiu: The "wrapper" is the extra object structure MCP requires around single properties. In the regular Abilities API, if you have just one property, you can define it directly:
```php
'input_schema' => [
    'type' => 'number',
    'description' => 'Maximum number of posts',
]
```

However, MCP requires the input schema to always be an object type (not number, string, etc.). So we need to wrap that single property inside an object structure:
```
'input_schema' => [
    'type' => 'object',
    'properties' => [
        'input' => [
            'type' => 'number',
            'description' => 'Maximum number of posts'
        ],
    ],
]
```

includes/Domain/Tools/McpTool.php
galatanovidiu: Yes, it's directly related to schema flattening! The metadata stores information about the flattening transformation itself. Specifically, I'm using it to track:

- Whether a schema was flattened (so we know it needs to be unwrapped)
- The original property name that was flattened (e.g., input from the wrapper example)

This way, when MCP sends back the wrapped object like {'input': 5}, we can use the metadata to know we should extract just the **5** and pass it to the Ability as a simple number, matching what the Ability's original schema expected. Without this metadata, we wouldn't be able to properly reverse the flattening transformation.

docs/guides/creating-abilities.md
JasonTheAdams: Ohhh, I see. Right, if it's an object then each property has to be called something. Fair enough! 😄

JasonTheAdams (approved):
I really thought about this one, and I think I concluded that if Tool was a DTO we could make some nice, declarative methods for this, but since it's an array then I think this is good! Nice work, @galatanovidiu!

PR #94, opened 2025-11-17
Create `.gitattributes` to exclude development files from releases

BrianHenryIE:
Currently, `composer install` installs the package with all its development files, `/tests/`, `.prettierrc.js` etc. which are unnecessary when installing the package in another project.

This PR adds `.gitattributes` file in the project root with a list of files marked `export-ignore`.

* https://git-scm.com/docs/gitattributes
* https://blog.madewithlove.be/post/gitattributes/
* https://github.com/WordPress/Requests/blob/v2.0.15/.gitattributes
* https://github.com/WordPress/wp-ai-client/blob/0.1.0/.gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,23 @@
+#
+# Exclude these files from release archives.
+# This will also make them unavailable when using Composer with `--prefer-dist`.
+# If you develop for this repo using Composer, use `--prefer-source`.
+#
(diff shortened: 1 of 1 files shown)


JasonTheAdams (approved):
Thanks for this, @BrianHenryIE!

PR #111, opened 2025-12-12
Fix tools/call error masking and preserve empty parameters

optimikelabs:
## Context

* Plugin: **WordPress/mcp-adapter** (latest release, default server)
* WordPress **6.9+** with **Abilities API**
* MCP clients call tools via **tools/list** and **tools/call**

### 1) `tools/call` incorrectly treats any top-level `error` key in a tool result as an MCP failure

* Some abilities legitimately return arrays containing `error` as part of their own schema.
* `ToolsHandler::call_tool()` currently checks `isset($result['error'])` and converts the response to `isError: true`, masking the structured payload and producing a generic client error message.

**Expected**

* Only internal MCP/JSON-RPC errors should be treated as failures.
* Legitimate ability outputs containing an `error` field should be returned normally.

### 2) `ExecuteAbilityAbility` turns empty `parameters` objects `{}` into `null`

* `empty($input['parameters'])` makes `{}` become `null`.
* Abilities requiring an object then fail validation.

**Expected**

* Preserve `{}` when the key exists.
* Use `null` only when the key is absent.

## Fix

* **ToolsHandler**: treat top-level `error` as reserved only when `_metadata.failure_reason` is present.
* **ExecuteAbilityAbility**: replace `empty()` checks with `array_key_exists()` to preserve `{}`.

## Notes / impact

* Prevents masking legitimate tool outputs and avoids generic MCP client errors.
* Restores correct behavior for tools that accept empty objects.
* No behavior change for genuine internal errors or `WP_Error` paths.

---

### DIFF
--- a/includes/Abilities/ExecuteAbilityAbility.php
+++ b/includes/Abilities/ExecuteAbilityAbility.php
@@ -114,8 +114,10 @@ final class ExecuteAbilityAbility {
 			return new \WP_Error( 'ability_not_found', "Ability '{$ability_name}' not found" );
 		}
 
-		// Check if the user has permission to execute the target ability
-		$parameters        = empty( $input['parameters'] ) ? null : $input['parameters'];
+		// Check if the user has permission to execute the target ability.
+		// Preserve empty objects {} instead of converting them to null. Some abilities
+		// require an object (type: object) and will fail if passed null.
+		$parameters        = array_key_exists( 'parameters', $input ) ? $input['parameters'] : null;
 		$permission_result = $ability->check_permissions( $parameters );
 
 		// Return WP_Error as-is, or convert other values to boolean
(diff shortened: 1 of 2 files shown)


galatanovidiu:
Hey @optimikelabs, thanks for this. Your analysis of both problems was spot on.

Both issues have been independently resolved in the v0.5.0 refactor, so this PR can't merge cleanly against trunk anymore. Here's what changed:

**Error masking** — `ToolsHandler` no longer uses the `handle_tool_call()` / `_metadata` / `failure_reason` pattern. The current code checks `is_wp_error($result)` first, then has a strict backward-compat check that only triggers when both `success === false` AND `error` is a non-empty string. So a tool that legitimately returns an `error` field without `success: false` won't be misinterpreted.

**Empty parameters** — Both `check_permission()` and `execute()` now use `$input['parameters'] ?? null` (which preserves `{}` since `[]` isn't null), plus parameters go through `AbilityArgumentNormalizer::normalize()` which handles the `{}` → `null` conversion only when the ability has no input schema.

Going to close this one, but the bug reports were solid. If you run into other issues, feel free to open a new one.

PR #118, opened 2026-01-07
Fix missing type in execute-ability output schema

bjornfix:
## Summary
- add explicit `type` for the `data` field in the `mcp-adapter/execute-ability` output schema

## Problem
`rest_validate_value_from_schema()` warns when a schema field lacks a `type`. The execute-ability output schema defines `data` with only a description, which triggers "Undefined array key \"type\"" warnings in `wp-includes/rest-api.php` when abilities return data.

## Fix
Provide an explicit multi-type declaration for `data` so schema validation is satisfied (object/array/string/number/integer/boolean/null).

## Test
- Ran `tools/call` -> `mcp-adapter/execute-ability` (e.g. `content/list-posts`) via `wp mcp-adapter serve` on a WP 6.9 site; warnings no longer appear and output validates.

### DIFF
--- a/includes/Abilities/ExecuteAbilityAbility.php
+++ b/includes/Abilities/ExecuteAbilityAbility.php
@@ -56,6 +56,7 @@ final class ExecuteAbilityAbility {
 					'properties' => array(
 						'success' => array( 'type' => 'boolean' ),
 						'data'    => array(
+							'type'        => array( 'object', 'array', 'string', 'number', 'integer', 'boolean', 'null' ),
 							'description' => 'The result data from the ability execution',
 						),
 						'error'   => array(


bjornfix:
Tested on WP 6.9 using `wp mcp-adapter serve` with `tools/call` -> `mcp-adapter/execute-ability` (e.g. `content/list-posts`). The "Undefined array key 'type'" warnings from `wp-includes/rest-api.php` no longer appear. If someone can review, that would be great — we’d like to ship this fix.

bjornfix:
Closing as already fixed in trunk. The current trunk includes the  field  declaration in  output schema. Thanks!

PR #122, opened 2026-01-08
[4/6] Refactor domain models to use MCP schema DTOs

galatanovidiu:
## Summary

Updates the three core domain models (McpTool, McpResource, McpPrompt) and their validators to use php-mcp-schema DTOs instead of manual array structures.

### Tools
- `McpTool` — Now wraps `WP\McpSchema\Server\Tools\Tool` DTO
- `McpToolValidator` — Updated validation logic
- `RegisterAbilityAsMcpTool` — Uses schema DTOs for conversion

### Resources
- `McpResource` — Now wraps `WP\McpSchema\Server\Resources\Resource` DTO
- `McpResourceValidator` — Updated validation logic
- `RegisterAbilityAsMcpResource` — Uses schema DTOs for conversion

### Prompts
- `McpPrompt` — Now wraps `WP\McpSchema\Server\Prompts\Prompt` DTO
- `McpPromptBuilder` — Updated to produce DTOs
- `McpPromptValidator` — Updated validation logic
- `RegisterAbilityAsMcpPrompt` — Uses schema DTOs for conversion

### Tests
- Updated all domain model tests
- Updated validator tests
- Updated ability conversion tests

## Why

- Type-safe domain models with IDE support
- Guaranteed MCP protocol compliance
- Cleaner serialization (DTOs handle JSON conversion)
- PHPStan Level 8 compliance

## Stacked PR

⚠️ **This is PR 4 of 6** in the MCP Schema integration series.

| # | Branch | Description |
|---|--------|-------------|
| 1 | `01-composer-deps` | Composer dependency |
| 2 | `02-infrastructure` | Error handling + observability |
| 3 | `03-domain-utils` | Domain contracts + utilities |
| **4** | **`04-domain-components`** | **← You are here** |
| 5 | `05-core` | Core + abilities + plugin |
| 6 | `06-handlers-transport` | Handlers + Transport + CLI |

**Merge order**: 1 → 2 → 3 → 4 → 5 → 6

## Testing

- [ ] Tool, Resource, Prompt model tests pass
- [ ] Validator tests pass
- [ ] Ability conversion tests pass

### DIFF
--- a/includes/Domain/Prompts/McpPromptValidator.php
+++ b/includes/Domain/Prompts/McpPromptValidator.php
@@ -11,6 +11,7 @@ namespace WP\MCP\Domain\Prompts;
 
 use WP\MCP\Domain\Resources\McpResourceValidator;
 use WP\MCP\Domain\Utils\McpValidator;
+use WP\McpSchema\Server\Prompts\DTO\Prompt;
 
 /**
  * Validates MCP prompts against the Model Context Protocol specification.
@@ -18,7 +19,7 @@ use WP\MCP\Domain\Utils\McpValidator;
  * Provides minimal, resource-efficient validation to ensure prompts conform
  * to the MCP schema requirements without heavy processing overhead.
  *
- * @link https://modelcontextprotocol.io/specification/2025-06-18/server/prompts
+ * @link https://modelcontextprotocol.io/specification/2025-11-25/server/prompts
  */
 class McpPromptValidator {
 
@@ -40,7 +41,49 @@ class McpPromptValidator {
 				__( 'Prompt validation failed: %s', 'mcp-adapter' ),
 				implode( ', ', $validation_errors )
 			);
-			return new \WP_Error( 'prompt_validation_failed', esc_html( $error_message ) );
+			return new \WP_Error( 'mcp_prompt_validation_failed', esc_html( $error_message ) );
+		}
+
+		return true;
+	}
+
+	/**
+	 * Validate a Prompt DTO against the MCP schema.
+	 *
+	 * @param \WP\McpSchema\Server\Prompts\DTO\Prompt $prompt The prompt DTO to validate.
+	 *
+	 * @return bool|\WP_Error True if valid, WP_Error otherwise.
+	 */
+	public static function validate_prompt_dto( Prompt $prompt ) {
+		$errors = array();
+
+		// Validate name.
+		if ( ! McpValidator::validate_name( $prompt->getName() ) ) {
+			$errors[] = __( 'Prompt name must be 1-128 characters and contain only [A-Za-z0-9_.-]', 'mcp-adapter' );
+		}
+
+		// Validate icons if present.
+		$icons = $prompt->getIcons();
+		if ( ! empty( $icons ) ) {
+			$icons_array  = array_map( static fn( $icon ) => $icon->toArray(), $icons );
+			$icons_result = McpValidator::validate_icons_array( $icons_array );
+			$icons_errors = self::format_icon_validation_errors( $icons_result );
+			$errors       = array_merge( $errors, $icons_errors );
+		}
+
+		// Validate annotations if present (shared annotations).
+		// Currently Prompt DTO doesn't have annotations field in spec, but if it did, we'd validate here.
+		// BaseMetadata has title and name, handled separately.
+
+		if ( ! empty( $errors ) ) {
+			return new \WP_Error(
+				'mcp_prompt_validation_failed',
+				sprintf(
+				/* translators: %s: list of validation errors */
+					__( 'Prompt validation failed: %s', 'mcp-adapter' ),
+					implode( '; ', $errors )
+				)
+			);
 		}
 
 		return true;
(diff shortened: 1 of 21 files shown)


JasonTheAdams (commented):
@galatanovidiu I'm curious why we're decorating the McpSchema `Prompt`, `Resource`, and `Tool`, instead of extending it in our own classes?

includes/Domain/Prompts/McpPromptValidator.php
```diff
@@ -40,53 +41,66 @@ public static function validate_prompt_data( array $prompt_data, string $context
...
-				__( "Prompt name '%s' is not unique. It already exists in the MCP server.", 'mcp-adapter' ),
-				$this_prompt_name
-			);
-			return new \WP_Error( 'prompt_not_unique', esc_html( $error_message ) );
-		}
-
-		return true;
+	public static function validate_prompt_instance( McpPrompt $prompt ) {
+		return self::validate_prompt_dto( $prompt->get_component() );
 	}
```
JasonTheAdams, line 101: I wonder if it would just be simpler to merge this with `validate_prompt_dto` into a `validate` which accepts either the schema DTO or `McpPrompt`? Do we need to differentiate?

Same with the other component validators if they work the same. It seems like we want folks to predominantly work with our decorated version, but this validates the underlying schema DTO — which is fine.

includes/Domain/Prompts/RegisterAbilityAsMcpPrompt.php
```diff
@@ -39,89 +44,312 @@
...
+		$prompt = new self( $ability );
 
 		return $prompt->get_prompt();
 	}
 
 	/**
-	 * Constructor.
+	 * Get the MCP prompt instance.
+	 *
+	 * @return \WP\McpSchema\Server\Prompts\DTO\Prompt|\WP_Error Prompt DTO or WP_Error if validation fails.
```
JasonTheAdams: Out of curiosity, why are we returning the schema `Prompt` here instead of our `McpPrompt`?

includes/Domain/Prompts/RegisterAbilityAsMcpPrompt.php
```diff
@@ -39,89 +44,312 @@
...
+	 *
+	 * Per MCP 2025-11-25 specification, PromptArgument has:
+	 * - name (string, required): Argument identifier
+	 * - title (string, optional): Human-readable display name
+	 * - description (string, optional): Human-readable description
+	 * - required (boolean, optional): Whether the argument must be provided
+	 *
+	 * @param array<int,array<string,mixed>> $explicit_arguments User-defined arguments array.
+	 *
+	 * @return array<int,\WP\McpSchema\Server\Prompts\DTO\PromptArgument>|\WP_Error PromptArgument DTOs or WP_Error.
```
JasonTheAdams, line 297:
```suggestion
	 * @return list<\WP\McpSchema\Server\Prompts\DTO\PromptArgument>|\WP_Error PromptArgument DTOs or WP_Error.
```

I feel like we've had this discussion. I'm inclined to use `list<t>` for this sort of thing as it's more specific to what this is actually returning, but maybe we opted not to use that in this project?

JasonTheAdams (commented):
Left some questions and comments, but nothing that's clearly required. I think most of these questions also span across the corresponding component classes.

galatanovidiu:
> @galatanovidiu I'm curious why we're decorating the McpSchema `Prompt`, `Resource`, and `Tool`, instead of extending it in our own classes?

@JasonTheAdams I went with composition because the schema DTOs are generated code. The php-mcp-schema package auto-generates these classes from the MCP spec, so extending them would create fragile coupling; if the generator changes signatures, our child classes break.

There's also a clean boundary: the schema `Tool` is pure protocol data (name, inputSchema, annotations), while `McpTool` adds WordPress-specific behavior (execution, permissions, observability). This keeps serialization simple—`$mcpTool->get_component()->toArray()` returns only protocol-compliant fields.

`McpComponentInterface` defines the contract (get_component, execute, check_permission), which gives us uniform handling across tools, resources, and prompts.

Is there a specific benefit to inheritance you're thinking of? Happy to reconsider if so.

includes/Domain/Prompts/McpPromptValidator.php
galatanovidiu: The instance method is intentionally a thin wrapper.

I kept them separate mainly for PHPStan type clarity. With separate methods, callers get precise type hints, and PHPStan catches misuse at analysis time. A merged `validate(Prompt|McpPrompt $prompt)` would work also.

Either approach is fine with me. Happy to merge them if you prefer a single entry point.

includes/Domain/Prompts/RegisterAbilityAsMcpPrompt.php
galatanovidiu: From `get_prompt()`, I only need the `Prompt` DTO (protocol-level data).                                                                  
This gets used to build the `prompts/list` response.

`McpPrompt` is our runtime component - it wraps the `Prompt` DTO along with the WP_Ability, observability context, permission callbacks, and adapter metadata.

Same pattern for `McpTool` and `McpResource`. 

This one is related to your first question, composition over inheritance.

includes/Domain/Prompts/RegisterAbilityAsMcpPrompt.php
galatanovidiu: Yes, I remember us having this discussion before, but I do not remember what the resolution was. I will create a new issue to make this migration during 0.5.0 development.

galatanovidiu: Issue created: https://github.com/WordPress/mcp-adapter/issues/127

JasonTheAdams:
@ovidiu-galatan Thanks for explaining that! Yeah, I guess what was confusing me the most in the code was the delineation between `Tool` and `McpTool`. On the one hand I'm like, "yup, one is for the spec, and the other is for our system." On the other hand, it's confusing to have two of the same thing, and to do things like pass the `McpTool` a `Tool`. That's why I intuited that we'd either be extending `Tool` or using a Decorator pattern. That way anyone interacting with this doesn't have to understand both and can simply use our `McpTool` as the end-all-be-all of how to create and manage tools.

Make sense?

galatanovidiu:
@JasonTheAdams I agree the current naming is confusing. Here's what I'm proposing.

### The problem

`Tool` and `McpTool` sound like two versions of the same thing. They're not. One is a data container for protocol serialization; the other is the actual component with execution and permissions.

### Why composition

Two reasons:

1. **Generated code breaks inheritance.** The php-mcp-schema DTOs are auto-generated from TypeScript. Constructor signatures can change between versions. Wrapping the DTO internally keeps our public API stable.

2. **Serialization must stay clean.** `Tool::toArray()` returns only MCP-compliant fields. If we inherit and add `$ability` or `$adapter_meta`, we'd have to override `toArray()` to filter them. That's fragile.

### Proposed improvements

**Option A: Rename in php-mcp-schema**

```
Tool     → ToolDto
Resource → ResourceDto
Prompt   → PromptDto
```

No releases yet, so now is the time. All consumers of the schema package benefit.

**Option B: Import aliases in mcp-adapter**

```php
use WP\McpSchema\Server\Tools\DTO\Tool as ToolDto;
```

Works without changing the schema package. Fallback if we need to ship php-mcp-schema first.

**Both options include:**

- Rename `get_component()` → `get_protocol_dto()`
- Add class-level docs explaining the wrapper pattern

### Result

Developers work with `McpTool`. The `ToolDto` is obviously just a data container for protocol responses.

Thoughts?

swissspidy:
Currently on vacation so can't do a thorough review, but at first glance I'd say let's go with option B for now.
From there we can always pivot to A if needed.

The lack of a release shouldn't be a factor. After all it's generated code so there will be breaking changes when the spec changes anyway.

JasonTheAdams (approved):
@ovidiu-galatan Let's go with Option B as @swissspidy suggested. I'm inclined to keep the other package simple since we're intending it for broad adoption outside of this project.

I'm approving for now since the direction is clear and we can move this along.

PR #124, opened 2026-01-08
[6/6] Complete MCP schema integration: handlers, transport, and CLI

galatanovidiu:
## Summary

Final PR in the MCP schema integration series. Updates all request handlers, HTTP/STDIO transports, and CLI bridge to use php-mcp-schema DTOs for request/response handling.

### Handlers
- `InitializeHandler` — Returns `InitializeResult` DTO
- `ToolsHandler` — Uses `Tool` DTOs for list/call operations
- `ResourcesHandler` — Uses `Resource` DTOs for list/read operations
- `PromptsHandler` — Uses `Prompt` DTOs for list/get operations
- `SystemHandler` — Updated system operations
- `HandlerHelperTrait` — Updated helper methods

### Transport
- `HttpTransport` — Updated HTTP request/response handling
- `McpRestTransportInterface` — Updated contract
- `HttpRequestHandler` — Uses DTOs for request parsing
- `HttpRequestContext` — Updated context handling
- `HttpSessionValidator` — Updated session management
- `JsonRpcResponseBuilder` — Produces DTO-based responses
- `McpTransportContext` — Updated transport context
- `RequestRouter` — Updated routing logic
- `SessionManager` — Updated session handling

### CLI
- `StdioServerBridge` — Updated STDIO transport bridge

### Tests
- All handler unit tests updated
- All transport unit tests updated
- Integration tests for full request/response flow
- DTO serialization regression tests

## Why

This completes the MCP schema integration:
- All JSON-RPC requests are parsed into DTOs
- All responses are built from DTOs
- Full type safety from request to response
- MCP 2025-11-25 specification compliance

## Stacked PR

⚠️ **This is PR 6 of 6** — Final PR in the MCP Schema integration series.

| # | Branch | Description |
|---|--------|-------------|
| 1 | `01-composer-deps` | Composer dependency |
| 2 | `02-infrastructure` | Error handling + observability |
| 3 | `03-domain-utils` | Domain contracts + utilities |
| 4 | `04-domain-components` | Tools, Resources, Prompts |
| 5 | `05-core` | Core + abilities + plugin |
| **6** | **`06-handlers-transport`** | **← You are here (FINAL)** |

**Merge order**: 1 → 2 → 3 → 4 → 5 → 6

## Testing

- [ ] All unit tests pass
- [ ] All integration tests pass
- [ ] PHPStan Level 8 passes
- [ ] PHPCS passes
- [ ] MCP Inspector validation passes


includes/Transport/Infrastructure/SessionManager.php
```diff
@@ -120,6 +89,104 @@ static function ( $a, $b ) {
...
+	private static function get_config(): array {
+		/**
+		 * Filters the maximum number of MCP sessions allowed per user.
+		 *
+		 * When a user exceeds this limit, the oldest inactive session is
+		 * automatically removed to make room for new sessions.
+		 *
+		 * @since 0.3.0
+		 *
+		 * @param int $max_sessions Maximum sessions per user. Default 5.
```
jorgefilipecosta: The default DEFAULT_MAX_SESSIONS seems to be 32, should we update the comment?

includes/Handlers/HandlerHelperTrait.php
```diff
@@ -35,13 +36,13 @@ protected function extract_params( array $data ): array {
...
-	 * @param string $message    Error message.
-	 * @param int    $request_id Optional. Request ID for JSON-RPC. Default 0.
+	 * @param int $code Error code.
+	 * @param string $message Error message.
+	 * @param string|int|null $request_id Optional. Request ID for JSON-RPC. Default 0.
 	 *
 	 * @return array Error response array with 'error' key.
 	 */
-	protected function create_error_response( int $code, string $message, int $request_id = 0 ): array {
+	protected function create_error_response( int $code, string $message, $request_id = 0 ): array {
```
jorgefilipecosta: Is there any caise where this function ends up being used? It seems McpErrorFactory:: is used everywhere, if thats the case could we remove error handling from the Trait?

includes/Handlers/Tools/ToolsHandler.php
```diff
@@ -37,411 +40,241 @@ public function __construct( McpServer $mcp ) {
...
+	 * This distinction is critical for LLM self-correction - execution errors are
+	 * visible to the LLM, while protocol errors indicate infrastructure issues.
+	 *
+	 * @param array $message Request message.
+	 * @param string|int|null $request_id Optional. The request ID for JSON-RPC. Default 0.
+	 *
+	 * @return \WP\McpSchema\Server\Tools\DTO\CallToolResult|\WP\McpSchema\Common\JsonRpc\DTO\JSONRPCErrorResponse
 	 */
-	public function call_tool( array $message, int $request_id = 0 ): array {
+	public function call_tool( array $message, $request_id = 0 ) {
```
jorgefilipecosta: On PromptsHandler::get_prompt and ResourcesHandler::read_resource we name the input as $params for consistency should we also use $params here instead of $message?

includes/Handlers/Tools/ToolsHandler.php
```diff
@@ -37,411 +40,241 @@ public function __construct( McpServer $mcp ) {
...
-					'error_type'     => get_class( $e ),
-				),
-			);
-		}
 
-		// Execute the tool callback.
-		try {
-			$result = $ability->execute( $args );
+			// Handle image results.
+			if ( isset( $result['type'] ) && 'image' === $result['type'] ) {
```
jorgefilipecosta: Would it make sense to include on this if "&& isset( $result['results'] )"? Just as a check the results are set.

jorgefilipecosta: svg images can contain script and be an attack vector, would it make sense to explictly disable svg type?

includes/Transport/Infrastructure/RequestRouter.php
```diff
@@ -202,44 +350,4 @@ private function categorize_error( \Throwable $exception ): string {
 
 		return $error_categories[ get_class( $exception ) ] ?? 'unknown';
```
jorgefilipecosta: We are matching by exact class name, this will not match common exceptions like OutOfBoundsException, UnexpectedValueException because we don't compare against the parent.
We may do something like:
foreach ( $error_categories as $class => $category ) {
    if ( $exception instanceof $class ) {
        return $category;
    }
}
return 'unknown';

To take into account inheritance.

includes/Handlers/Prompts/PromptsHandler.php
```diff
@@ -39,208 +66,489 @@ public function __construct( McpServer $mcp ) {
 	/**
```
jorgefilipecosta, line 65: Nitpick: we have a double blank line above.

includes/Handlers/Prompts/PromptsHandler.php
```diff
@@ -39,208 +66,489 @@ public function __construct( McpServer $mcp ) {
...
-	 * @return array Response with prompts list and metadata.
+	 * @return \WP\McpSchema\Server\Prompts\DTO\ListPromptsResult Response with prompts list DTO.
 	 */
-	public function list_prompts( int $request_id = 0 ): array {
-		// Get the registered prompts from the MCP instance and extract only the args.
-		$prompts = array();
-		foreach ( $this->mcp->get_prompts() as $prompt ) {
-			$prompts[] = $prompt->to_array();
-		}
+	public function list_prompts( $request_id = 0 ): ListPromptsResult {
```
jorgefilipecosta: $request_id is never used, could we remove it?

includes/Transport/Infrastructure/McpTransportContext.php
```diff
@@ -33,15 +33,15 @@ class McpTransportContext {
...
+	 * @param \WP\MCP\Handlers\Tools\ToolsHandler $tools_handler The tools handler.
+	 * @param \WP\MCP\Handlers\Resources\ResourcesHandler $resources_handler The resources handler.
+	 * @param \WP\MCP\Handlers\Prompts\PromptsHandler $prompts_handler The prompts handler.
+	 * @param \WP\MCP\Handlers\System\SystemHandler $system_handler The system handler.
+	 * @param string $observability_handler The observability handler class name.
 	 * @param \WP\MCP\Transport\Infrastructure\RequestRouter|null $request_router The request router service.
-	 * @param callable|null         $transport_permission_callback Optional custom permission callback for transport-level authentication.
+	 * @param callable|null $transport_permission_callback Optional custom permission callback for transport-level authentication.
 	 */
 	/**
```
jorgefilipecosta, line 47: We should */ ... /** with just * to fix the docblock.

tests/Unit/Transport/Infrastructure/DtoSerializationRegressionTest.php
```diff
@@ -0,0 +1,127 @@
...
+			$dto = ReadResourceResult::fromArray(
+				array(
+					'contents' => array( $text, $blob ),
+					'_meta'    => array(
+						'keep' => 'top-meta',
+					),
+				)
+			);
+
+			$result = $dto->toArray();
```
jorgefilipecosta: Nitpick: The identation has an extra tab.

includes/Transport/Infrastructure/HttpSessionValidator.php
```diff
@@ -106,13 +106,13 @@ public static function terminate_session( HttpRequestContext $context ) {
...
 		}
 
 		// Validate user authentication
 		$user_id = get_current_user_id();
 		if ( ! $user_id ) {
-			return McpErrorFactory::unauthorized( 0, 'User not authenticated' );
+			return McpErrorFactory::unauthorized( null, 'User not authenticated' )->toArray();
 		}
 
 		// Terminate the session
```
jorgefilipecosta, line 118: Even if delete_session returns false, we are returning true, should we return false if that case?

includes/Transport/Infrastructure/HttpRequestHandler.php
```diff
@@ -274,4 +244,35 @@ static function ( $response ) use ( $session_id ) {
 
 		$current_session_id = $session_id;
```
jorgefilipecosta, line 245: When the session id changes should we remove the previous filter?

tests/Unit/Transport/Infrastructure/RequestRouterTest.php
```diff
@@ -157,13 +160,33 @@ public function test_route_request_tools_call(): void {
...
+				'name'      => 'test-meta-leak',
+				'arguments' => array(),
+			),
+			1
+		);
+
+		$this->assertIsArray( $result );
+		$this->assertArrayHasKey( 'content', $result );
+		$this->assertIsArray( $result['content'] );
+			$this->assertNotEmpty( $result['content'] );
```
jorgefilipecosta: Nitpick: Extra tab.

includes/Handlers/Prompts/PromptsHandler.php
```diff
@@ -39,208 +66,489 @@ public function __construct( McpServer $mcp ) {
...
 
 			return array(
-				'error'     => McpErrorFactory::internal_error( $request_id, 'Prompt execution failed' )['error'],
-				'_metadata' => array(
-					'component_type' => 'prompt',
-					'prompt_name'    => $prompt_name,
-					'failure_reason' => 'execution_failed',
-					'error_type'     => get_class( $e ),
+				'type' => 'text',
+				'text' => isset( $content['text'] ) ? (string) $content['text'] : wp_json_encode( $content, JSON_PRETTY_PRINT ),
```
jorgefilipecosta: wp_json_encode may return false, should we check the return value?

includes/Handlers/Tools/ToolsHandler.php
```diff
@@ -37,411 +40,241 @@ public function __construct( McpServer $mcp ) {
...
-				// Protocol errors (keep JSON-RPC error format):
-				// - not_found (tool doesn't exist)
-				// - ability_retrieval_failed (internal error getting ability)
-				$protocol_errors = array( 'not_found', 'ability_retrieval_failed' );
-
-				if ( in_array( $failure_reason, $protocol_errors, true ) ) {
-					// Return as JSON-RPC error
-					return $result;
-				}
+			$tool_name = (string) $request_params['name'];
```
jorgefilipecosta: On list_prompts we do trim on the name, on read_resource too. Should we also do a trim here, or should we remove the trim from the other 2 functions?

includes/Handlers/Tools/ToolsHandler.php
```diff
@@ -37,411 +40,241 @@ public function __construct( McpServer $mcp ) {
...
-			);
-		}
 
-		// Execute the tool callback.
-		try {
-			$result = $ability->execute( $args );
+			// Handle image results.
+			if ( isset( $result['type'] ) && 'image' === $result['type'] ) {
+				$image_data = base64_encode( $result['results'] ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode
+				$mime_type  = $result['mimeType'] ?? 'image/png';
```
jorgefilipecosta: Nitpick: Would it make sense to have a DEFAULT_IMAGE_MIME_TYPE constant?

jorgefilipecosta (commented):
The changes look good, I just left some comments for consideration nothing major only small things I noticed while checking the code.

includes/Transport/Infrastructure/RequestRouter.php
galatanovidiu: Nice catch, you're right. Fixed in 1ac0909. Also reordered the array so that more specific exceptions (e.g., ArgumentCountError) are matched before their parents.

includes/Transport/Infrastructure/HttpSessionValidator.php
galatanovidiu: I think returning true here is actually the right behavior.

The only case where `delete_session()` returns false is if the session doesn't exist in storage anymore (the $user_id/$session_id empty checks are already guarded above). So we're really talking about a race condition where the session got deleted between the header validation and the delete call, which is an extremely narrow window.

If we return an error in that case, the client receives a 500 response for a session that's already ended. That's worse than returning success, since the client's intent (terminate the session) was already achieved. Termination endpoints work better when they're idempotent.

Also worth noting: MCP clients don't really inspect termination responses beyond the HTTP status code. They send DELETE and move on. And even if somehow a session record gets orphaned, SessionManager cleans up stale sessions on the next session creation via max-session enforcement.

So I'd lean toward keeping the current behavior. What do you think?

includes/Transport/Infrastructure/HttpRequestHandler.php
galatanovidiu: We can remove the filter, but the session ID can't actually change during a single request. The MCP spec says initialize MUST NOT be part of a JSON-RPC batch, and that's the only method that sets a _session_id. So there's no spec-compliant flow in which this method is called with two different session IDs.

includes/Handlers/Prompts/PromptsHandler.php
galatanovidiu: Nice catch, the invalid-type branch just below already handled this. Fixed in 8027f77, same pattern (fallback to '{}' on false).

includes/Handlers/Tools/ToolsHandler.php
galatanovidiu: "&& isset( $result['results'] )" addresed in b4ac351. The resource handling block above was already checking isset() on nested keys, so this was just an inconsistency.

---

Related to SVG's: 
SVGs are definitely an attack vector in rendering contexts. But I think this is a client-side responsibility, not something we should filter at the transport layer.

The adapter is just serializing tool results into MCP protocol DTOs. The data goes out as base64 inside a JSON-RPC response, not as rendered HTML. The MCP schema's ImageContent doesn't restrict mime types either ("Different providers may support different image types").

SVG sanitization should happen at the MCP client level (Claude, Cursor, etc.) when it decides how to render the image.

includes/Handlers/Tools/ToolsHandler.php
galatanovidiu: fixed in 988b15c. Now consistent with get_prompt and read_resource

includes/Handlers/Tools/ToolsHandler.php
galatanovidiu: Added trim() to match get_prompt and read_resource. Better to be defensive :).
Fixed in 85439c9

includes/Transport/Infrastructure/SessionManager.php
galatanovidiu: Nice catch, done on 0833c46.

includes/Handlers/HandlerHelperTrait.php
galatanovidiu: Ohhh, this was a good one, lots of dead code after refactoring 
Fixed in f2967d0

includes/Handlers/Prompts/PromptsHandler.php
galatanovidiu: Resolved on b3a21e1

includes/Transport/Infrastructure/McpTransportContext.php
galatanovidiu: Resolved on b3a21e1

tests/Unit/Transport/Infrastructure/DtoSerializationRegressionTest.php
galatanovidiu: Resolved on b3a21e1

tests/Unit/Transport/Infrastructure/RequestRouterTest.php
galatanovidiu: Resolved on b3a21e1

includes/Handlers/Tools/ToolsHandler.php
galatanovidiu: Fair point, added `DEFAULT_IMAGE_MIME_TYPE` constant. Fixed in d85e1c4.

includes/Handlers/Prompts/PromptsHandler.php
galatanovidiu: Yes, this is something that bugs me for some time, but I never got to resolve it. And it is on multiple locations.
Now that you are mentioning it ...
Fixed on 133736a

PR #123, opened 2026-01-08
[5/6] Update core layer and built-in abilities for MCP schema

galatanovidiu:
## Summary

Updates the Core layer (McpAdapter, McpServer, McpComponentRegistry), built-in abilities, and plugin bootstrap to work with the new MCP schema DTOs.

### Core
- `McpAdapter` — Updated server creation and management
- `McpServer` — Updated component access methods
- `McpComponentRegistry` — Updated to store DTO-based components
- `McpTransportFactory` — Updated transport instantiation

### Built-in Abilities
- `DiscoverAbilitiesAbility` — Updated response format
- `ExecuteAbilityAbility` — Updated execution flow
- `GetAbilityInfoAbility` — Updated info retrieval
- `McpAbilityHelperTrait` — Updated helper methods

### Plugin
- `Plugin` — Updated bootstrap sequence
- `Autoloader` — Updated class loading
- `DefaultServerFactory` — Updated default server creation

### CLI
- `McpCommand` — Updated CLI commands

### Tests
- Core component tests updated
- Ability tests updated
- Server factory tests updated

## Why

- Core layer now produces and consumes MCP schema DTOs
- Built-in abilities return properly formatted MCP responses
- Plugin initialization uses the new DTO-based flow

## Stacked PR

⚠️ **This is PR 5 of 6** in the MCP Schema integration series.

| # | Branch | Description |
|---|--------|-------------|
| 1 | `01-composer-deps` | Composer dependency |
| 2 | `02-infrastructure` | Error handling + observability |
| 3 | `03-domain-utils` | Domain contracts + utilities |
| 4 | `04-domain-components` | Tools, Resources, Prompts |
| **5** | **`05-core`** | **← You are here** |
| 6 | `06-handlers-transport` | Handlers + Transport + CLI |

**Merge order**: 1 → 2 → 3 → 4 → 5 → 6

## Testing

- [ ] Core component tests pass
- [ ] Ability tests pass
- [ ] Server factory tests pass

PR #121, opened 2026-01-08
[3/6] Add domain contracts and utility classes for MCP components

galatanovidiu:
## Summary

Introduces domain layer foundation: contracts for MCP components and utility classes for schema transformation, name sanitization, and validation.

### Contracts
- `McpComponentInterface` — Base interface for all MCP components
- `McpPromptBuilderInterface` — Updated prompt builder contract

### Utilities
- `AbilityArgumentNormalizer` — Normalizes ability arguments to MCP format
- `ContentBlockHelper` — Helper for creating MCP content blocks
- `McpAnnotationMapper` — Maps ability metadata to MCP annotations
- `McpNameSanitizer` — Ensures MCP-compliant naming
- `McpValidator` — Updated validation utilities
- `SchemaTransformer` — Transforms ability schemas to MCP JSON Schema

### Tests
- Full test coverage for all utility classes
- Updated `DummyAbility` fixture for testing

## Why

These utilities provide the building blocks for converting WordPress Abilities to MCP components:
- Consistent name formatting across tools, resources, and prompts
- Schema transformation from ability format to MCP JSON Schema
- Content block generation for responses

## Stacked PR

⚠️ **This is PR 3 of 6** in the MCP Schema integration series.

| # | Branch | Description |
|---|--------|-------------|
| 1 | `01-composer-deps` | Composer dependency |
| 2 | `02-infrastructure` | Error handling + observability |
| **3** | **`03-domain-utils`** | **← You are here** |
| 4 | `04-domain-components` | Tools, Resources, Prompts |
| 5 | `05-core` | Core + abilities + plugin |
| 6 | `06-handlers-transport` | Handlers + Transport + CLI |

**Merge order**: 1 → 2 → 3 → 4 → 5 → 6

## Testing

- [ ] All utility class tests pass
- [ ] Schema transformation produces valid MCP schemas

PR #129, opened 2026-01-28
Fix: Preserve empty arrays instead of converting to null

bjornfix:
## Summary

Fixes a bug where abilities with empty parameter objects fail validation with "input is not of type object" error.

## Root Cause

In `ExecuteAbilityAbility.php`, both `check_permission()` and `execute()` methods used:
```php
$parameters = empty( $input['parameters'] ) ? null : $input['parameters'];
```

In PHP, `empty([])` returns `true` for empty arrays. When an MCP client calls an ability with empty parameters `{}`, this code converted the empty object to `null`, causing target abilities to fail validation because `rest_validate_value_from_schema()` expects an object type, not `null`.

### Example Failing Ability

Abilities with this input schema (empty object with no required properties):
```php
'input_schema' => array(
    'type' => 'object',
    'additionalProperties' => false,
),
```

Expect to receive `[]` (empty array), NOT `null`. When they receive `null`, WordPress's `rest_validate_value_from_schema()` fails with "input is not of type object".

## Fix

Changed both occurrences to use the null coalescing operator:
```php
$parameters = $input['parameters'] ?? null;
```

This preserves empty arrays as `[]` while still handling missing parameters correctly (`null` is only used when the key is truly absent).

## Test Verification

```php
// Before fix
empty([])          // true  → converts to null  → FAILS validation

// After fix
[] ?? null         // []    → passes validation
null ?? null       // null  → still works for missing params
```

## Impact

- Abilities that accept empty parameter objects now work correctly
- Existing behavior for abilities with required or optional parameters is unchanged
- The change is minimal and surgical - only 2 lines modified

## Testing

- All toolset abilities with no required parameters now work correctly
- Abilities with required parameters continue to work as expected
- The fix was verified against real-world abilities in the mcp-abilities-toolset plugin

### DIFF
--- a/includes/Abilities/ExecuteAbilityAbility.php
+++ b/includes/Abilities/ExecuteAbilityAbility.php
@@ -115,7 +115,7 @@ final class ExecuteAbilityAbility {
 		}
 
 		// Check if the user has permission to execute the target ability
-		$parameters        = empty( $input['parameters'] ) ? null : $input['parameters'];
+		$parameters        = $input['parameters'] ?? null;
 		$permission_result = $ability->check_permissions( $parameters );
 
 		// Return WP_Error as-is, or convert other values to boolean
@@ -161,7 +161,7 @@ final class ExecuteAbilityAbility {
 	 */
 	public static function execute( $input = array() ): array {
 		$ability_name = $input['ability_name'] ?? '';
-		$parameters   = empty( $input['parameters'] ) ? null : $input['parameters'];
+		$parameters   = $input['parameters'] ?? null;
 
 		if ( empty( $ability_name ) ) {
 			return array(


justlevine:
Related: 
- https://core.trac.wordpress.org/ticket/64483
- https://github.com/WordPress/php-ai-client/pull/175

galatanovidiu:
Hey! Thanks for tackling this 🙏

There are 3 failing tests. The fix works for abilities with input_schema, but breaks abilities without one (the Abilities API rejects non-null input when no schema is defined).

This is already being addressed in the v0.5.0 refactor work:

PR #121 added an AbilityArgumentNormalizer class:
https://github.com/WordPress/mcp-adapter/blob/release/v0.5.0/includes/Domain/Utils/AbilityArgumentNormalizer.php

PR #123 wires it up in ExecuteAbilityAbility:

The approach: check if the target ability has an input_schema. If not, and params is [], convert to null. This handles both cases.

Since this is already in flight for v0.5.0,
There are 2 things we can do:

1) Replicate the same fix and release a patch 
2) Close this PR and let the refactor work land instead (hopefully we will land this soon)

bjornfix:
Right. Then we wait for 0.50. No need to address this separately.

PR #132, opened 2026-02-02
Compatibility with ChatGPT Apps and MCP Apps

artpi:
This adds ability to pass raw meta field so that it's exposed in the ChatGPT apps and MCP apps.

### DIFF
--- a/includes/Domain/Tools/McpTool.php
+++ b/includes/Domain/Tools/McpTool.php
@@ -72,11 +72,12 @@ class McpTool {
 	private array $annotations;
 
 	/**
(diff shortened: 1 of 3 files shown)


galatanovidiu:
Thank you for this, @artpi!

I'm going to close this one since the _meta support is already covered in #136 (the v0.5.0 release). The refactored architecture handles it through the ToolDto layer natively, so ability authors can pass _meta via ability.meta.mcp._meta, and it flows through to MCP clients automatically.

#136 should land on trunk soon, so this will be available with the release.

PR #140, opened 2026-03-02
Import WP_Error class instead of fully qualified references

### LINKED ISSUE #128: Import `WP_Error` class instead of using fully qualified references
### Summary

Update PHPCS configuration to allow importing `WP_Error` via `use` statements instead of requiring fully qualified references (`\WP_Error`) throughout the codebase.

### Background

During PR #123 review, @JasonTheAdams suggested importing `WP_Error` to drop the root namespace prefix. Investigation revealed that the current `SlevomatCodingStandard.Namespaces.FullyQualifiedExceptions` rule enforces fully qualified references for classes ending with `Error`, which includes `WP_Error`.

However, `WP_Error` is **not** an actual exception class (it doesn't extend `Exception` or `Throwable`) - it's a WordPress utility class for error handling. The rule's documentation explicitly supports adding such classes to `ignoredNames`.

### Proposed Changes

1. **Update `phpcs.xml`** to add `WP_Error` to `ignoredNames`:

```xml
<rule ref="SlevomatCodingStandard.Namespaces.FullyQualifiedExceptions">
    <properties>
        <property name="ignoredNames" type="array">
            <element value="WP_Error" />
        </property>
    </properties>
</rule>
```

2. **Update all namespaced files** that use `\WP_Error`:
   - Add `use WP_Error;` import statement
   - Replace `\WP_Error` with `WP_Error` in code
   - Update PHPDoc `@return` types from `\WP_Error` to `WP_Error`

### References

- PR #123 review comment: https://github.com/WordPress/mcp-adapter/pull/123
- Slevomat Coding Standard docs: https://github.com/slevomat/coding-standard/blob/master/doc/namespaces.md

galatanovidiu:
## What

- Add `WP_Error` to `ignoredNames` in the `SlevomatCodingStandard.Namespaces.FullyQualifiedExceptions` PHPCS rule
- Import `WP_Error` via `use` statements and replace `\WP_Error` with `WP_Error` in code across 28 files (17 source + 11 test)
- Keep `\WP_Error` fully qualified in PHPDoc annotations (required by `FullyQualifiedClassNameInAnnotation` rule)

## Why

`WP_Error` is not an exception class — it doesn't extend `Exception` or `Throwable`. The `FullyQualifiedExceptions` rule was enforcing FQN for it solely because the class name ends with "Error". The rule's `ignoredNames` property exists for exactly this case, allowing imported short names in code while annotations remain fully qualified.

Closes #128

### DIFF
--- a/includes/Abilities/DiscoverAbilitiesAbility.php
+++ b/includes/Abilities/DiscoverAbilitiesAbility.php
@@ -9,6 +9,8 @@ declare( strict_types=1 );
 
 namespace WP\MCP\Abilities;
 
+use WP_Error;
+
 /**
  * Discover Abilities - Lists all available WordPress abilities in the system.
  *
@@ -120,7 +122,7 @@ final class DiscoverAbilitiesAbility {
 	public static function check_permission( $input = array() ) {
 		// Verify caller identity - ensure user is authenticated
 		if ( ! is_user_logged_in() ) {
-			return new \WP_Error( 'authentication_required', 'User must be authenticated to access this ability' );
+			return new WP_Error( 'authentication_required', 'User must be authenticated to access this ability' );
 		}
 
 		/**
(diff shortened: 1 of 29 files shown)


gziolo (approved):
Verified locally — tests pass, PHPCS clean, PHPStan Level 8 zero errors.

Confirmed PHPDoc annotations retain `\WP_Error` (FQ) while code uses the imported short name — correct per the two separate PHPCS rules (`FullyQualifiedClassNameInAnnotation` vs `FullyQualifiedExceptions`).

PR #142, opened 2026-03-02
feat: allow custom annotation fields to pass through McpAnnotationMapper

### LINKED ISSUE #134: Allow custom annotation fields to pass through McpAnnotationMapper
### Problem

`McpAnnotationMapper::map()` uses a strict whitelist of known MCP annotation fields (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`, `audience`, `priority`, `title`, `lastModified`). Any annotation key not in this list is silently discarded during the mapping from WordPress Abilities API to MCP tools.

This prevents consumers from attaching domain-specific metadata to tool annotations.

Compliant MCP clients will ignore fields they don't recognize, so passing through additional fields is protocol-safe.

### Proposed Change

After mapping known MCP annotation fields, preserve any remaining keys from the source annotations array. Unknown fields should be included as-is in the mapped output.

galatanovidiu:
Fixes #134

## What

`McpAnnotationMapper::map()` now preserves unknown annotation keys from the
source array, passing them through as-is alongside the mapped MCP fields.

**Changes:**

- After the known-field mapping loop, a second pass collects any keys that
  are not recognized MCP fields or WordPress ability property aliases
- Unknown fields are included without type normalization (they are
  domain-specific metadata the mapper has no schema for)
- A new `get_known_annotation_keys()` private helper builds and caches the
  full set of known keys for fast `isset()` lookups
- Pass-through applies to `tool` and `resource` feature types only; `prompt`
  annotations remain unaffected
- Null unknown values are filtered out, consistent with existing behavior
  for known fields

## Why

The strict whitelist silently discarded any annotation key not in the MCP
specification. This prevented consumers from attaching domain-specific
metadata to tool and resource annotations — metadata that compliant MCP
clients will simply ignore per the protocol's extensibility model.

Passing through unknown fields is protocol-safe and enables use cases like
vendor-prefixed tags (`x-vendor-category`), deprecation markers, or
grouping metadata without requiring upstream schema changes.

## Testing

- 7 new unit tests in `McpAnnotationMapperTest`: pass-through for tools,
  resources, prompt exclusion, feature-type exclusion, null filtering,
  mixed known+unknown fields, and WordPress alias exclusion
- 2 existing integration tests updated to reflect new pass-through behavior
  (`RegisterAbilityAsMcpResourceTest`, `RegisterAbilityAsMcpToolTest`)
- Full suite: 596 tests, 2353 assertions — all pass
- PHPCS clean, PHPStan Level 8 zero errors

Linear: AIINT-273

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ Part of the [**AI Building Blocks for WordPress** initiative](https://make.wordp
 
 The official WordPress package for MCP integration that exposes WordPress abilities as [Model Context Protocol (MCP)](https://modelcontextprotocol.io) tools, resources, and prompts for AI agents.
 
+[![Ask DeepWiki](https://deepwiki.com/badge.svg)](https://deepwiki.com/WordPress/mcp-adapter)
+
 ## Overview
 
 This adapter bridges WordPress's Abilities API with the [MCP specification](https://modelcontextprotocol.io/specification/2025-06-18/), providing a standardized way for AI agents to interact with WordPress functionality. It includes HTTP and STDIO transport support, comprehensive error handling, and an extensible architecture for custom integrations.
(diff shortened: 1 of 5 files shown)


galatanovidiu:
Closing: this feature cannot be implemented correctly at the mapper level — the DTO layer strips unknown fields during serialization.

PR #173, opened 2026-04-22
Guard DefaultServerFactory against cross-dispatch from other vendored copies

### LINKED ISSUE #172: `duplicate_server_id` when two plugins vendor this library
# Summary

When two or more active WordPress plugins each bundle `wordpress/mcp-adapter` via composer (with or without namespace prefixing), every callback subscribed to the global action `'mcp_adapter_init'` runs on every copy's `do_action( 'mcp_adapter_init', $this )` dispatch. The second dispatch re-enters servers already registered on the *same* adapter singleton, triggering `duplicate_server_id` errors.

Each REST request on an affected site produces repeated `_doing_it_wrong` notices like:

```
PHP Notice: Function create_server was called incorrectly.
  Server with ID "mcp-adapter-default-server" already exists.
  Each server must have a unique ID.
  (in wp-includes/functions.php on line 6131)
```

And, when the error propagates up through `DefaultServerFactory::create`:

```
PHP Notice: Function ...\WP\MCP\Servers\DefaultServerFactory::create was called incorrectly.
  MCP Adapter: Failed to create default server.
  Error: Server with ID "mcp-adapter-default-server" already exists.
  (Code: duplicate_server_id)
```

The same mechanism also affects any server that a consumer plugin registers via `add_action( 'mcp_adapter_init', ... )` — not just the library's hardcoded default server.

Reported in the wild: [r/woocommerce — "duplicate_server_id" blocking official WooCommerce MCP plugin](https://www.reddit.com/r/woocommerce/comments/1sra6dd/duplicate_server_id_blocking_official_woocommerce/).

# Affected versions

Confirmed in **v0.5.0** (current release at time of filing). Same mechanism exists in v0.4.x.

# Reproduction

1. Install two WordPress plugins, each of which composer-requires `wordpress/mcp-adapter` and ships its own `vendor/` directory. Optionally, one may prefix the namespace of its vendored copy to avoid class collisions — the bug fires either way.
2. Activate both plugins.
3. Ensure `WP_DEBUG` and `WP_DEBUG_LOG` are enabled.
4. Make any REST API request (e.g. `curl /wp-json/`) to trigger `rest_api_init`.
5. Inspect `wp-content/debug.log`. Expected: `duplicate_server_id` / "Server with ID ... already exists" notices, one per vendored copy, on every REST request.

# Root cause

All references below are to tag [`v0.5.0`](https://github.com/WordPress/mcp-adapter/tree/v0.5.0).

1. **`includes/Core/McpAdapter.php:81`** — adapter init fires a global, plain-string action:
   ```php
   do_action( 'mcp_adapter_init', $this );
   ```
2. **`includes/Core/McpAdapter.php:104`** — the default factory is subscribed to that same global action name inside `maybe_create_default_server()`:
   ```php
   add_action( 'mcp_adapter_init', array( DefaultServerFactory::class, 'create' ) );
   ```
3. **`includes/Servers/DefaultServerFactory.php:31`** — the default server ID is hardcoded:
   ```php
   'server_id' => 'mcp-adapter-default-server',
   ```
4. **`includes/Core/McpAdapter.php:180-193`** — re-registering the same ID raises `duplicate_server_id`.

Every vendored copy of this library — regardless of PHP namespace, because PHP namespaces do not apply to WordPress hook *names* (they are plain strings) — hooks its own `DefaultServerFactory::create` onto the shared `'mcp_adapter_init'` action. When adapter A fires the action during its `init()`, *both* factories (A's and B's) run. Each factory internally calls its own `McpAdapter::instance()` singleton, so A's factory registers the default server on A's adapter and B's factory registers it on B's adapter. Fine so far.

When adapter B later fires the same action during **its** `init()`, *both* factories run again. A's factory now tries to re-register `'mcp-adapter-default-server'` on A's adapter where it already exists → `duplicate_server_id`.

The same mechanism affects any server a consumer registers via `add_action( 'mcp_adapter_init', ... )`, not just the library's own default server: that consumer callback is re-entered on every other vendored copy's dispatch and collides on the same singleton.

The `'mcp_adapter_create_default_server'` filter (`McpAdapter.php:96`), which is the only current opt-out for the default server, has the identical cross-talk problem: if any one vendored copy sets `__return_false`, all copies see it and all default servers disappear. And it does nothing for consumer-registered servers.

# Impact

- Noisy `_doing_it_wrong` notices on every REST request in any multi-plugin setup using this library. Depending on `WP_DEBUG` / `WP_DEBUG_LOG` configuration, this fills error logs.
- Consumer-registered servers are affected, not just the library's default — any plugin that does `add_action( 'mcp_adapter_init', ... )` will see its callback re-entered on every other vendored copy's dispatch.
- User-visible bug reports blaming whichever plugin's `DefaultServerFactory` appears in the stack trace, even though the issue is library-level.
- Because the Abilities API shipped in WordPress 6.9 and PHP AI Client is merging in WP 7.0, more plugins are expected to vendor this library to expose abilities over MCP. The collision surface will grow.
- Not currently a functional blocker for correctly-configured consumers: the factory logs and returns, it doesn't halt, and plugins that register servers with unique IDs remain functional. But the log noise is real and the diagnostic trail misleads users.

arunshenoy99:
## Problem

Fixes #172.

When two or more active plugins each vendor `wordpress/mcp-adapter`, they collide on the shared `mcp_adapter_init` action. Every copy's dispatch runs every copy's callbacks, so each copy's `DefaultServerFactory::create` re-enters `create_server` on its own singleton, producing `duplicate_server_id` errors and a `_doing_it_wrong` notice on every REST request.

## Fix

`DefaultServerFactory::create` now accepts the dispatching adapter (passed by WordPress as the first argument to the `mcp_adapter_init` callback) and bails when it is not this copy's singleton. The parameter is intentionally untyped: in a multi-copy setup the dispatching adapter may be a different class (different namespace) than this factory's own `McpAdapter` import, so a strict type hint would raise a `TypeError` before the guard could run.

The change is backwards compatible: the parameter is optional, the `mcp_adapter_init` hook name is unchanged, and direct calls to `create()` (with no argument) behave exactly as before.

## Scope

This PR is intentionally limited to the factory guard, which fixes the crash on its own. The `McpAdapter::on_init()` consumer helper that was previously part of this PR has been split out into a separate PR so the new public API can be discussed and documented on its own.

## Tests

Two new unit tests cover the matching-adapter (registers the default server) and foreign-adapter (bails out) paths. Full unit suite passes locally (963 tests).

### DIFF
--- a/includes/Servers/DefaultServerFactory.php
+++ b/includes/Servers/DefaultServerFactory.php
@@ -31,9 +31,32 @@ class DefaultServerFactory {
 	 * WordPress filters for customization, making it perfect for use within
 	 * the McpAdapter.
 	 *
+	 * When hooked to the `mcp_adapter_init` action, WordPress passes the dispatching
+	 * adapter instance as the first argument. If that instance is not this copy's
+	 * singleton, the call bails out. This prevents duplicate server registration
+	 * when multiple copies of this library are active in the same request (e.g.
+	 * two plugins that each vendor `wordpress/mcp-adapter`, with or without
+	 * namespace prefixing) and each copy's dispatch fires every copy's callbacks.
+	 *
+	 * The parameter is intentionally untyped: in a multi-copy setup the dispatching
+	 * adapter may be a different class (different namespace) than this factory's
+	 * own `McpAdapter` import, so a strict type hint would raise a TypeError before
+	 * the guard could run.
+	 *
+	 * @since n.e.x.t Added optional `$dispatching_adapter` parameter and cross-dispatch guard.
+	 *
+	 * @param mixed $dispatching_adapter The dispatching adapter instance passed by WordPress.
+	 *                                   Null when invoked directly (not via the action).
+	 *
 	 * @return void
 	 */
-	public static function create(): void {
+	public static function create( $dispatching_adapter = null ): void {
+
+		// Bail when the dispatching adapter isn't this copy's singleton.
+		// See the docblock for why the parameter is untyped.
+		if ( null !== $dispatching_adapter && McpAdapter::instance() !== $dispatching_adapter ) {
+			return;
+		}
 
 		// Auto-discover resources and prompts from abilities
 		$auto_discovered_resources = self::discover_abilities_by_type( 'resource' );
--- a/tests/phpunit/Unit/Servers/DefaultServerFactoryTest.php
+++ b/tests/phpunit/Unit/Servers/DefaultServerFactoryTest.php
@@ -251,4 +251,34 @@ final class DefaultServerFactoryTest extends TestCase {
 		$this->assertNotNull( $server );
 		$this->assertInstanceOf( NullMcpObservabilityHandler::class, $server->get_observability_handler() );
 	}
+
+	public function test_create_with_matching_adapter_registers_server(): void {
+		global $wp_current_filter;
+		$wp_current_filter[] = 'mcp_adapter_init';
+
+		DefaultServerFactory::create( McpAdapter::instance() );
+
+		array_pop( $wp_current_filter );
+
+		$this->assertNotNull( $this->adapter->get_server( 'mcp-adapter-default-server' ) );
+	}
+
+	public function test_create_with_foreign_adapter_bails_out(): void {
+		global $wp_current_filter;
+		$wp_current_filter[] = 'mcp_adapter_init';
+
+		// Simulate a second adapter instance — the kind that would be dispatched
+		// by another vendored copy of this library in the same request.
+		$foreign = ( new \ReflectionClass( McpAdapter::class ) )->newInstanceWithoutConstructor();
+		$this->assertNotSame( McpAdapter::instance(), $foreign );
+
+		DefaultServerFactory::create( $foreign );
+
+		array_pop( $wp_current_filter );
+
+		$this->assertNull(
+			$this->adapter->get_server( 'mcp-adapter-default-server' ),
+			'Default server must not be registered on our singleton when a foreign adapter dispatched.'
+		);
+	}
 }


galatanovidiu:
Nice work on this, the fix is solid. Could we split it into two PRs?

The `DefaultServerFactory::create` guard fixes the `duplicate_server_id` crash on its own, and I'd like to land that one fast (it's basically there, just needs a rebase on trunk).

`on_init()` I'd keep separate. It's a new public method, so once it ships we're committed to it. And I'm not sure we need it: a consumer that uses the `$adapter` passed to its `mcp_adapter_init` callback is already safe. Worth its own PR + docs so we can talk it through.

Sound ok?

arunshenoy99:
Thanks @galatanovidiu! Done — split as suggested, and rebased on the latest `trunk`.

This PR is now scoped to just the `DefaultServerFactory::create` guard that fixes the `duplicate_server_id` crash on its own.

The new `McpAdapter::on_init()` helper has moved to its own PR — #217 — along with docs, so we can talk through the new public API separately. As you noted, a consumer that uses the `$adapter` passed to its `mcp_adapter_init` callback is already safe; `on_init()` just makes that guarantee explicit, so it's purely additive and not needed for the fix here.

galatanovidiu:
Closing this because we don't want to support multiple namespace-prefixed copies of MCP Adapter in the same request.

The collision is real in that setup. Prefixing the package creates a separate Adapter class and singleton, while `mcp_adapter_init` and the default server ID remain global. But this is not a supported way to use the package. MCP Adapter will ship as the canonical plugin, and plugins that bundle it should resolve the same `WP\MCP` implementation through Jetpack Autoloader.

The guard in this PR would only cover `DefaultServerFactory`. It would not make the other hooks, filters, routes, and global identifiers safe across prefixed copies. I don't think we should maintain a partial compatibility layer for that setup.

Issue #35, opened 2025-09-01
Abilities with empty input_schema behave weirdly on different applications.


jonathanbossenger:
Depending on the AI application you use to connect to MCP, Abilities registered with an empty `input_schema` array and then set to be adapted as tools may or may not work.

In VS Code, an Ability with an empty array for the `input_schema` and then passed to McpAdapter::create_server as a tool will be adapted correctly, the tool will be available to the AI agent.

In Cursor, the same Ability will throw the following error when attempting to list the available tools:

```
2025-09-01 14:37:19.006 [error] Error listing tools: [
  {
    "code": "invalid_type",
    "expected": "object",
    "received": "array",
    "path": [
      "tools",
      3,
      "inputSchema",
      "properties"
    ],
    "message": "Expected object, received array"
  }
]
```

jonathanbossenger:
@galatanovidiu a question on this, should Abilities that don't require any input be registered as tools or resources? when I first encountered this issue I thought maybe they should be adapted as resources, not tools, but doing that doesn't seem to expose them to either VS code or Cursor.

galatanovidiu:
Looking at this issue, the quick solution would be to return an empty object {} instead of an empty array [].
However, this will need testing to ensure it doesn't break existing functionality.

**Important note about resources vs tools:**

Resources use a different format than tools and require specific metadata fields. Example ability:

```php
...
'meta' => array(
    'mimeType' => 'application/json',
    'uri'      => 'core://wordpress/site-info',
),
```

The uri and mimeType fields are required for proper resource definition in the MCP specification.

galatanovidiu:
It looks like we are already trying to do this, but it does not work 
https://github.com/WordPress/mcp-adapter/blob/trunk/includes/Domain/Tools/McpTool.php#L278

jonathanbossenger:
> The uri and mimeType fields are required for proper resource definition in the MCP specification.

Noted, thanks. I'll check to make sure this is also recorded in the docs

jonathanbossenger:
Fixed with #36 and #37

PR #174, opened 2026-04-23
Fix: Add placeholder property for parameterless ability schemas

lbbms:
## Summary
This PR updates `SchemaTransformer` to make parameterless ability schemas compatible with stricter MCP clients (notably Anthropic integrations) that reject `type: object` schemas when `properties` is missing.

### 1. Added a placeholder property constant
- Introduced `private const NOOP_PROPERTY = '_mcp_noop';`
- Used as an optional, no-op placeholder key for tools that otherwise have no input fields

### 2. Normalized empty/object schemas to always include properties
Changed all object-schema return paths in `transform_to_object_schema()` to go through a new helper:
- Empty schema case (`empty($schema)`)
- No explicit type case (default to `type: object`)
- Already object schema case (`type === 'object'`)

All now call: `self::ensure_object_properties($schema)`

### 3. Added `ensure_object_properties()` helper
New private method with the following behavior:
- If schema already has non-empty `properties` (array), it is returned unchanged
- Otherwise, injects:
  - `properties._mcp_noop` with:
    - `type: string`
    - `description: "Optional no-op placeholder for parameterless tools."`
- Ensures `required` exists and is an array
- Does **not** require `_mcp_noop` (so no real behavioral change for callers)

### 4. Why this approach (instead of `stdClass` in schema)
A previous approach used `new \stdClass()` for empty properties, which helps JSON encoding but can be risky in internal DTO/validation paths expecting arrays.

This patch keeps everything array-based and client-safe:
- `inputSchema.properties` is always present
- DTO and validator flow remains consistent
- No-op property is optional, so tool execution semantics are unchanged

## Compatibility Impact
✅ Fixes clients that require object schemas to include `properties`  
✅ Keeps existing tools with real input schemas untouched  
✅ Parameterless tools now expose one optional placeholder input key, which is harmless and ignored unless explicitly sent

## Rationale
This improves interoperability with strict MCP client schema validation while preserving adapter internals and tool behavior. It specifically addresses cases where parameterless WordPress abilities were transformed into `{ type: "object" }` without properties, causing client-side tool registration failures.

### DIFF
--- a/includes/Domain/Utils/SchemaTransformer.php
+++ b/includes/Domain/Utils/SchemaTransformer.php
@@ -19,6 +19,13 @@ namespace WP\MCP\Domain\Utils;
  */
 class SchemaTransformer {
 	/**
+	 * No-op property placeholder for tools without input fields.
+	 *
+	 * @var string
+	 */
+	private const NOOP_PROPERTY = '_mcp_noop';
+
+	/**
 	 * Transform a schema to MCP-compatible object format.
 	 *
 	 * If the schema is already an object type, it is returned unchanged.
@@ -39,8 +46,10 @@ class SchemaTransformer {
 		// Handle null or empty schema - return minimal valid MCP object schema.
 		if ( empty( $schema ) ) {
 			return array(
-				'schema'           => array(
-					'type' => 'object',
+				'schema'           => self::ensure_object_properties(
+					array(
+						'type' => 'object',
+					)
 				),
 				'was_transformed'  => false,
 				'wrapper_property' => null,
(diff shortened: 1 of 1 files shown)


galatanovidiu:
Thanks @lbbms, and for the detailed write-up. This is the same problem as #35: parameterless tools get rejected by strict clients (Cursor, OpenAI strict mode) because their `inputSchema` has no `properties` object.

I went with a fix at the serialization layer instead. #212 bumps `wordpress/php-mcp-schema` to v0.1.2, where `ToolInputSchema::toArray()` and `ToolOutputSchema::toArray()` always emit `properties` as a JSON object. So a parameterless tool now serializes as `{"type":"object","properties":{}}`, which strict validators accept. `McpToolValidator` is updated to accept that empty `{}` too.

I prefer this route because it keeps the tool contract honest. No placeholder param shown to clients, prompts and output schemas stay untouched, and it doesn't change the `SchemaTransformer` behavior the unit tests pin. The `_mcp_noop` approach would advertise a phantom argument on every parameterless tool, and it also leaks into prompt arguments and output schemas.

Closing in favor of #212. Thanks again for digging into this, the investigation helped confirm the root cause. 🙏

Issue #178, opened 2026-05-01
[Tracking 📌] Path to WordPress.org Plugin Directory


justlevine:
With submission to W.ORG approved, it's time to get this repo polished and ready for submission.

This is a _nonexhaustive_ list of steps that need to happen:

## TODO

- [x] Align plugin scaffolding to WordPress/ai (in particular, there's some required files missing like readme.txt) @justlevine 
- [x] Run plugin-check and remediate exposed issues #187 
- [x] Bump minimum WordPress to v6.9 #197 
- [ ] Cleanup and update install docs to prioritize usage via plugin (and wpackagist/wp-packages for Composer)
- [ ] Make plugin backwards-compatible with Composer Library + migration docs (May require an intermediary) 
- [ ] Optional: Enforce SemVer auto-updates protection.
- [ ] Submit to W.ORG

_Comment below if I'm missing something, and I'll add it to the list_

justlevine:
> Optional: Enforce SemVer auto-updates protection.

@galatanovidiu This question is for you. Are we expecting to need to issue future breaking changes (e.g. updating/dropping schema versions?)

If so, then I'll add a modal warning to prevent autoupdates to major versions. If you think we're ready to keep things backwards, we can skip this and only introduce if the need presents itself in the future.

galatanovidiu:
> > Optional: Enforce SemVer auto-updates protection.
> 
> [@galatanovidiu](https://github.com/galatanovidiu) This question is for you. Are we expecting to need to issue future breaking changes (e.g. updating/dropping schema versions?)
> 
> If so, then I'll add a modal warning to prevent autoupdates to major versions. If you think we're ready to keep things backwards, we can skip this and only introduce if the need presents itself in the future.

I'd skip it for now. The schema package is still additive (0.1.1), the spec move from 2025-06-18 to 2025-11-25 was additive only, and the adapter already negotiates all three protocol versions side by side, so older clients keep working. If we ever hit a real break that forces a major bump, we can add the modal then with a concrete context for users.

justlevine:
Noting: as part of updating docs/cleanup we need to explicitly bump minimum WP to 6.9 (tooling+ci were bumped in #185, ~#189~ #193 )

andreilupu:
Hi @justlevine 👋 I'd like to help push this further, and I'd be happy to put some time and energy into it. Engineering, testing, docs, or anything else, as long as it's helpful.

gziolo:
[...]

PR #193, opened 2026-05-19
test: deslop and fix PHPUnit Tests

justlevine:
## What

This PR relocates PHPUnit tests to `tests/phpunit`, then deslops the internals so tests pass.

(Replaces #189 which was auto-closed instead of being reapplied to `trunk` )

## Why

Tests have been failing due to compounding tech-debt around the `TestCase`.

- Part of #178 

## How

- Updated `TestCase` to extend the correct `WP_UnitTestCase`, and use the built-in methods.
- Fixed an error caused by calling a method from a trait, instead of from a class implementing the trait.
- Cleaned up the various filter-stubbing and generation slop. Not everything, just what was getting in the way of fixing the tests.
- Moved problematic and test-case-specific abilities out of DummyAbility and into the test cases that use them.
- Added `setExpectedIncorrectUsage()` to the tests that should have it.

(The goal was not to audit all of the existing tests, just to fix the central issues causing all the failures).


includes/Transport/Infrastructure/RequestRouter.php
```diff
@@ -289,7 +289,7 @@ private function sanitize_params_for_logging( array $params ): array {
 			// Filter argument keys to exclude sensitive-looking ones.
 			$safe_keys = array();
 			foreach ( array_keys( $params['arguments'] ) as $arg_key ) {
-				if ( McpObservabilityHelperTrait::is_sensitive_key( (string) $arg_key ) ) {
+				if ( ErrorLogMcpObservabilityHandler::is_sensitive_key( (string) $arg_key ) ) {
```
galatanovidiu, line 293: We will need to clean this up at some point.

galatanovidiu (approved):
Well-deserved cleanup 👍 
I have not gone through all the changes, but from what I've seen, all looks good.
Thank you for taking the time to do this 🙏

includes/Transport/Infrastructure/RequestRouter.php
justlevine: I'll add a `// @todo` comment inline so this doesn't fall through the cracks 👍

justlevine: f8e4c03d6366128a0d086237ce5f5208ed2fca22

PR #201, opened 2026-05-29
Add mcp_adapter_server_config filter and fix default ability hook timing

### LINKED ISSUE #117: Default abilities fail to register: `wp_abilities_api_init` hook fires before adapter initializes
## Bug Description

The default MCP adapter abilities (`mcp-adapter/discover-abilities`, `mcp-adapter/get-ability-info`, `mcp-adapter/execute-ability`) fail to register because the ability registration hooks are added too late.

## Expected Behavior

When the plugin loads, the default abilities should register successfully via the `wp_abilities_api_init` hook.

## Actual Behavior

The abilities are NOT registered because:
1. `McpAdapter::instance()` defers initialization to `rest_api_init` (line 64)
2. [maybe_create_default_server()](cci:1://file:///Users/ehjewel/Local%20Sites/develop/app/public/wp-content/plugins/mcp-adapter-0.4.1/includes/Core/McpAdapter.php:267:1-282:2) is called during [init()](cci:1://file:///Users/ehjewel/Local%20Sites/develop/app/public/wp-content/plugins/mcp-adapter-0.4.1/includes/Core/McpAdapter.php:89:2-104:2) which runs on `rest_api_init`
3. But `wp_abilities_api_init` fires during the [init](cci:1://file:///Users/ehjewel/Local%20Sites/develop/app/public/wp-content/plugins/mcp-adapter-0.4.1/includes/Core/McpAdapter.php:89:2-104:2) hook (via `WP_Abilities_Registry::get_instance()`)
4. Since `rest_api_init` fires AFTER [init](cci:1://file:///Users/ehjewel/Local%20Sites/develop/app/public/wp-content/plugins/mcp-adapter-0.4.1/includes/Core/McpAdapter.php:89:2-104:2), the `add_action('wp_abilities_api_init', ...)` calls are too late

## Error in Debug Log
[ERROR] WordPress ability 'mcp-adapter/discover-abilities' does not exist. 
[ERROR] WordPress ability 'mcp-adapter/get-ability-info' does not exist. 
[ERROR] WordPress ability 'mcp-adapter/execute-ability' does not exist.

## Root Cause

In [includes/Core/McpAdapter.php](cci:7://file:///Users/ehjewel/Local%20Sites/develop/app/public/wp-content/plugins/mcp-adapter-0.4.1/includes/Core/McpAdapter.php:0:0-0:0), the [instance()](cci:1://file:///Users/ehjewel/Local%20Sites/develop/app/public/wp-content/plugins/mcp-adapter-0.4.1/includes/Core/McpAdapter.php:49:1-72:2) method schedules initialization on `rest_api_init`:

```php
add_action( 'rest_api_init', array( self::$instance, 'init' ), 15 );
```

But the wp_abilities_api_init hook fires during init, which happens BEFORE rest_api_init.

## Proposed Fix
Move the ability registration hooks out of maybe_create_default_server() and into a new method register_ability_hooks() that's called immediately during instance():
```php
public static function instance(): self {
        self::$instance = new self();

        // Register ability hooks IMMEDIATELY so they fire before wp_abilities_api_init
        self::$instance->register_ability_hooks();

        if ( defined( 'WP_CLI' ) && constant( 'WP_CLI' ) ) {
            add_action( 'init', array( self::$instance, 'init' ), 20 );
        } else {
            add_action( 'rest_api_init', array( self::$instance, 'init' ), 15 );
        }
    }
    return self::$instance;
}

private function register_ability_hooks(): void {
    if ( ! apply_filters( 'mcp_adapter_create_default_server', true ) ) {
        return;
    }
    add_action( 'wp_abilities_api_categories_init', array( $this, 'register_default_category' ) );
    add_action( 'wp_abilities_api_init', array( $this, 'register_default_abilities' ) );
}
```

## Environment
- WordPress: 6.9.0
- MCP Adapter: 0.4.1
- PHP: 7.4+

## Steps to Reproduce
1. Install MCP Adapter plugin v0.4.1
2. Activate the plugin
3. Make a REST API request to /wp-json/mcp/mcp-adapter-default-server
4. Check wp-content/debug.log for errors about missing abilities

### LINKED ISSUE #190: Add mcp_adapter_server_config filter to McpServer constructor for all servers
## Summary

There is currently no filter that allows modifying the configuration of **all** MCP servers at construction time. The existing `mcp_adapter_default_server_config` filter only covers the default server created by `DefaultServerFactory`.

## Problem

When `McpServer::__construct()` is called — whether for the default server or any custom server created via `mcp_adapter_init` — there is no way to alter its configuration (server name, route namespace, tools, resources, prompts, etc.) without forking the creation logic.

This makes it impossible for plugins to:
- Add or remove abilities from any server's tools/resources/prompts list
- Rename or reroute a server they did not create
- Apply site-wide rules (e.g. visibility allowlists) across all servers uniformly

## Proposed Solution

Add a single filter at the top of `McpServer::__construct()`, before parameters are assigned to properties:

```php
/**
 * Filters the configuration of any MCP server before it is applied.
 *
 * Fires for ALL servers during construction. Use $config['server_id']
 * to target a specific server or apply changes to all servers.
 *
 * @since 0.6.0
 *
 * @param array  $config {
 *     Server configuration array.
 *
 *     @type string   $server_id              Server identifier.
 *     @type string   $server_route_namespace REST API namespace.
 *     @type string   $server_route           REST API route.
 *     @type string   $server_name            Human-readable server name.
 *     @type string   $server_description     Server description.
 *     @type string   $server_version         Server version.
 *     @type string[] $tools                  Ability names to expose as tools.
 *     @type string[] $resources              Ability names to expose as resources.
 *     @type string[] $prompts                Ability names to expose as prompts.
 * }
 * @param string $server_id The server identifier (for convenience — same as $config['server_id']).
 */
$config = apply_filters( 'mcp_adapter_server_config', $config, $server_id );
```

The filter fires inside `__construct()` after parameters are packed into a `$config` array and before they are unpacked into instance properties. The rest of the constructor continues unchanged.

## Location

`includes/Core/McpServer.php` — top of `__construct()`, before line 144 where properties are assigned.

## Use Case Example

```php
add_filter( 'mcp_adapter_server_config', function( array $config, string $server_id ) {
    // Remove a specific ability from tools on all servers.
    $config['tools'] = array_diff( $config['tools'], array( 'core/get-environment-info' ) );

    // Rename a specific server.
    if ( 'mcp-adapter-default-server' === $server_id ) {
        $config['server_name'] = 'My Custom Server Name';
    }

    return $config;
}, 10, 2 );
```

## Why One Filter Instead of Many

A single filter receiving the full config array is preferred over per-parameter filters (`mcp_adapter_server_tools`, `mcp_adapter_server_name`, etc.) because:
- Single hook to learn and maintain
- Tools, resources, and prompts are often modified together based on the same logic
- `$server_id` as the second parameter allows targeting specific or all servers
- Consistent shape with the existing `mcp_adapter_default_server_config` filter

## Relationship to `mcp_adapter_default_server_config`

`mcp_adapter_default_server_config` fires inside `DefaultServerFactory` before the default server is constructed — it shapes the config before it reaches `McpServer::__construct()`. The proposed `mcp_adapter_server_config` fires inside `__construct()` itself, covering all servers regardless of how they were created. The two filters are complementary, not redundant.

ibrahimhajjaj:
## What?

Closes #190
Closes #117

Two related changes to server construction and the default server's ability registration:

1. Adds an `mcp_adapter_server_config` filter inside `McpServer::__construct()` so plugins can shape the config of any server (not just the default one) before it is applied.
2. Moves the default abilities' registration hooks so they are wired at `McpAdapter::instance()` time instead of during `rest_api_init`, fixing the default abilities silently failing to register.

## Why?

**#190** — The existing `mcp_adapter_default_server_config` filter only covers the default server created by `DefaultServerFactory`. There is no way to alter the config of a custom server created via `mcp_adapter_init`, or to apply a site-wide rule across all servers, without forking the creation logic. Plugins want to add/remove abilities from a server's tools, rename or reroute a server they did not create, or apply visibility allowlists uniformly.

**#117** — The three default abilities (`mcp-adapter/discover-abilities`, `get-ability-info`, `execute-ability`) fail to register. `McpAdapter::instance()` defers `init()` to `rest_api_init`, but `wp_abilities_api_init` fires during the earlier `init` action. The `add_action()` calls for ability registration lived inside `maybe_create_default_server()`, which runs on `rest_api_init`, so they were added too late to catch the abilities-api init pass, producing `WordPress ability '...' does not exist` errors.

## How?

**#190** — `__construct()` now packs the host-facing parameters (`server_id`, route namespace/route, name, description, version, tools, resources, prompts) into a `$config` array, applies `mcp_adapter_server_config` with `$server_id` as the second arg, then assigns from the filtered result. Runtime-critical wiring (transport classes, error/observability handlers, the transport permission callback) is intentionally left out of the filter. A non-array return or dropped keys fall back to the constructor-supplied values via `wp_parse_args`, and the array fields are normalized with `array_values()` to preserve their `list<string>` shape.

**#117** — Ability registration hooks move into a new `register_default_ability_hooks()` method called directly from `instance()`, so they are wired before `init` fires. The method is gated by the same `mcp_adapter_create_default_server` filter, so disabling the default server still skips registering its abilities. `maybe_create_default_server()` keeps the `mcp_adapter_init` server-creation hook, which has no timing issue.

## Testing Instructions

1. `composer install && npm install && npm run wp-env start`
2. `npm run test:php` — full suite passes (1008 tests).
3. For #117: with the plugin active, confirm the debug log no longer shows `WordPress ability 'mcp-adapter/discover-abilities' does not exist` and the default server's tools resolve.
4. For #190: hook `mcp_adapter_server_config`, mutate `$config['server_name']` (or `tools`), and confirm the change applies to a custom server, not just the default.

New tests:
- `tests/phpunit/Integration/WordPressFiltersTest.php` — filter rewrites fields, receives `$server_id`, falls back on non-array/dropped keys, fires once per server.
- `tests/phpunit/Unit/Core/McpAdapterDefaultAbilityHooksTest.php` — ability hooks are registered at `instance()` time.

## Changelog Entry

> Added - `mcp_adapter_server_config` filter to customize any MCP server's configuration at construction time.
> Fixed - Default abilities failing to register because their hooks were added after `wp_abilities_api_init` had fired.

### DIFF
--- a/includes/Core/McpAdapter.php
+++ b/includes/Core/McpAdapter.php
@@ -58,6 +58,12 @@ final class McpAdapter {
 		if ( ! isset( self::$instance ) ) {
 			self::$instance = new self();
 
+			// Wire the default abilities now so they catch `wp_abilities_api_init`,
+			// which fires during the `init` action, before `rest_api_init` runs.
+			// Deferred initialization handles transports and the `mcp_adapter_init`
+			// action, neither of which need to be in place before `init`.
+			self::$instance->register_default_ability_hooks();
+
 			// In WP-CLI context, initialize immediately so commands have access to servers
 			if ( defined( 'WP_CLI' ) && constant( 'WP_CLI' ) ) {
 				add_action( 'init', array( self::$instance, 'init' ), 20 );
@@ -68,6 +74,32 @@ final class McpAdapter {
 		}
 
 		return self::$instance;
+	}
+
+	/**
+	 * Wire the default server's ability registration hooks at instance() time.
+	 *
+	 * `wp_abilities_api_categories_init` and `wp_abilities_api_init` fire during
+	 * WordPress's `init` action. Because `instance()` is typically called during
+	 * plugin bootstrap (before `init`), wiring these handlers here ensures they
+	 * catch the abilities-api init pass. The previous placement inside
+	 * {@see self::maybe_create_default_server()} ran on `rest_api_init`, which is
+	 * after `init`, so the default abilities never registered for the abilities
+	 * API to discover.
+	 *
+	 * Gated by `mcp_adapter_create_default_server` so a host disabling the
+	 * default server does not register default abilities either.
+	 *
+	 * @internal For use by adapter initialization only.
+	 */
+	private function register_default_ability_hooks(): void {
+		/** This filter is documented in {@see self::maybe_create_default_server()}. */
+		if ( ! apply_filters( 'mcp_adapter_create_default_server', true ) ) {
+			return;
+		}
+
+		add_action( 'wp_abilities_api_categories_init', array( $this, 'register_default_category' ) );
+		add_action( 'wp_abilities_api_init', array( $this, 'register_default_abilities' ) );
 	}
 
 	/**
@@ -118,10 +150,9 @@ final class McpAdapter {
 			return;
 		}
 
-		// Register category before abilities
-		add_action( 'wp_abilities_api_categories_init', array( $this, 'register_default_category' ) );
-		add_action( 'wp_abilities_api_init', array( $this, 'register_default_abilities' ) );
-
+		// Ability registration hooks are wired earlier in
+		// {@see self::register_default_ability_hooks()} so they catch
+		// `wp_abilities_api_init`. This stage only schedules server creation.
 		add_action( 'mcp_adapter_init', array( DefaultServerFactory::class, 'create' ) );
 	}
 
--- a/includes/Core/McpServer.php
+++ b/includes/Core/McpServer.php
@@ -143,13 +143,68 @@ class McpServer {
 		array $prompts = array(),
 		?callable $transport_permission_callback = null
 	) {
-		// Store server configuration
-		$this->server_id                     = $server_id;
-		$this->server_route_namespace        = $server_route_namespace;
-		$this->server_route                  = $server_route;
-		$this->server_name                   = $server_name;
-		$this->server_description            = $server_description;
-		$this->server_version                = $server_version;
+		// Pack the host-facing configuration so plugins can shape any
+		// server (not just the default) before properties are assigned.
+		$config = array(
+			'server_id'              => $server_id,
+			'server_route_namespace' => $server_route_namespace,
+			'server_route'           => $server_route,
+			'server_name'            => $server_name,
+			'server_description'     => $server_description,
+			'server_version'         => $server_version,
+			'tools'                  => $tools,
+			'resources'              => $resources,
+			'prompts'                => $prompts,
+		);
+
+		/**
+		 * Filters the configuration of an MCP server before properties are assigned.
+		 *
+		 * Fires for every server during construction, allowing site-wide customization
+		 * across the default server and any plugin-created servers. Use the
+		 * `$server_id` parameter to target a specific server, or apply changes
+		 * uniformly across all servers.
+		 *
+		 * Runtime-critical wiring is intentionally outside this filter:
+		 * transport classes, error/observability handlers, and the transport
+		 * permission callback are not exposed here.
+		 *
+		 * @since 0.6.0
+		 *
+		 * @param array  $config {
+		 *     Server configuration array.
+		 *
+		 *     @type string   $server_id              Server identifier.
+		 *     @type string   $server_route_namespace REST API namespace.
+		 *     @type string   $server_route           REST API route.
+		 *     @type string   $server_name            Human-readable server name.
+		 *     @type string   $server_description     Server description.
+		 *     @type string   $server_version         Server version.
+		 *     @type string[] $tools                  Ability names to expose as tools.
+		 *     @type string[] $resources              Ability names to expose as resources.
+		 *     @type string[] $prompts                Ability names to expose as prompts.
+		 * }
+		 * @param string $server_id The server identifier (matches `$config['server_id']`).
+		 */
+		$filtered = apply_filters( 'mcp_adapter_server_config', $config, $server_id );
+
+		// Tolerate filters that drop keys or return a non-array by falling
+		// back to the constructor-supplied defaults.
+		if ( ! is_array( $filtered ) ) {
+			$filtered = $config;
+		}
+		$filtered = wp_parse_args( $filtered, $config );
+
+		// Store server configuration from the (possibly filtered) config.
+		$this->server_id                     = (string) $filtered['server_id'];
+		$this->server_route_namespace        = (string) $filtered['server_route_namespace'];
+		$this->server_route                  = (string) $filtered['server_route'];
+		$this->server_name                   = (string) $filtered['server_name'];
+		$this->server_description            = (string) $filtered['server_description'];
+		$this->server_version                = (string) $filtered['server_version'];
+		$tools                               = array_values( (array) $filtered['tools'] );
+		$resources                           = array_values( (array) $filtered['resources'] );
+		$prompts                             = array_values( (array) $filtered['prompts'] );
 		$this->transport_permission_callback = $transport_permission_callback;
 
 		/**
--- a/tests/phpunit/Unit/Core/McpAdapterDefaultAbilityHooksTest.php
+++ b/tests/phpunit/Unit/Core/McpAdapterDefaultAbilityHooksTest.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace WP\MCP\Tests\Unit\Core;
+
+use WP\MCP\Core\McpAdapter;
+use WP\MCP\Tests\TestCase;
+
+/**
+ * Regression coverage for the timing of the default server's ability hooks.
+ *
+ * `wp_abilities_api_categories_init` and `wp_abilities_api_init` fire during
+ * the WordPress `init` action. Wiring their handlers inside the deferred
+ * `rest_api_init` initialization missed both, leaving the default abilities
+ * unregistered (see issue #117). The hooks are now wired during `instance()`
+ * so they catch the abilities-api init pass.
+ */
+final class McpAdapterDefaultAbilityHooksTest extends TestCase {
+
+	public function test_default_ability_hooks_are_registered_at_instance_time(): void {
+		$adapter = McpAdapter::instance();
+
+		$this->assertNotFalse(
+			has_action(
+				'wp_abilities_api_categories_init',
+				array( $adapter, 'register_default_category' )
+			),
+			'register_default_category should be hooked to wp_abilities_api_categories_init at instance() time'
+		);
+
+		$this->assertNotFalse(
+			has_action(
+				'wp_abilities_api_init',
+				array( $adapter, 'register_default_abilities' )
+			),
+			'register_default_abilities should be hooked to wp_abilities_api_init at instance() time'
+		);
+	}
+}
--- a/tests/phpunit/Integration/WordPressFiltersTest.php
+++ b/tests/phpunit/Integration/WordPressFiltersTest.php
@@ -30,4 +30,154 @@ final class WordPressFiltersTest extends TestCase {
 
 		remove_filter( 'mcp_adapter_validation_enabled', '__return_false' );
 	}
+
+	public function test_server_config_filter_can_rewrite_string_fields(): void {
+		add_filter(
+			'mcp_adapter_server_config',
+			static function ( array $config ): array {
+				$config['server_id']          = 'rewritten-id';
+				$config['server_name']        = 'Rewritten Name';
+				$config['server_description'] = 'Rewritten description';
+				return $config;
+			}
+		);
+
+		$server = new McpServer(
+			'original-id',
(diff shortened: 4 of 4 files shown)


raftaar1191:
Thanks for picking this up, @ibrahimhajjaj. The implementation is cleaner than what I'd sketched in #190 — the per-server fan-out plus the `wp_parse_args` fallback for non-array returns is good defensive design.

gziolo:
Just noting that I commented on the issue https://github.com/WordPress/mcp-adapter/issues/117#issuecomment-4978075925 asking whether it's possible to register abilities and the category outside of other actions. It looks like this PR exercises exactly that approach 👍🏻

includes/Core/McpServer.php
```diff
@@ -143,13 +143,68 @@ public function __construct(
...
+
+		// Tolerate filters that drop keys or return a non-array by falling
+		// back to the constructor-supplied defaults.
+		if ( ! is_array( $filtered ) ) {
+			$filtered = $config;
+		}
+		$filtered = wp_parse_args( $filtered, $config );
+
+		// Store server configuration from the (possibly filtered) config.
+		$this->server_id                     = (string) $filtered['server_id'];
```
galatanovidiu, line 199: The filter can rewrite `server_id`, but by then `McpAdapter::create_server()` has already used the original id: both the duplicate check (McpAdapter.php:273) and the registry entry `$this->servers[ $server_id ] = $server` (McpAdapter.php:334) use the constructor argument, not the filtered value. After a rewrite, `get_server( $server->get_server_id() )` returns null, and `wp mcp-adapter list` prints an id that `serve --server=<id>` cannot find. Could we keep `server_id` out of the filterable keys? The filter's second argument already lets callbacks target a specific server. If renaming must stay supported, the filter would need to run in `create_server()` before the duplicate check, so the filtered id is the one that gets validated and stored.

includes/Core/McpAdapter.php
```diff
@@ -70,6 +76,32 @@ public static function instance(): self {
...
+	 * API to discover.
+	 *
+	 * Gated by `mcp_adapter_create_default_server` so a host disabling the
+	 * default server does not register default abilities either.
+	 *
+	 * @internal For use by adapter initialization only.
+	 */
+	private function register_default_ability_hooks(): void {
+		/** This filter is documented in {@see self::maybe_create_default_server()}. */
+		if ( ! apply_filters( 'mcp_adapter_create_default_server', true ) ) {
```
galatanovidiu, line 97: `instance()` runs when the plugin file loads, so this filter is checked before themes and later-loading plugins can hook it: a site that disables the default server from a theme still gets the category and the three default abilities registered, even though `docs/guides/default-server.md` promises they are skipped. The second check in `maybe_create_default_server()` still stops the server itself, but when the two checks disagree the other way you can get a default server whose abilities never registered, the same failure #117 fixed. What if we wire the two `add_action` calls unconditionally and check the filter inside `register_default_category()` and `register_default_abilities()`? Every evaluation then runs after plugins and the theme can hook the filter, and evaluating it once through a small cached helper would close the remaining gap between `init` and `rest_api_init`.

tests/phpunit/Unit/Core/McpAdapterDefaultAbilityHooksTest.php
```diff
@@ -0,0 +1,40 @@
...
+ * `rest_api_init` initialization missed both, leaving the default abilities
+ * unregistered (see issue #117). The hooks are now wired during `instance()`
+ * so they catch the abilities-api init pass.
+ */
+final class McpAdapterDefaultAbilityHooksTest extends TestCase {
+
+	public function test_default_ability_hooks_are_registered_at_instance_time(): void {
+		$adapter = McpAdapter::instance();
+
+		$this->assertNotFalse(
```
galatanovidiu, line 24: This test passes even with the fix reverted: `TestCase::set_up_before_class()` (tests/phpunit/TestCase.php:43-44) adds the same callbacks on the same singleton, so `has_action()` matches the test-base wiring, not the production one. I confirmed by reverting the McpAdapter change and running it; both assertions still pass. Could we drop that TestCase wiring (the production code now does it at `instance()` time) and assert the outcome instead, for example that the three default abilities exist via `wp_get_ability()` after `init`? The filter-false path of `register_default_ability_hooks()` also has no test; it needs an isolated bootstrap (`@runInSeparateProcess`) because the singleton already exists before any test can add the filter.

includes/Core/McpServer.php
```diff
@@ -143,13 +143,68 @@ public function __construct(
...
+		 * Fires for every server during construction, allowing site-wide customization
+		 * across the default server and any plugin-created servers. Use the
+		 * `$server_id` parameter to target a specific server, or apply changes
+		 * uniformly across all servers.
+		 *
+		 * Runtime-critical wiring is intentionally outside this filter:
+		 * transport classes, error/observability handlers, and the transport
+		 * permission callback are not exposed here.
+		 *
+		 * @since 0.6.0
```
galatanovidiu, line 172: #190's sketch had `0.6.0`, but we use `@since n.e.x.t` for new code; the real version is filled in at release.
```suggestion
		 * @since n.e.x.t
```

includes/Core/McpServer.php
```diff
@@ -143,13 +143,68 @@ public function __construct(
...
+		 *     @type string   $server_name            Human-readable server name.
+		 *     @type string   $server_description     Server description.
+		 *     @type string   $server_version         Server version.
+		 *     @type string[] $tools                  Ability names to expose as tools.
+		 *     @type string[] $resources              Ability names to expose as resources.
+		 *     @type string[] $prompts                Ability names to expose as prompts.
+		 * }
+		 * @param string $server_id The server identifier (matches `$config['server_id']`).
+		 */
+		$filtered = apply_filters( 'mcp_adapter_server_config', $config, $server_id );
```
galatanovidiu, line 189: Could we document the new filter before this ships? `docs/guides/default-server.md` documents `mcp_adapter_default_server_config`, and the default server now runs through both filters with near-identical names, so the order needs stating somewhere. The same guide (line 73) also says the disable filter is applied only in `McpAdapter::maybe_create_default_server()`, while this PR adds an earlier check in `instance()`.

galatanovidiu (changes requested):
The no-filter path is safe: I replayed the new normalization and the output is identical to the old assignments. Three things need rework before merge, details inline:

- The filter can rewrite `server_id` after the registry stored the server under the original one. `wp mcp-adapter list` then shows an id `serve --server=` can't resolve.
- `mcp_adapter_create_default_server` is now checked at plugin load and again at `rest_api_init`. A disable filter added from a theme or a later plugin only catches the second check, so the default abilities register anyway.
- The new hook-timing test passes even with the fix reverted (the TestCase base wires the same hooks).

Non-blocking, worth doing in the same pass:
- `@since n.e.x.t` instead of `0.6.0` (repo convention; I know the #190 sketch had 0.6.0).
- The new filter is not in the docs, and `docs/guides/default-server.md` still describes only the old evaluation of the disable filter.
- `mcp.server.created` observability still reports the pre-filter id and counts.
- No test exercises filtering `tools`/`resources`/`prompts`, the headline use case of #190.

PHPCS and PHPStan green. Codecov flags 6 lines: 4 only run at suite bootstrap so coverage can't see them, and the filter-false branch is genuinely untested (see the test comment). The Plugin Check CI failure is the action's own env setup, not this change.

<sub>Reviewed by Claude and Codex, checked by me.</sub>

mdjwel:
> The following accounts have interacted with this PR and/or linked issues. I will continue to update these lists as activity occurs. You can also manually ask me to refresh this list by adding the `props-bot` label.
> 
> ## Unlinked Accounts
> The following contributors have not linked their GitHub and WordPress.org accounts: @gelakhubutia, @scottblackburn, @mdjwel.
> 
> Contributors, please [read how to link your accounts](https://make.wordpress.org/core/2020/03/19/associating-github-accounts-with-wordpress-org-profiles/) to ensure your work is properly credited in WordPress releases.
> 
> If you're merging code through a pull request on GitHub, copy and paste the following into the bottom of the merge commit message.
> 
> ```
> Unlinked contributors: gelakhubutia, scottblackburn, mdjwel.
> 
> Co-authored-by: ibrahimhajjaj <ibrahimhajjaj@git.wordpress.org>
> Co-authored-by: raftaar1191 <raftaar1191@git.wordpress.org>
> Co-authored-by: gziolo <gziolo@git.wordpress.org>
> Co-authored-by: chubes4 <extrachill@git.wordpress.org>
> ```
> 
> **To understand the WordPress project's expectations around crediting contributors, please [review the Contributor Attribution page in the Core Handbook](https://make.wordpress.org/core/handbook/best-practices/contributor-attribution-props/).**

My WordPress.org account is linked to this GitHub profile, but for the sake of clarity, here is my profile link: https://profiles.wordpress.org/mdjwel/

includes/Core/McpServer.php
justlevine: (As the current release is already at 0.6.x this is now blocking)

PR #217, opened 2026-06-17
Add McpAdapter::on_init() helper for safe server registration

arunshenoy99:
## Summary

Adds `McpAdapter::on_init()`, a small convenience wrapper around `add_action( 'mcp_adapter_init', ... )` that builds an adapter-identity check in for the caller.

This was split out of #173 (per review feedback) so the new public API can be discussed and documented on its own. The `duplicate_server_id` crash itself is fixed independently by the factory guard in #173; this helper is an ergonomic addition, not required for that fix.

## Why

The `mcp_adapter_init` action name is global. When more than one active plugin vendors `wordpress/mcp-adapter` (with or without namespace prefixing), every copy's dispatch runs every copy's subscribers. A subscriber that registers servers without checking *which* adapter dispatched ends up registering on every dispatch, causing duplicate-registration errors.

A consumer that uses the `$adapter` passed to its `mcp_adapter_init` callback is already safe. `on_init()` simply makes that guarantee explicit and harder to get wrong:

```php
\WP\MCP\Core\McpAdapter::instance()->on_init( function ( $adapter ) {
    $adapter->create_server( /* ... */ );
} );
```

The wrapped callback only runs when this adapter is the one dispatching.

## Public API note

This adds a new public method, so it is a commitment once shipped — hence the dedicated PR and docs. Feedback on the name/shape welcome.

## Tests & docs

- Two unit tests: callback runs for the matching adapter, and is skipped when a foreign adapter dispatches the shared action.
- README section documenting the helper alongside the existing custom-server example, plus a note to prefer the passed-in `$adapter` over `McpAdapter::instance()` inside `mcp_adapter_init` callbacks.
- Full unit suite passes locally (963 tests).

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,23 @@ add_action( 'mcp_adapter_init', function( $adapter ) {
 } );
 ```
 
+> [!IMPORTANT]
+> Always register servers against the `$adapter` instance passed to your callback (as above). Do **not** call `\WP\MCP\Core\McpAdapter::instance()` from inside the callback. The `mcp_adapter_init` action name is global, so if two active plugins each bundle their own copy of this library, every copy's dispatch runs every copy's `mcp_adapter_init` callbacks. Using the passed-in `$adapter` keeps each callback registering against the copy that actually dispatched it; reaching for `instance()` instead resolves to *your* copy's singleton on every dispatch and produces `duplicate_server_id` errors.
+
+### Registering servers with `on_init()`
+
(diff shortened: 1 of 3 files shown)


galatanovidiu:
Closing this because we don't need another public API for init callbacks.

`mcp_adapter_init` already passes the Adapter instance that fired the hook. Our documented example uses that `$adapter` to create the server, which is the correct and safe pattern.

`on_init()` wraps the same behavior, but it would become another public method we need to support. The cross-copy case from #173 depends on namespace-prefixing the Adapter, which is not a supported setup. There is no remaining supported use case that justifies this helper.

PR #218, opened 2026-06-23
fix: guard SchemaTransformer recursion and strip WP-only schema keys (#207)

### LINKED ISSUE #207: create_server() with an explicit tools list aborts the whole server, and ability-to-tool conversion recurses to a stack overflow on object-valued sanitize_callback
## Summary

When an ability's input schema contains an object-valued `sanitize_callback` (a normal WordPress pattern, for example `array( $this, 'method' )`), converting that ability to an MCP tool recurses until the Zend stack limit is reached. Inside `create_server()` with an explicit `tools` list, this single failing ability takes down the whole server: no REST route is registered and the endpoint returns `404`. The default-server discovery path, by contrast, skips a failing ability and keeps the server alive.

The asymmetry plus the unbounded recursion make `mcp-adapter` fragile against the very abilities shipped by the official `WordPress/ai` plugin.

## Environment

| | |
|---|---|
| `wordpress/mcp-adapter` | 0.5.0 |
| `wordpress/php-mcp-schema` | 0.1.x |
| WordPress | 7.0 (Abilities API in core) |
| PHP | 8.3 |
| Trigger plugin | `WordPress/ai` 1.0.1 (ability `ai/alt-text-generation`) |

## Steps to reproduce

**Minimal, self-contained** (no `WordPress/ai` needed):

```php
add_action( 'wp_abilities_api_init', function () {
    wp_register_ability( 'demo/object-sanitizer', array(
        'label'               => 'Object sanitizer demo',
        'description'         => 'Reproduces stack overflow on an object-valued sanitize_callback.',
        'input_schema'        => array(
            'type'       => 'object',
            'properties' => array(
                'image_url' => array(
                    'type'              => 'string',
                    // [0] is an object, exactly like array( $this, 'method' ) in WordPress/ai.
                    'sanitize_callback' => array( new \stdClass(), 'noop' ),
                ),
            ),
        ),
        'execute_callback'    => '__return_empty_string',
        'permission_callback' => '__return_true',
    ) );
} );

// Convert it:
$result = \WP\MCP\Domain\Tools\McpTool::fromAbility( wp_get_ability( 'demo/object-sanitizer' ) );
```

**Real-world:** enable the `WordPress/ai` experiments and expose `ai/alt-text-generation` as a tool on a server created with an explicit `tools` list. The server fails to register and `POST /wp-json/<namespace>/<route>` returns `404`.

## Expected behavior

Converting an ability that carries WordPress-only schema keys (`sanitize_callback`, `validate_callback`, `arg_options`) should produce a clean MCP tool schema by ignoring those keys. A single unconvertible ability should never crash the process, and within `create_server()` it should not take down the entire server.

## Actual behavior

Calling the converter per ability (wrapped in `try/catch`) shows:

```
ai/get-post-terms       => WP_Error[mcp_tool_dto_creation_failed]: Expected array or object for key "type", got string
ai/alt-text-generation  => Error: Maximum call stack size of 8339456 bytes (zend.max_allowed_stack_size - zend.reserved_stack_size) reached. Infinite recursion?
ai/title-generation     => OK (tool built)
```

Inside `create_server()` the `alt-text-generation` recursion is not caught, so the call returns a `WP_Error` and the server's REST route is never registered (endpoint returns `404`).

## Root cause

`ai/alt-text-generation` defines (in `WordPress/ai`, `includes/Abilities/Image/Alt_Text_Generation.php`):

```php
'image_url' => array(
    'type'              => 'string',
    'sanitize_callback' => array( $this, 'sanitize_image_reference_input' ), // [0] is the ability object
    'description'       => '...',
),
```

`sanitize_callback` is a standard WordPress schema key whose value here is a PHP callable in `array( $object, $method )` form. `RegisterAbilityAsMcpTool::build()` forwards the raw schema into the tool / `php-mcp-schema` layer, which walks every node. When it descends into the `sanitize_callback` array it enters the bound object (`$this`), whose object graph references back into WordPress registries, and the walk recurses until the stack limit is hit. Properties with string callables (`'absint'`, `'sanitize_textarea_field'`) serialize fine; only the object-valued callable triggers the recursion.

This is a robustness gap in `mcp-adapter` rather than misuse: `sanitize_callback` and `validate_callback` are legitimate on WordPress schemas, so the converter must strip non-JSON-Schema keys and must never recurse into arbitrary object graphs.

## Proposed fix

1. **Strip WordPress-only keys** (`sanitize_callback`, `validate_callback`, `arg_options`, and any callable-valued node) before building the MCP tool schema.
2. **Add a depth and cycle guard** to the schema walker, so a stray object or circular reference degrades to a `WP_Error` instead of fatal recursion.
3. **Fail soft in `create_server()` with an explicit `tools` list:** skip an ability whose conversion fails (logging a `WP_Error` that names the ability) instead of aborting the whole server, matching the discovery path. At minimum, name the offending ability in the returned error.

## Additional finding (separate, likely `WordPress/ai`)

`ai/get-post-terms` fails conversion with `Expected array or object for key "type", got string`. Its `output_schema` is structurally malformed: `properties` is set to `array( 'type' => 'array', 'items' => array( ... ) )`, so `properties.type` is the string `"array"` where a sub-schema is expected. `mcp-adapter` rejects this correctly, so the real fix belongs in `WordPress/ai`. Filed separately there.

## Willing to contribute

Glad to open a PR for fixes 1 to 3 (key-stripping, a guarded walker, and fail-soft server creation) with a regression test built on the minimal repro above.

i-anubhav-anand:
## What?
Closes #207 (recursion guard + key-stripping; create_server fail-soft split to a follow-up)

This PR hardens `SchemaTransformer` against a stack-overflow fatal triggered by object-valued schema callbacks, and stops WordPress-only argument keys from leaking into the emitted MCP schema.

## Why?
Issue #207 reports three problems. Two of them share a single root cause in `SchemaTransformer::convert_objects_to_arrays()`:

1. **Stack overflow (uncatchable fatal).** The method recursed over the schema graph, casting every object to an array and `array_map`-ing over it with no depth or cycle guard. When a schema argument carries an object-valued callable — for example an object-valued `sanitize_callback` whose object references itself — the cast preserves the cyclic graph and the recursion blows the PHP stack. The fatal is uncatchable, so it escapes the per-ability handling and takes down the whole server response. A real-world trigger is the WordPress/ai ability `ai/alt-text-generation`, which uses an object-valued `sanitize_callback`.
2. **WordPress-only keys leaking.** The keys `sanitize_callback`, `validate_callback`, and `arg_options` are WordPress argument conventions, not JSON Schema, yet they passed straight through into the emitted MCP schema.

The third problem reported in #207 — `create_server()` aborting the whole server on a single bad ability — is intentionally **not** in this PR (see Changelog/Testing notes). It is a narrower design decision and is split to a design-gated follow-up; once this fix lands, an uncatchable stack-overflow fatal no longer escapes the existing per-ability handling.

## How?
The unguarded recursive `array_map` in `convert_objects_to_arrays()` is replaced with a guarded `foreach` walk that:

- **Caps recursion** at `MAX_DEPTH = 64` to stop any pathological nesting.
- **Short-circuits cycles** by tracking visited objects via `spl_object_id()`, captured **before** the `(array)` cast — the cast discards object identity, so the id must be read first. A revisited object on the current branch returns `null`.
- **Strips the WordPress-only keys** (`sanitize_callback`, `validate_callback`, `arg_options`) during the same walk, so they never reach the output and any cyclic callable they carry is dropped before it can be traversed.

The public signature and return shape of `transform_to_object_schema()` are unchanged; the new `$depth` and `$visited` parameters are internal-only defaults on the private helper. The code is PHP 7.4 compatible.

### Use of AI Tools
AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 4.8
Used for: multi-agent investigation of the issue, test authoring, implementation, and review. All output was reviewed, edited, and is owned by the submitter.

## Testing Instructions
1. Install dev dependencies: `composer install`.
2. Run the unit suite: `npm run test:php` (or `composer test`). The suite passes — 23 tests / 116 assertions.
3. Run static analysis and linting: `npm run lint:php:stan` and `npm run lint:php` — both clean.
4. The four new tests in `tests/phpunit/Unit/Domain/Utils/SchemaTransformerTest.php` specifically cover:
   - a cyclic object-valued callback no longer overflows the stack;
   - a cycle placed under a non-stripped key pins the depth/cycle guard directly;
   - the WordPress-only keys are stripped from the output;
   - a legitimate nested object schema still converts correctly.
5. Optional end-to-end check: register an ability with an object-valued `sanitize_callback` (e.g. via WordPress/ai `ai/alt-text-generation`) and confirm the MCP server returns its tool schema instead of fataling.

## Changelog Entry

> Fixed - Guard `SchemaTransformer` against stack overflow from cyclic object-valued schema callbacks, and strip the WordPress-only keys (`sanitize_callback`, `validate_callback`, `arg_options`) from the emitted MCP schema.

### DIFF
--- a/includes/Domain/Utils/SchemaTransformer.php
+++ b/includes/Domain/Utils/SchemaTransformer.php
@@ -19,6 +19,31 @@ namespace WP\MCP\Domain\Utils;
  */
 class SchemaTransformer {
 	/**
+	 * Maximum recursion depth for the normalize walk.
+	 *
+	 * Guards against pathological/cyclic structures blowing the PHP stack.
+	 *
+	 * @var int
+	 */
+	private const MAX_DEPTH = 64;
+
+	/**
+	 * WordPress-only argument keys that are not part of the JSON Schema spec.
+	 *
+	 * These must be stripped from the emitted MCP schema. They can carry
+	 * object-valued callables (e.g. sanitize_callback => array( $obj, 'method' ))
+	 * whose object graph may be cyclic.
+	 *
+	 * @var array<int,string>
+	 */
+	// phpcs:ignore SlevomatCodingStandard.Classes.DisallowMultiConstantDefinition -- False positive: sniff mistakes array() commas for multi-const commas (only handles short syntax).
+	private const WP_ONLY_KEYS = array(
+		'sanitize_callback',
+		'validate_callback',
+		'arg_options',
+	);
+
+	/**
 	 * Transform a schema to MCP-compatible object format.
 	 *
 	 * If the schema is already an object type, it is returned unchanged.
@@ -120,17 +145,47 @@ class SchemaTransformer {
 	/**
 	 * Recursively convert objects to arrays.
 	 *
-	 * @param mixed $value The value to convert.
+	 * Casts objects to arrays as before, but guards against runaway recursion:
+	 * cyclic object graphs (e.g. an object-valued sanitize_callback whose object
+	 * references itself) are short-circuited via spl_object_id() tracking, and a
+	 * MAX_DEPTH ceiling stops any other pathological nesting. While walking, the
+	 * WordPress-only keys (sanitize_callback, validate_callback, arg_options) are
+	 * stripped because they are not valid JSON Schema and may carry such cyclic
+	 * callables. They are stripped wherever encountered: that is the behaviour the
+	 * issue asks for and the simplest correct approach. The risk of clobbering a
+	 * legitimately named data property is accepted as negligible for schema input.
+	 *
+	 * @param mixed              $value   The value to convert.
+	 * @param int                $depth   Current recursion depth.
+	 * @param array<int,bool>    $visited Object ids already seen on this branch.
 	 *
 	 * @return mixed The converted value.
 	 */
-	private static function convert_objects_to_arrays( $value ) {
+	private static function convert_objects_to_arrays( $value, int $depth = 0, array $visited = array() ) {
+		if ( $depth >= self::MAX_DEPTH ) {
+			return null;
+		}
+
 		if ( is_object( $value ) ) {
-			$value = (array) $value;
+			// Track identity BEFORE the cast, since (array) loses object identity.
+			$object_id = spl_object_id( $value );
+			if ( isset( $visited[ $object_id ] ) ) {
+				return null;
+			}
+			$visited[ $object_id ] = true;
+			$value                 = (array) $value;
 		}
 
 		if ( is_array( $value ) ) {
-			return array_map( array( self::class, 'convert_objects_to_arrays' ), $value );
+			$result = array();
+			foreach ( $value as $key => $item ) {
+				if ( in_array( $key, self::WP_ONLY_KEYS, true ) ) {
+					continue;
+				}
+				$result[ $key ] = self::convert_objects_to_arrays( $item, $depth + 1, $visited );
+			}
+
+			return $result;
 		}
 
 		return $value;
--- a/tests/phpunit/Unit/Domain/Utils/SchemaTransformerTest.php
+++ b/tests/phpunit/Unit/Domain/Utils/SchemaTransformerTest.php
@@ -373,4 +373,164 @@ final class SchemaTransformerTest extends TestCase {
 		$this->assertIsArray( $result['schema']['properties']['name'] );
 		$this->assertSame( 'string', $result['schema']['properties']['name']['type'] );
 	}
+
+	/**
+	 * Issue #207: A schema property may carry an object-valued WP callback
+	 * (e.g. sanitize_callback => array( $obj, 'method' )). If that object's
+	 * graph contains a back-reference (cycle), convert_objects_to_arrays()
+	 * recurses without a cycle guard until the PHP stack/memory blows.
+	 *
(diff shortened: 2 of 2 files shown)


galatanovidiu:
This PR will not be merged. I am closing it.

The failure comes from an invalid Ability schema. Ability input and output schemas must be JSON Schema. `sanitize_callback` is REST argument metadata and is not part of that contract. In WordPress/ai 1.0.1, the callback contains `$this`, which creates a cycle back to the Ability. The empty `stdClass` example in #207 does not reproduce the failure.

WordPress/ai fixed the invalid producer in [WordPress/ai#688](https://github.com/WordPress/ai/pull/688), released in 1.0.2. No valid Ability schema has been shown to trigger this problem.

For an explicitly configured server, `create_server()` returning a `WP_Error` when a selected tool is invalid is correct. The MCP Adapter does not promise to silently skip invalid tools from an explicit tools list. The default discovery path does not perform this conversion.

This PR adds special handling for malformed producer data. It removes matching keys everywhere and replaces cycles or depth limits with `null`. That silently changes schemas without solving a demonstrated MCP Adapter problem.

A new adapter change requires either a reproduction using valid Ability JSON Schema or a documented requirement that invalid explicitly selected tools must be skipped.

PR #222, opened 2026-06-24
Move HttpTransport hook registration to boot()

### LINKED ISSUE #52: Follow-up #48: Can we move this to an `initialize()` method or somewhere else? I think it's better to not have hooks as a side-effect of constructors.
Can we move this to an `initialize()` method or somewhere else? I think it's better to not have hooks as a side-effect of constructors.

_Originally posted by @JasonTheAdams in https://github.com/WordPress/mcp-adapter/pull/48#discussion_r2445425031_

mathetos:
## Summary

Follow-up to #52 (review comment on #48): HttpTransport no longer registers 
est_api_init in its constructor.

- Add `boot()` to `McpTransportInterface` for WordPress hook registration after construction
- `McpTransportFactory` calls `boot()` immediately after instantiating each transport
- `HttpTransport::boot()` keeps the existing `rest_api_init` priority (16) behavior
- Update custom transport docs and add a small unit test for the boot lifecycle

## Motivation

Constructors should set up dependencies only; hook registration is explicit and easier to test in isolation. Happy to adjust naming or scope if maintainers prefer a lighter-touch approach (e.g. factory-only hook without an interface change).

Closes #52

## Test plan

- [ ] CI unit/integration suite passes
- [ ] `HttpTransportBootTest` confirms the constructor does not register hooks and `boot()` registers `rest_api_init` at priority 16
- [ ] Smoke: MCP REST route still resolves after plugin activation (`rest_url( 'mcp/mcp-adapter-default-server' )`)

### DIFF
--- a/docs/guides/custom-transports.md
+++ b/docs/guides/custom-transports.md
@@ -27,6 +27,7 @@ Custom transports implement one of two interfaces:
 ```php
 interface McpTransportInterface {
     public function __construct( McpTransportContext $context );
+    public function boot(): void;
     public function register_routes(): void;
 }
 ```
@@ -68,6 +69,9 @@ class ApiKeyTransport implements McpRestTransportInterface {
     
     public function __construct( McpTransportContext $context ) {
         $this->context = $context;
+    }
+
+    public function boot(): void {
         $this->register_routes();
     }
     
@@ -158,7 +162,8 @@ See [Transport Permissions](transport-permissions.md) for simpler authentication
 ## Implementation Notes
 
 ### Required Methods
-- `__construct()`: Accept `McpTransportContext` and call `register_routes()`
+- `__construct()`: Accept `McpTransportContext` and set up dependencies (no WordPress hooks)
+- `boot()`: Register WordPress hooks; `McpTransportFactory` calls this after construction
 - `register_routes()`: Register WordPress REST API endpoints
 - `check_permission()`: Validate request access (REST transports only)
 - `handle_request()`: Process MCP requests (REST transports only)
--- a/includes/Core/McpTransportFactory.php
+++ b/includes/Core/McpTransportFactory.php
@@ -83,8 +83,9 @@ class McpTransportFactory {
 			}
 
 			// Interface-based instantiation with dependency injection
-			$context = $this->create_transport_context();
-			new $mcp_transport( $context );
+			$context   = $this->create_transport_context();
+			$transport = new $mcp_transport( $context );
+			$transport->boot();
 		}
 	}
 
--- a/includes/Transport/Contracts/McpTransportInterface.php
+++ b/includes/Transport/Contracts/McpTransportInterface.php
@@ -29,6 +29,14 @@ interface McpTransportInterface {
 	public function __construct( McpTransportContext $context );
 
 	/**
+	 * Boot the transport after construction.
+	 *
+	 * Register WordPress hooks here rather than in the constructor.
+	 * Called by McpTransportFactory immediately after instantiation.
+	 */
+	public function boot(): void;
+
+	/**
 	 * Register transport-specific routes.
 	 *
 	 * Called during WordPress REST API initialization to register
--- a/includes/Transport/HttpTransport.php
+++ b/includes/Transport/HttpTransport.php
@@ -41,12 +41,20 @@ class HttpTransport implements McpRestTransportInterface {
 	protected HttpRequestHandler $request_handler;
 
 	/**
-	 * Initialize the class and register routes
+	 * Initialize the transport with the provided context.
 	 *
 	 * @param \WP\MCP\Transport\Infrastructure\McpTransportContext $transport_context The transport context.
 	 */
 	public function __construct( McpTransportContext $transport_context ) {
 		$this->request_handler = new HttpRequestHandler( $transport_context );
+	}
+
+	/**
+	 * Boot the transport and register WordPress hooks.
+	 *
+	 * Called by McpTransportFactory after construction so constructors stay free of hook side effects.
+	 */
+	public function boot(): void {
 		add_action( 'rest_api_init', array( $this, 'register_routes' ), 16 );
 	}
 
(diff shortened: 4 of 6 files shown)


tests/phpunit/Unit/Transport/HttpTransportBootTest.php
```diff
@@ -0,0 +1,68 @@
...
+
+	public function test_boot_registers_rest_api_init_hook_at_expected_priority(): void {
+		$transport = new HttpTransport( $this->createTransportContext() );
+
+		$transport->boot();
+
+		$this->assertSame( 16, has_action( 'rest_api_init', array( $transport, 'register_routes' ) ) );
+	}
+
+	private function createTransportContext(): McpTransportContext {
```
abhi3315, line 43: These could go in `tests/phpunit/Integration/HttpTransportTest.php` instead. Its `set_up()` already builds the server, context and an `HttpTransport`, so you'd just assert against `$this->transport` and drop this helper. There are already five other copies of `createTransportContext()` in the suite.

Assertions look fine. This is only about placement.

includes/Transport/HttpTransport.php
```diff
@@ -41,12 +41,20 @@ class HttpTransport implements McpRestTransportInterface {
...
 	public function __construct( McpTransportContext $transport_context ) {
 		$this->request_handler = new HttpRequestHandler( $transport_context );
+	}
+
+	/**
+	 * Boot the transport and register WordPress hooks.
+	 *
+	 * Called by McpTransportFactory after construction so constructors stay free of hook side effects.
+	 */
+	public function boot(): void {
```
abhi3315, line 57: "so constructors stay free of hook side effects" is only true of this one. `new McpServer( ..., array( HttpTransport::class ), ... )` still registers the hook before the constructor returns, through `setup_components()` -> `initialize_transports()` -> `boot()`.

#52 is about this constructor, so the PR does close it. The plural here just overstates it. Maybe narrow it to `HttpTransport::__construct()`.

Also worth adding `@since n.e.x.t` here and on the interface method.

docs/guides/custom-transports.md
```diff
@@ -68,6 +69,9 @@ class ApiKeyTransport implements McpRestTransportInterface {
     
     public function __construct( McpTransportContext $context ) {
         $this->context = $context;
+    }
+
+    public function boot(): void {
```
abhi3315, line 74: This calls `register_routes()` directly, but `HttpTransport::boot()` adds a hook. That matters under WP-CLI.

`McpAdapter.php:62-66` hooks `init` at 20 under WP-CLI and `rest_api_init` at 15 otherwise. So on the CLI path this runs before `rest_api_init` has fired, and `register_rest_route()` hits core's `! did_action( 'rest_api_init' )` check and calls `_doing_it_wrong()`.

Same shape as `HttpTransport` works. 16 because the adapter itself runs at 15:

```php
public function boot(): void {
    add_action( 'rest_api_init', array( $this, 'register_routes' ), 16 );
}
```

Not new in this PR, the constructor on trunk did the same thing. But line 166 now says `boot()` registers hooks, so the example doesn't match it.

includes/Core/McpTransportFactory.php
```diff
@@ -83,8 +83,9 @@ public function initialize_transports( array $mcp_transports ): void {
 			}
 
 			// Interface-based instantiation with dependency injection
-			$context = $this->create_transport_context();
-			new $mcp_transport( $context );
+			$context   = $this->create_transport_context();
+			$transport = new $mcp_transport( $context );
+			$transport->boot();
```
abhi3315, line 88: If you go with the separate interface (see my comment on `McpTransportInterface.php`), this becomes:

```php
if ( $transport instanceof BootableTransportInterface ) {
    $transport->boot();
}
```

includes/Transport/Contracts/McpTransportInterface.php
```diff
@@ -28,6 +28,14 @@ interface McpTransportInterface {
 	 */
 	public function __construct( McpTransportContext $context );
 
+	/**
+	 * Boot the transport after construction.
+	 *
+	 * Register WordPress hooks here rather than in the constructor.
+	 * Called by McpTransportFactory immediately after instantiation.
+	 */
+	public function boot(): void;
```
abhi3315, line 37: I think this one blocks. Adding `boot()` here breaks any existing custom transport that doesn't already have a public `boot(): void`. PHP throws a fatal when the class is declared and you can't catch it. Under autoload that happens on the `class_exists()` at line 48, before the interface check at :67.

There is a real implementation outside this repo. [wp-media/mcp-oauth](https://github.com/wp-media/mcp-oauth/blob/v1.1/inc/Transport/OAuthHttpTransport.php) v1.1 has `OAuthHttpTransport implements McpRestTransportInterface` with no `boot()`. Its `^0.5` constraint means Composer won't pull it in on 0.6.x, so it isn't broken today, but it shows the interface does get implemented downstream. #159 is the closer precedent, that reverted a property visibility change because it broke a downstream transport.

A separate interface avoids touching the existing contract:

```php
interface BootableTransportInterface {
    public function boot(): void;
}
```

`HttpTransport` implements both, and the factory checks `instanceof` before calling `boot()`.

includes/Core/McpTransportFactory.php
abhi3315, line 88: No test asserts that the factory calls `boot()`. The new tests call `HttpTransport::boot()` directly. I removed this line locally and the suite still passed, `OK (1004 tests, 3876 assertions)`. Codecov marks it covered because it runs, but nothing checks it.

A counter on `DummyTransport::boot()` would test it. Needs to be static since the factory never returns the instance, and reset per test since the fixture is shared:

```php
DummyTransport::$boot_count = 0;
$this->transport_factory->initialize_transports( array( DummyTransport::class ) );
$this->assertSame( 1, DummyTransport::$boot_count );
```

If `boot()` moves to its own interface, worth checking the other case too, that a transport with only the old interface still initializes.

abhi3315 (commented):
Thanks for working on this. I found one issue I think blocks merging: adding boot() to the existing public interface breaks current custom transports. The other comments are smaller test and documentation fixes.

One documentation gap isn’t on a changed line: docs/architecture/overview.md still shows the old interface and a custom transport without the new lifecycle. It should match whichever contract is chosen. The branch also predates 0.6.1, so it will need an update from trunk and fresh CI before merge.

PR #231, opened 2026-07-03
feat: map open_world ability annotation to openWorldHint

jorgefilipecosta:
## What?

Follow up to WordPress/ai#739, which introduces the WordPress-format `open_world` annotation on the `core/read-content` ability.

Adds `open_world` to the WordPress → MCP annotation conversion in `McpAnnotationMapper`, so abilities can declare the MCP `openWorldHint` tool annotation using the WordPress Abilities API naming convention, like the existing `readonly`, `destructive`, and `idempotent` keys.

## Why?

The adapter already converts WordPress-format annotation keys to MCP tool annotations (`readonly` → `readOnlyHint`, `destructive` → `destructiveHint`, `idempotent` → `idempotentHint`), but `openWorldHint` had no WordPress-format equivalent (`ability_property => null`), so a WordPress-format `'open_world' => false` annotation was silently dropped.

Per the MCP specification, clients assume `openWorldHint` is `true` when the hint is absent. That means closed-world abilities that declare `'open_world' => false` — such as `core/read-content` in WordPress/ai#739, which only reads the local database — were presented to MCP clients as tools that may interact with external systems.

## How?

- Sets the `ability_property` for `openWorldHint` to `open_world` in `McpAnnotationMapper::$mcp_annotations`. The existing resolution logic handles the rest: the WordPress-format key takes precedence when both are present, and the direct MCP-format `openWorldHint` key keeps working unchanged.
- Adds unit tests for the new mapping and its precedence over the MCP-format key, and updates the null-`ability_property` test (for tools, only `title` remains without a WordPress equivalent).
- Documents the new mapping in `docs/guides/creating-abilities.md`.

## Testing Instructions

1. Register an ability with `'meta' => array( 'annotations' => array( 'open_world' => false ) )` and expose it as a tool on an MCP server.
2. Call `tools/list` on the server and confirm the tool's annotations include `"openWorldHint": false` (before this change the hint was missing).

Or run the mapper unit tests:

```
npm run test:php -- --filter McpAnnotationMapperTest
```

## Changelog Entry

> Added - Support the WordPress-format `open_world` ability annotation, mapped to the MCP `openWorldHint` tool annotation.

### DIFF
--- a/docs/guides/creating-abilities.md
+++ b/docs/guides/creating-abilities.md
@@ -312,7 +312,7 @@ The MCP Adapter automatically converts WordPress Abilities API annotation names 
         'readonly' => true,        // Auto-converted to readOnlyHint
         'destructive' => false,    // Auto-converted to destructiveHint
         'idempotent' => true,      // Auto-converted to idempotentHint
-        'openWorldHint' => false,  // No WordPress equivalent, use MCP format
+        'open_world' => false,     // Auto-converted to openWorldHint
         'title' => 'My Tool'       // No WordPress equivalent, use MCP format
     ]
 ]
@@ -336,7 +336,7 @@ The MCP Adapter automatically converts WordPress Abilities API annotation names 
 | `readonly` | `readOnlyHint` | Tool doesn't modify data |
 | `destructive` | `destructiveHint` | Tool may delete/destroy data |
 | `idempotent` | `idempotentHint` | Same input → same output |
-| *(no equivalent)* | `openWorldHint` | Can work with arbitrary data |
+| `open_world` | `openWorldHint` | Can work with arbitrary data |
 | *(no equivalent)* | `title` | Custom display title |
 
 **Why Use WordPress Format?**
@@ -386,6 +386,7 @@ Tools support these MCP specification annotations:
 - `readonly` → `readOnlyHint`
 - `destructive` → `destructiveHint`
 - `idempotent` → `idempotentHint`
+- `open_world` → `openWorldHint`
 
 ### Resource & Prompt Annotations (Annotations)
 
@@ -427,7 +428,7 @@ wp_register_ability('my-plugin/analyze-data', [
             'readonly' => true,              // WordPress format → readOnlyHint
             'destructive' => false,          // WordPress format → destructiveHint
             'idempotent' => true,            // WordPress format → idempotentHint
-            'openWorldHint' => false,        // No WordPress equivalent
+            'open_world' => false,           // WordPress format → openWorldHint
             'title' => 'Data Analysis Tool'  // No WordPress equivalent
         ],
         'mcp' => [
--- a/includes/Domain/Utils/McpAnnotationMapper.php
+++ b/includes/Domain/Utils/McpAnnotationMapper.php
@@ -74,7 +74,7 @@ class McpAnnotationMapper {
 		'openWorldHint'   => array(
 			'type'             => 'boolean',
 			'features'         => array( 'tool' ),
-			'ability_property' => null,
+			'ability_property' => 'open_world',
 		),
 		'title'           => array(
 			'type'             => 'string',
--- a/tests/phpunit/Unit/Domain/Utils/McpAnnotationMapperTest.php
+++ b/tests/phpunit/Unit/Domain/Utils/McpAnnotationMapperTest.php
@@ -150,6 +150,30 @@ final class McpAnnotationMapperTest extends TestCase {
 		$this->assertTrue( $result['idempotentHint'] );
 	}
 
+	public function test_map_maps_open_world_to_openworldhint(): void {
+		$annotations = array(
+			'open_world' => false,
+		);
+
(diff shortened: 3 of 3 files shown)


docs/guides/creating-abilities.md
```diff
@@ -336,7 +336,7 @@ The MCP Adapter automatically converts WordPress Abilities API annotation names
 | `readonly` | `readOnlyHint` | Tool doesn't modify data |
 | `destructive` | `destructiveHint` | Tool may delete/destroy data |
 | `idempotent` | `idempotentHint` | Same input → same output |
-| *(no equivalent)* | `openWorldHint` | Can work with arbitrary data |
+| `open_world` | `openWorldHint` | Can work with arbitrary data |
```
galatanovidiu, line 339: question: small naming one. The other three WordPress-format hints are single words (`readonly`, `destructive`, `idempotent`), so by that pattern this key would read `openworld`. I'd lean `open_world` anyway since snake_case is more WordPress, but since it's becoming an Abilities API convention, could we settle the spelling on purpose so the adapter and the Abilities API stay in sync?

tests/phpunit/Unit/Domain/Utils/McpAnnotationMapperTest.php
```diff
@@ -150,6 +150,30 @@ public function test_map_maps_idempotent_to_idempotenthint(): void {
...
+		$annotations = array(
+			'open_world' => false,
+		);
+
+		$result = McpAnnotationMapper::map( $annotations, 'tool' );
+
+		$this->assertArrayHasKey( 'openWorldHint', $result );
+		$this->assertArrayNotHasKey( 'open_world', $result );
+		$this->assertFalse( $result['openWorldHint'] );
+	}
```
galatanovidiu, line 163: `open_world` itself is never exercised on the feature-gating or invalid-boolean paths. Could we add `'open_world' => false` to the resource and prompt exclusion tests with an `assertArrayNotHasKey( 'openWorldHint', $result )`, plus one non-boolean `open_world` case that gets dropped? That locks the new key in against a future regression.

includes/Domain/Utils/McpAnnotationMapper.php
```diff
@@ -74,7 +74,7 @@ class McpAnnotationMapper {
 		'openWorldHint'   => array(
 			'type'             => 'boolean',
 			'features'         => array( 'tool' ),
-			'ability_property' => null,
+			'ability_property' => 'open_world',
```
galatanovidiu, line 77: One edge from replaying old registrations: `['open_world' => 'invalid', 'openWorldHint' => false]` used to emit `openWorldHint => false`, now it emits nothing, and the spec default for a missing hint is true. It comes from the shared resolution path: `resolve_annotation_value()` picks the WP-format value whenever the key exists, and when `normalize_boolean()` rejects it the whole hint is dropped instead of falling back to the still-valid MCP key. The same edge already exists for the other three hints, so nothing to change here. I'll open an issue to add the fallback for all four.

galatanovidiu, line 77: nit: could we add `@since n.e.x.t` for the new mapping in the `$mcp_annotations` or `map()` docblock? The file has no `@since` tags anywhere, so nothing records that `open_world` support arrives in the next release.

galatanovidiu (commented):
Handy addition: `open_world` now works as the WordPress-format alias for `openWorldHint`, like the other three hints. Before this you had to set the MCP-format `openWorldHint` directly (which the docs said to do), and a WordPress-format `open_world` key was silently dropped, so an ability like `core/read-content` in WordPress/ai#739 that wrote `open_world => false` lost its closed-world hint and fell back to the spec default of true. Nothing blocking. Worth a look, left inline:

- naming: by the one-word pattern of the other hints this would read `openworld`; worth settling the spelling on purpose since it's an Abilities API convention.
- no test exercises the tool-only feature gate or the invalid-value branch through the `open_world` key itself.
- one replay edge: an invalid `open_world` next to a valid `openWorldHint` now drops the hint. Shared resolution path (pre-exists for the other three hints), so nothing for this PR; I'll open a separate issue.

PHPCS, PHPStan L8, and the mapper tests (24/24) pass locally. Registrations using only `openWorldHint` replay byte-identical.

<sub>Reviewed by Claude and Codex, checked by me.</sub>

docs/guides/creating-abilities.md
justlevine: > , but since it's becoming an Abilities API convention,

@galatanovidiu what do you mean by this part? "openWorldHint" seems very mcp-specific. IMO even if we bring support for marking Abilities that handle external data, I would assume we choose something that's semantic and self-defining to _WordPress_ and then have MCP Adapter map it to here.

---

Regardless of the Abilities part of this, +1 to `open_world`, and consider `readonly` an exception-to-the-rule, instead of establishing a future pattern of stripping multi-word semantics when going from `camelCase`.

justlevine:
(rest of this lgtm once the open comments are handled)

Issue #229, opened 2026-07-02
Empty {} / omitted parameters still fail in v0.5.0 for abilities WITH input schemas (follow-up to #116)


pkhaninejad:
## Description

#116 was closed as fixed in v0.5.0 via `AbilityArgumentNormalizer`, but the fix only covers abilities **without** an input schema. Abilities **with** an input schema (the exact case in #116's reproduction steps) still fail on v0.5.0 when called with empty `{}` parameters:

```
Ability "wds/get-posts" has invalid input. Reason: input is not of type object.
```

We reproduced this today in production on **mcp-adapter v0.5.0** across two independent WordPress sites, for every registered ability that defines an `input_schema` (type `object`, all-optional or required properties alike).

## Steps to Reproduce

1. Register an ability with an `input_schema` of type `object` (optional properties, no `default` key).
2. Call `mcp-adapter-execute-ability` with:

```json
{
    "ability_name": "example/get-info",
    "parameters": {}
}
```

3. **Expected:** ability executes (no required properties).
4. **Actual:** `Ability "example/get-info" has invalid input. Reason: input is not of type object.`

Any **non-empty** parameters object (e.g. `{"anything": true}`) succeeds.

## Evidence that the ability receives `null`, not `[]`

`AbilityArgumentNormalizer::normalize()` only converts `[]` → `null` when the ability has **no** schema, and never converts `null` → `[]` when it **has** one. On v0.5.0 the inner ability demonstrably receives `null`:

We patched our plugin to add `'default' => array()` to every ability's `input_schema` — and the failure disappeared on the same v0.5.0 install, with no other change. `WP_Ability::normalize_input()` only applies the schema `default` **when input is `null`**, and `rest_validate_value_from_schema( array(), … )` passes — so the pre-patch failure must be `rest_validate_value_from_schema( null, {type: object} )` → "input is not of type object".

So somewhere in the v0.5.0 execution chain, `"parameters": {}` (or an omitted `parameters` key — `$input['parameters'] ?? null` in `ExecuteAbilityAbility::execute()`) reaches `WP_Ability::execute()` as `null` for schema-defining abilities, and validation rejects it with the misleading type error.

## Suggested fix

In `AbilityArgumentNormalizer::normalize()` (or `ExecuteAbilityAbility::execute()` / the tools handler), handle the inverse case as well: when the ability **has** an input schema and the arguments are `null` or an empty array, pass `array()` so it validates as an empty object. Roughly:

```php
[...]

PR #233, opened 2026-07-06
fix: use Jetpack Autoloader to resolve WP\MCP\ version conflicts

galatanovidiu:
## What

The standalone plugin now loads its classes through the [Jetpack Autoloader](https://github.com/Automattic/jetpack-autoloader) instead of the plain Composer autoloader.

- Add `automattic/jetpack-autoloader` to `require` (and allow its Composer plugin).
- Bootstrap prefers `vendor/autoload_packages.php`, falling back to the classic `vendor/autoload.php` if it is missing.
- Set `COMPOSER_ROOT_VERSION` in the release build so the generated class manifest records the real plugin version, not a guessed `dev-*` one.

## Why

WooCommerce (and other plugins) bundle their own copy of mcp-adapter and autoload the same `WP\MCP\` namespace. When both the standalone plugin and a bundled copy are active, PHP loads whichever registered its autoloader **first** — not the newest version.

A maintainer traced exactly this on production WooCommerce sites in #229: WooCommerce 10.9.1 vendors mcp-adapter `^0.3.0`, and its old `ExecuteAbilityAbility` won the class race, so fixes shipped in the standalone plugin never ran. New classes (only present in the newer copy) loaded from the standalone plugin — a mixed-version load.

The Jetpack Autoloader fixes this: every active copy registers into a shared manifest, and the newest version of each class is loaded regardless of plugin load order. This is the same approach WooCommerce and Jetpack already use, and it is the pattern this project's own README recommends to downstream consumers.

### Why `COMPOSER_ROOT_VERSION`

"Newest wins" needs a real version to compare. Composer knows dependency versions, but for the root package (this plugin) it guesses a `dev-*` version from the git ref unless told otherwise. `dev-*` is treated as "always newest", which would wrongly make the standalone copy win even against a genuinely newer bundled copy. Passing the release tag as `COMPOSER_ROOT_VERSION` bakes the real semver (e.g. `0.5.0`) into the manifest at build time.

## Scope

This is the structural root-cause fix for #229. It is separate from #230, which is the narrower empty-input normalization fix. Note: this only helps once shipped copies carry the autoloader — it cannot retroactively change WooCommerce's already-vendored 0.3.x. Full resolution on today's WooCommerce sites still needs WooCommerce to bump its bundled dependency.

## Testing

- `composer validate`: clean.
- PHPCS: clean on the changed bootstrap.
- Confirmed `composer install` generates `vendor/autoload_packages.php` and a class manifest with correct per-package versions.
- Confirmed `COMPOSER_ROOT_VERSION=0.5.0` makes the manifest record `0.5.0` for the root `WP\MCP\` classes instead of `dev-trunk`.
- Runtime smoke test (WordPress functions shimmed): both a plugin class (`WP\MCP\...\AbilityArgumentNormalizer`) and a dependency class (`WP\McpSchema\...\DTO\Tool`) resolve through the Jetpack Autoloader.
- Unit test bootstrap is unaffected — it uses the classic `vendor/autoload.php`, which is still generated alongside.
- Full PHPUnit suite not run locally (wp-env has a pre-existing config error); please rely on CI.

Refs #229.

### DIFF
--- a/includes/Autoloader.php
+++ b/includes/Autoloader.php
@@ -40,10 +40,7 @@ final class Autoloader {
 			return self::$is_loaded;
 		}
 
-		// Load through the Jetpack Autoloader. When more than one active plugin
-		// ships the WP\MCP\ classes (for example WooCommerce vendors its own
-		// copy), it loads the newest version of each class instead of whichever
-		// plugin registered its autoloader first.
+		// Jetpack Autoloader uses `autoload_packages.php` instead of `autoload.php`.
 		$autoloader      = WP_MCP_DIR . '/vendor/autoload_packages.php';
 		self::$is_loaded = self::require_autoloader( $autoloader );
 
--- a/includes/Autoloader.php
+++ b/includes/Autoloader.php
@@ -40,19 +40,11 @@ final class Autoloader {
 			return self::$is_loaded;
 		}
 
-		// Prefer the Jetpack Autoloader. When more than one active plugin ships
-		// the WP\MCP\ classes (for example WooCommerce vendors its own copy),
-		// it loads the newest version of each class instead of whichever plugin
-		// registered its autoloader first.
-		$jetpack_autoloader = WP_MCP_DIR . '/vendor/autoload_packages.php';
-		if ( is_readable( $jetpack_autoloader ) ) {
-			self::$is_loaded = self::require_autoloader( $jetpack_autoloader );
-
-			return self::$is_loaded;
-		}
-
-		// Fallback to the classic Composer autoloader.
-		$autoloader      = WP_MCP_DIR . '/vendor/autoload.php';
+		// Load through the Jetpack Autoloader. When more than one active plugin
+		// ships the WP\MCP\ classes (for example WooCommerce vendors its own
+		// copy), it loads the newest version of each class instead of whichever
+		// plugin registered its autoloader first.
+		$autoloader      = WP_MCP_DIR . '/vendor/autoload_packages.php';
 		self::$is_loaded = self::require_autoloader( $autoloader );
 
 		return self::$is_loaded;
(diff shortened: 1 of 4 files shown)


includes/Autoloader.php
```diff
@@ -40,6 +40,18 @@ public static function autoload(): bool {
...
+		// it loads the newest version of each class instead of whichever plugin
+		// registered its autoloader first.
+		$jetpack_autoloader = WP_MCP_DIR . '/vendor/autoload_packages.php';
+		if ( is_readable( $jetpack_autoloader ) ) {
+			self::$is_loaded = self::require_autoloader( $jetpack_autoloader );
+
+			return self::$is_loaded;
+		}
+
+		// Fallback to the classic Composer autoloader.
```
justlevine: What's the scenario where jetpack autoloader doesn't exist but the normal composer autoloader does?

galatanovidiu: When using `composer install --no-plugins`
Release zips always have both files.
It was meant as a cheap safety net, but now that you ask, I think we should drop it entirely. We want to always force loading through the Jetpack autoloader.

galatanovidiu: Dropt it on [42adfbe](https://github.com/WordPress/mcp-adapter/pull/233/commits/42adfbe0871a77c6cd5ecee96b34d708b6e7d1b1)

includes/Autoloader.php
```diff
@@ -40,7 +40,11 @@ public static function autoload(): bool {
 			return self::$is_loaded;
 		}
 
-		$autoloader      = WP_MCP_DIR . '/vendor/autoload.php';
+		// Load through the Jetpack Autoloader. When more than one active plugin
+		// ships the WP\MCP\ classes (for example WooCommerce vendors its own
+		// copy), it loads the newest version of each class instead of whichever
+		// plugin registered its autoloader first.
```
justlevine:
```suggestion
		// Jetpack Autoloader uses `autoload_packages.php` instead of `autoload.php`.
```

justlevine (approved):
I replaced the ai-generated comment with one of value. lgtm 🚀

PR #238, opened 2026-07-10
Fix static trait property access deprecation in RequestRouter

### LINKED ISSUE #182: PHP deprecated warnings: Calling static trait method / Accessing static trait property
PHP 8.1 Deprecated:  Calling static trait method WP\MCP\Infrastructure\Observability\McpObservabilityHelperTrait::is_sensitive_key is deprecated, it should only be called on a class using the trait in [path]/mcp-adapter/includes/Transport/Infrastructure/RequestRouter.php on line 292
PHP 8.1 Deprecated:  Accessing static trait property WP\MCP\Infrastructure\Observability\McpObservabilityHelperTrait::$sensitive_patterns is deprecated, it should only be accessed on a class using the trait in [path]/mcp-adapter/includes/Infrastructure/Observability/McpObservabilityHelperTrait.php on line 180

lhero-org:
## Problem

`RequestRouter::sanitize_params_for_logging()` calls `ErrorLogMcpObservabilityHandler::is_sensitive_key()` as a static call, reaching across into a class that isn't `RequestRouter` itself. `is_sensitive_key()` internally reads `self::$sensitive_patterns`, a private static property declared in `McpObservabilityHelperTrait`. Since `RequestRouter` never brings that trait in, this is effectively borrowing another class's copy of a trait-provided static member, which triggers a PHP 8.1+ deprecation notice. Per the PHP manual, this pattern is slated to become a fatal error in a future major version.

Also reported independently by @ahmed0magdy in the comments with the same root-cause diagnosis (0.5.0, WP 7.0, PHP 8.2).

## Fix

Bring `McpObservabilityHelperTrait` into `RequestRouter` directly and call the method via `self::` instead of via the unrelated class:

- Swap the `ErrorLogMcpObservabilityHandler` import for `McpObservabilityHelperTrait`
- `use McpObservabilityHelperTrait;` inside `RequestRouter`
- `ErrorLogMcpObservabilityHandler::is_sensitive_key(...)` → `self::is_sensitive_key(...)`
- Drop the now-resolved `@todo` above it

This is the smaller of the two fixes discussed on the issue (the other being extraction of `is_sensitive_key`/`$sensitive_patterns` into a separate shared utility class both handlers delegate to — also reasonable, but more invasive for the same outcome).

Note: the trait also defines `categorize_error()`, which `RequestRouter` already implements itself (with a different, non-static signature). PHP resolves this without conflict — a class's own method silently takes precedence over a same-named trait method — so no behavior change there; verified locally.

Fixes #182

### DIFF
--- a/includes/Transport/Infrastructure/RequestRouter.php
+++ b/includes/Transport/Infrastructure/RequestRouter.php
@@ -10,7 +10,7 @@ declare( strict_types=1 );
 namespace WP\MCP\Transport\Infrastructure;
 
 use WP\MCP\Infrastructure\ErrorHandling\McpErrorFactory;
-use WP\MCP\Infrastructure\Observability\ErrorLogMcpObservabilityHandler;
+use WP\MCP\Infrastructure\Observability\McpObservabilityHelperTrait;
 use WP\McpSchema\Common\AbstractDataTransferObject;
 use WP\McpSchema\Common\Content\DTO\TextContent;
 use WP\McpSchema\Common\JsonRpc\DTO\JSONRPCErrorResponse;
@@ -23,6 +23,8 @@ use WP\McpSchema\Server\Tools\DTO\CallToolResult;
  * all transport implementations via dependency injection.
  */
 class RequestRouter {
+
+	use McpObservabilityHelperTrait;
 
 	/**
 	 * The transport context.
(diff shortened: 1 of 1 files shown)


galatanovidiu:
Closing this because #193 (`7cc42a0`) already fixed the deprecation on `trunk`.

In v0.5.0, `RequestRouter` called `McpObservabilityHelperTrait::is_sensitive_key()` directly, which is deprecated in PHP 8.1+. #193 changed this to `ErrorLogMcpObservabilityHandler::is_sensitive_key()`. That class uses the trait, so the call is valid PHP and emits no deprecation.

This PR starts from a base that already includes #193. Importing the entire trait into `RequestRouter` is therefore not needed and would also bring in unrelated helper methods.

The fix has not shipped yet, which explains the reports on #182. We should include the existing `trunk` fix in the next release.

PR #244, opened 2026-07-14
Make the built-in discover/get-info/execute tools' MCP exposure check filterable (#243)

### LINKED ISSUE #243: Make the built-in discover/get-info/execute tools' MCP exposure check filterable
## Problem

The three built-in abilities that ship with `mcp-adapter` — `mcp-adapter/discover-abilities`, `mcp-adapter/get-ability-info`, and `mcp-adapter/execute-ability` — decide what a client is allowed to see or invoke by reading `meta.mcp.public` directly on each ability. That check is hard-coded, so the answer to "is this ability exposed via MCP?" is a single global boolean baked into the ability itself.

For simple installs this works fine. It breaks down as soon as an integrator wants exposure to depend on *context* rather than a static flag on the ability — most importantly, when the same ability should be visible on one MCP server but not another.

## Where the decision lives

Both helpers in `includes/Abilities/McpAbilityHelperTrait.php` read `$meta['mcp']['public']` directly:

- `is_ability_mcp_public()` — used by `DiscoverAbilitiesAbility::execute()` to decide which abilities to list.
- `check_ability_mcp_exposure()` — used by `GetAbilityInfoAbility::check_permission()` and `ExecuteAbilityAbility::check_permission()` to decide whether the call is allowed.

Neither read is wrapped in `apply_filters()`. The only extensibility the three abilities expose today is the `*_capability` filters, which control *who* can call the tool — not *which abilities* the tool is willing to talk about.

## What a filter would enable

If both reads were wrapped in a single filter — `mcp_adapter_is_ability_public( bool $is_public, WP_Ability $ability, ?McpServer $server )` — the default behavior stays identical (unpatched sites are unaffected), but integrators gain the ability to answer the "is this ability MCP-visible?" question dynamically **and per-server**.

### Example: different exposure per MCP server

A single site can register multiple MCP servers (public storefront server, internal ops server, per-tenant server, etc.). Today, an ability is either `mcp.public = true` for **all** of them or none of them. With the filter, an integrator can decide per server based on the server actually handling the request:

```php
add_filter( 'mcp_adapter_is_ability_public', function ( $is_public, $ability, $server ) {
    // Called outside a request context (WP-CLI, direct wp_get_ability(...)->execute()).
    if ( ! $server ) {
        return $is_public;
    }

    // Per-server allowlist stored by the operator in admin UI.
    $allowlist = my_plugin_get_allowlist_for_server( $server->get_server_id() );

    if ( in_array( $ability->get_name(), $allowlist, true ) ) {
        return true;
    }

    return $is_public;
}, 10, 3 );
```

Now `internal/delete-user` can be visible on the ops server but invisible on the public server, without touching the ability's own `meta`.

### Other use cases the filter unlocks

- **Role / capability-aware exposure** — only show `analytics/export-report` if `current_user_can( 'view_reports' )`, rather than gating with a static flag on the ability.
- **Tenant / site scoping** in multi-tenant setups — expose tenant-specific abilities only to the matching server.
- **Feature flags / phased rollout** — flip an ability from private to public via a runtime flag without re-registering it.
- **Auditability** — a downstream can log every exposure decision (which server accepted or rejected which ability) by hooking the filter.

## Why the workarounds don't cover this

- **Flip `mcp.public = true` on the ability** (directly or via `wp_register_ability_args` in core) — this is all-or-nothing. The same flag also opts the ability into every other MCP auto-exposure path (`RegisterAbilityAsMcpTool`, `RegisterAbilityAsMcpResource`, `RegisterAbilityAsMcpPrompt`, `DefaultServerFactory`). Integrators who want the built-in three tools to see the ability only on *specific* servers cannot express that.
- **Register the ability directly as a tool on a specific server** — clients can invoke it via the MCP protocol, but `discover-abilities` / `get-ability-info` / `execute-ability` on that same server still deny its existence, which is inconsistent from the client's point of view.

## Proposed change

Three small pieces, all inside `mcp-adapter`. The ability callbacks receive only `$input` (that's core `WP_Ability::execute()` behavior), so the server is threaded via a per-request holder on the `McpAdapter` singleton rather than through the ability call itself.

### 1. Track the current server on `McpAdapter`

Add one private field and two accessors on the existing singleton (`includes/Core/McpAdapter.php`):

```php
private ?McpServer $current_server = null;

public function set_current_server( ?McpServer $server ): void {
    $this->current_server = $server;
}

public function get_current_server(): ?McpServer {
    return $this->current_server;
}
```

### 2. Set / clear it around tool execution in `ToolsHandler`

In `includes/Handlers/Tools/ToolsHandler.php`, wrap the existing `$mcp_tool->execute( $args )` call in `try / finally` so exceptions still clear the holder:

```php
McpAdapter::instance()->set_current_server( $this->mcp );
try {
    $result = $mcp_tool->execute( $args );
} finally {
    McpAdapter::instance()->set_current_server( null );
}
```

Optional for consistency: apply the same wrapping in `ResourcesHandler` and `PromptsHandler`. Not strictly required for the three built-in tools (they're tool-only today) but keeps the accessor honest for any future use.

### 3. Wrap the two reads in `McpAbilityHelperTrait` with the filter

```php
protected static function is_ability_mcp_public( \WP_Ability $ability ): bool {
    $meta      = $ability->get_meta();
    $is_public = (bool) ( $meta['mcp']['public'] ?? false );
    $server    = \WP\MCP\Core\McpAdapter::instance()->get_current_server();

    /**
     * Filters whether an ability is considered publicly exposed via MCP
     * for the purposes of the built-in discover/get-info/execute tools.
     *
     * @param bool                                  $is_public Whether the ability's meta.mcp.public is truthy.
     * @param \WP_Ability                           $ability   The ability being checked.
     * @param \WP\MCP\Core\McpServer|null           $server    The MCP server handling the current request, or null
     *                                                          if the ability is invoked outside a request context
     *                                                          (e.g. WP-CLI or direct wp_get_ability()->execute()).
     */
    return (bool) apply_filters( 'mcp_adapter_is_ability_public', $is_public, $ability, $server );
}
```

Apply the same filter inside `check_ability_mcp_exposure()` before returning the `ability_not_public_mcp` `WP_Error`.

## Non-goals

- No change to the default behavior. The filter defaults to the existing computed value, so unpatched sites remain unaffected.
- No change to the `*_capability` filters — capability gating stays orthogonal.
- No change to `WP_Ability::execute()` / core WP.

Happy to open a PR if the approach is acceptable.

raftaar1191:
Closes #243.

## Summary

Adds a filter — `mcp_adapter_is_ability_exposed` — that lets integrators decide, per ability and per request context, whether the three built-in MCP tools (`discover-abilities`, `get-ability-info`, `execute-ability`) treat an ability as exposed. Default behavior is unchanged: the filter defaults to the ability's existing `meta.mcp.public` flag, so unpatched sites are unaffected.

## Motivation

Today the three built-in tools decide what a client is allowed to see or invoke by reading `meta.mcp.public` directly on each ability. That is a single global boolean baked into the ability, so exposure cannot vary by request context — most importantly, an ability cannot be visible on one MCP server and hidden on another.

See #243 for the full motivation and use cases.

## Filter shape

```php
apply_filters(
    'mcp_adapter_is_ability_exposed',
    bool                        $is_exposed,   // default: ability's meta.mcp.public
    \WP_Ability                 $ability,      // the ability being checked
    McpAbilityExposureContext   $context       // request context
);
```

`McpAbilityExposureContext` is an immutable value object with:

| field           | type                           | notes                                                        |
|-----------------|--------------------------------|--------------------------------------------------------------|
| `server`        | `McpServer\|null`              | Null when invoked outside an MCP request (WP-CLI, cron, direct execute) |
| `principal_id`  | `int`                          | 0 if unauthenticated                                         |
| `roles`         | `array<int, string>`           | Role slugs of the current user                               |
| `site_id`       | `int`                          | `get_current_blog_id()`                                      |
| `exposure_path` | `string` (PATH_* constant)     | `'discover'`, `'get_info'`, or `'execute'`                   |

Downstreams can differentiate exposure decisions by path — e.g. list an ability but block execution:

```php
add_filter( 'mcp_adapter_is_ability_exposed', function ( $is_exposed, $ability, $context ) {
    if ( ! $context->server ) {
        return $is_exposed; // outside an MCP request
    }
    if ( 'internal/delete-user' === $ability->get_name() ) {
        // Visible on the ops server for all paths; visible on the public
        // server for discovery only.
        if ( 'ops' === $context->server->get_server_id() ) {
            return true;
        }
        return McpAbilityExposureContext::PATH_DISCOVER === $context->exposure_path;
    }
    return $is_exposed;
}, 10, 3 );
```

## Design notes (why a context object, not positional args)

Ability callbacks only receive `$input` — that's core `WP_Ability::execute()` behavior — so the server can't be threaded through the ability call itself. Two things follow:

1. **Single point of context resolution.** A lightweight per-request holder on the existing `McpAdapter` singleton (`set_current_server()` / `get_current_server()`) is set by `ToolsHandler::call_tool()` on entry and cleared in `finally`. The trait's `build_exposure_context()` consumes that holder and the ambient user/site once, in one place. All three tools therefore see an identical context shape — no drift.
2. **Extensible surface.** New context fields (e.g. `transport`, `feature_flags`) can be added without changing the filter signature.

## Security contract (pinned in tests)

Two invariants are asserted by tests so future refactors can't silently break them:

- **Same helper, same context shape across the three tools.** `McpAbilityExposureInvariantsTest::test_exposure_filter_receives_identical_context_shape_across_all_three_tools()` registers one filter, invokes each of the three tools against the same ability, and asserts every non-path context field is identical.
- **Exposure is not authorization.** `ExecuteAbilityAbilityTest::test_exposure_filter_does_not_replace_ability_permission_check()` widens exposure with the filter but registers an ability whose `permission_callback` returns `WP_Error` — and asserts the execute path still rejects with the ability's error, not the exposure gate's.

The filter docblock also spells this out and calls out cache-key hygiene: if a downstream caches the exposure decision, the key must include the ability name, `exposure_path`, and every context field the callback reads. Missing dimensions cross-contaminate contexts.

## Backward compatibility

- The `ability_not_public_mcp` `WP_Error` code and message are preserved verbatim for clients catching them.
- The `meta.mcp.public` ability metadata key is unchanged (it's the static default the filter starts from).
- `mcp_adapter_is_ability_exposed` is new in this PR; no prior name shipped.

## Files

**Code (7 files)**
- `includes/Abilities/McpAbilityExposureContext.php` — new value object.
- `includes/Abilities/McpAbilityHelperTrait.php` — renamed helper, filter, `build_exposure_context()`.
- `includes/Abilities/DiscoverAbilitiesAbility.php` — passes `PATH_DISCOVER`.
- `includes/Abilities/GetAbilityInfoAbility.php` — passes `PATH_GET_INFO`.
- `includes/Abilities/ExecuteAbilityAbility.php` — passes `PATH_EXECUTE`.
- `includes/Core/McpAdapter.php` — `set_current_server()` / `get_current_server()`.
- `includes/Handlers/Tools/ToolsHandler.php` — set/clear the holder in `try/finally`.

**Tests (5 files)**
- `tests/phpunit/Unit/Core/McpAdapterCurrentServerTest.php` — accessor.
- `tests/phpunit/Unit/Abilities/McpAbilityExposureContextTest.php` — value object + path constants stability.
- `tests/phpunit/Unit/Abilities/McpAbilityExposureInvariantsTest.php` — SECURITY INVARIANT: same context shape across all three tools.
- `tests/phpunit/Unit/Abilities/ExecuteAbilityAbilityTest.php` — SECURITY INVARIANT: exposure ≠ authorization + renamed filter override test.
- `DiscoverAbilitiesAbilityTest.php`, `GetAbilityInfoAbilityTest.php` — renamed filter override tests asserting the context object shape.

## Verification

- `phpcs` (WPCS ruleset) clean on all modified `includes/` files.
- `phpstan --memory-limit=1G` clean on all modified `includes/` files.
- `php -l` clean on all touched files.
- PHPUnit not run locally (needs `wp-env`/Docker); will validate in CI.

## Non-goals

- No change to default behavior — the filter defaults to `meta.mcp.public`.
- No change to `*_capability` filters — capability gating stays orthogonal.
- No change to `WP_Ability::execute()` / core WP.
- Only `ToolsHandler` sets/clears the current-server holder in this PR. `ResourcesHandler` and `PromptsHandler` can adopt the same pattern in a follow-up if the built-in tools ever grow resource/prompt equivalents; the three built-in abilities today are tool-only.

### DIFF
--- a/includes/Abilities/DiscoverAbilitiesAbility.php
+++ b/includes/Abilities/DiscoverAbilitiesAbility.php
@@ -89,8 +89,8 @@ final class DiscoverAbilitiesAbility {
 		foreach ( $abilities as $ability ) {
 			$ability_name = $ability->get_name();
 
-			// Check if ability is publicly exposed via MCP
-			if ( ! self::is_ability_mcp_public( $ability ) ) {
+			// Check if ability is exposed via MCP for the discover path.
+			if ( ! self::is_ability_mcp_exposed( $ability, McpAbilityExposureContext::PATH_DISCOVER ) ) {
 				continue;
 			}
 
--- a/includes/Abilities/ExecuteAbilityAbility.php
+++ b/includes/Abilities/ExecuteAbilityAbility.php
@@ -174,7 +174,7 @@ final class ExecuteAbilityAbility {
 		}
 
 		// Check MCP exposure restrictions
-		$exposure_check = self::check_ability_mcp_exposure( $ability_name );
+		$exposure_check = self::check_ability_mcp_exposure( $ability_name, McpAbilityExposureContext::PATH_EXECUTE );
 		if ( is_wp_error( $exposure_check ) ) {
 			return $exposure_check;
 		}
--- a/includes/Abilities/GetAbilityInfoAbility.php
+++ b/includes/Abilities/GetAbilityInfoAbility.php
@@ -153,7 +153,7 @@ final class GetAbilityInfoAbility {
 		}
 
 		// Check MCP exposure restrictions
-		return self::check_ability_mcp_exposure( $ability_name );
+		return self::check_ability_mcp_exposure( $ability_name, McpAbilityExposureContext::PATH_GET_INFO );
 	}
 
 	/**
--- a/includes/Abilities/McpAbilityExposureContext.php
+++ b/includes/Abilities/McpAbilityExposureContext.php
@@ -1,0 +1,123 @@
+<?php
+/**
+ * Value object describing the context of an MCP ability exposure check.
+ *
+ * @package McpAdapter
+ */
+
+declare( strict_types=1 );
+
+namespace WP\MCP\Abilities;
+
+use WP\MCP\Core\McpServer;
+
+/**
+ * Describes *when* and *for whom* the built-in discover / get-info /
+ * execute tools are asking whether an ability is exposed via MCP.
+ *
+ * Passed as the third argument to the `mcp_adapter_is_ability_exposed`
+ * filter so integrators can make exposure decisions that vary by:
+ *
+ *   - the MCP server handling the current request (multiple servers can
+ *     be registered per site, each with its own scope),
+ *   - the authenticated principal and its WordPress roles,
+ *   - the current site (multisite / multi-tenant),
+ *   - the *path* by which the decision is being asked (listing vs.
+ *     inspecting vs. executing an ability — see the PATH_* constants).
+ *
+ * Treat instances as immutable: exposure decisions and any caches built
+ * on top of them should be keyed by the full context, not by a subset.
+ *
+ * @since 0.6.0
+ */
+final class McpAbilityExposureContext {
+
+	/**
+	 * Exposure path: the discover-abilities listing.
+	 *
+	 * @var string
+	 */
+	public const PATH_DISCOVER = 'discover';
+
+	/**
+	 * Exposure path: the get-ability-info inspection.
+	 *
+	 * @var string
+	 */
+	public const PATH_GET_INFO = 'get_info';
+
+	/**
+	 * Exposure path: the execute-ability invocation gate.
+	 *
+	 * Note: exposure is not authorization. An ability that is "exposed"
+	 * via this path still has its own `permission_callback` checked by
+	 * the execute-ability tool before the ability actually runs.
+	 *
+	 * @var string
+	 */
+	public const PATH_EXECUTE = 'execute';
+
+	/**
+	 * The MCP server handling the current request, if any.
+	 *
+	 * Null when the ability is invoked outside of an MCP request
+	 * (e.g. WP-CLI, cron, or a direct `wp_get_ability( ... )->execute()`
+	 * call), in which case per-server allowlists should fall back to
+	 * the ability's default (`$is_exposed`).
+	 *
+	 * @var \WP\MCP\Core\McpServer|null
+	 */
+	public ?McpServer $server;
+
+	/**
+	 * The authenticated principal's user ID. Zero if unauthenticated.
+	 *
+	 * @var int
+	 */
+	public int $principal_id;
+
+	/**
+	 * The authenticated principal's WordPress roles.
+	 *
+	 * @var array<int, string>
+	 */
+	public array $roles;
+
+	/**
+	 * The current site ID (`get_current_blog_id()`). Non-zero on both
+	 * multisite and single-site installs.
+	 *
+	 * @var int
+	 */
+	public int $site_id;
+
+	/**
+	 * The exposure path — one of the PATH_* constants above.
+	 *
+	 * @var string
+	 */
+	public string $exposure_path;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param \WP\MCP\Core\McpServer|null $server        The current server, or null.
+	 * @param int                         $principal_id  User ID (0 if not logged in).
+	 * @param array<int, string>          $roles         Role slugs.
+	 * @param int                         $site_id       Current site ID.
+	 * @param string                      $exposure_path One of the PATH_* constants.
+	 */
+	public function __construct(
+		?McpServer $server,
+		int $principal_id,
+		array $roles,
+		int $site_id,
+		string $exposure_path
+	) {
+		$this->server        = $server;
+		$this->principal_id  = $principal_id;
+		$this->roles         = $roles;
+		$this->site_id       = $site_id;
+		$this->exposure_path = $exposure_path;
+	}
+}
--- a/includes/Abilities/McpAbilityHelperTrait.php
+++ b/includes/Abilities/McpAbilityHelperTrait.php
@@ -11,6 +11,7 @@ namespace WP\MCP\Abilities;
 
 use WP\MCP\Core\McpAdapter;
 use WP_Error;
+use WP_User;
 
 /**
  * Trait McpAbilityHelperTrait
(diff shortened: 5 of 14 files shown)


shigeki7777:
Thanks for turning this into a concrete PR so quickly.

The current-server holder combined with `try / finally` looks like a pragmatic way to provide the server context without changing the `WP_Ability` callback contract. Centralizing both discovery and execution exposure checks through the same helper also addresses the consistency concern from #243.

A few small things seem worth pinning before merge:

- Add an integration-level test confirming that the filter receives the actual `McpServer` when invoked through `ToolsHandler::call_tool()`, not only through direct holder/filter tests.
- Keep the `ability_not_public_mcp` error message filter-agnostic. Since the filter can now hide an ability whose `mcp.public` value is true, the current `mcp.public!=true` wording may be misleading.
- Add a test confirming that returning `true` from the exposure filter does not bypass the separate capability/permission check for an unauthorized principal.

Passing the server as the third argument is already a useful and backward-compatible foundation. Broader tenant or principal context can be added later without blocking this improvement.

raftaar1191:
Thanks for the review — all three items addressed in `dfd614c`.

**1. Integration test through `ToolsHandler::call_tool()`** — new `tests/phpunit/Integration/ExposureFilterIntegrationTest.php`:
- `test_call_tool_threads_actual_server_into_exposure_filter()` — builds a real `McpServer` with `mcp-adapter/discover-abilities` in its tools, dispatches through `ToolsHandler::call_tool()`, and asserts the filter callback observed `$context->server === $server` (same instance) with `exposure_path === PATH_DISCOVER`.
- `test_call_tool_clears_current_server_after_dispatch()` — verifies the `finally` clause actually clears the holder so subsequent unrelated calls don't observe stale server state.

**2. Filter-agnostic error message** — `McpAbilityHelperTrait::check_ability_mcp_exposure()` now returns:

```
Ability "%s" is not exposed via MCP
```

Dropping the `(mcp.public!=true)` parenthetical — as you noted, that's misleading once the filter can hide an ability whose `mcp.public` value is `true`. The error **code** (`ability_not_public_mcp`) is intentionally unchanged to preserve BC for clients catching it; only the human-readable message is trimmed.

**3. Auth-gate test** — new `ExecuteAbilityAbilityTest::test_exposure_filter_does_not_bypass_capability_check_for_unauthorized_principal()`:
- Sets `wp_set_current_user( 0 )`, adds a filter returning `true`, uses an ability whose own `permission_callback` also returns `true`.
- Asserts the failure is `authentication_required` (from the tool's own auth gate), **not** `ability_not_public_mcp`, **not** success.
- Complements the existing `test_exposure_filter_does_not_replace_ability_permission_check` (which pins the ability-level permission gate). Together they assert: **the exposure gate runs neither before the auth gate nor in place of the ability's permission_callback.**

Agreed on the broader tenant/principal point — that's why the refactor to `McpAbilityExposureContext` in `c5d5156` already carries `principal_id`, `roles`, and `site_id` alongside `server` and `exposure_path`, so the same integration surface can grow richer context without another BC bump.

`php -l`, `phpcs`, and `phpstan` all clean on the touched files. Happy to squash before merge.

shigeki7777:
Nice turnaround — going by your writeup, all three items land the way we hoped:

1. **Integration test** — asserting `$context->server === $server` (same instance, not just same class) through a real `ToolsHandler::call_tool()` dispatch is exactly the pin we were after. The extra `test_call_tool_clears_current_server_after_dispatch()` is a welcome addition we didn't ask for: stale-holder leakage is the classic failure mode of the current-server pattern, and now it's pinned too.

2. **Message/code split** — trimming the message while freezing the `ability_not_public_mcp` error code is the right BC trade-off. One small thought, take or leave: the code name itself still implies the `mcp.public` semantics, so a one-line docblock noting it's kept for BC (and no longer means `mcp.public !== true`) would spare future readers the same confusion the old message caused.

3. **Auth-gate test** — asserting the failure is specifically `authentication_required` (not `ability_not_public_mcp`, not success) is stronger than a generic "it fails" check, and paired with the existing permission-callback test it covers both gates the exposure filter must not bypass.

One honesty note on where we stand: this is a read of your description, not an independent reproduction — we haven't pulled `dfd614c` and run the suite from this thread. Nothing in the writeup conflicts with what we reviewed, and we'll re-run our checks against the merged result on our next verification pass.

Squash vs. not is the maintainers' call; from our side it doesn't affect anything we flagged, since the error-code BC holds either way.

PR #246, opened 2026-07-20
Fix concurrent session overwrites

### LINKED ISSUE #239: Race condition in SessionManager: concurrent initialize calls for the same user can silently overwrite each other's session
**Summary**

`SessionManager` stores all of a user's MCP sessions as a single serialized array under one user meta key, using a read-modify-write cycle with no locking:

1. Read the full sessions array (`get_user_meta`)
2. Add/modify an entry in PHP
3. Write the entire array back (`update_user_meta`)

If two `initialize` requests for the same WordPress user land close together (e.g. two MCP clients/connectors starting up around the same moment), both can read the same starting array before either writes back. Whichever write lands second silently overwrites the first session out of existence — even though it was already returned to its client as a valid session ID. Every subsequent call on the lost session then fails permanently with a session-not-found error (surfaced as either HTTP 200 with a JSON-RPC error body, or HTTP 404 depending on context), even though nothing was actually wrong with that session.

**Where**

`includes/Transport/Infrastructure/SessionManager.php` — this pattern appears in `create_session()`, and likely affects `delete_session()`/session cleanup similarly, since they follow the same unlocked read-modify-write shape.

**Impact**

Any deployment where multiple MCP clients or connectors can authenticate as the same WordPress user (a common pattern for agent/service accounts) is exposed to this. It's intermittent and timing-dependent, which makes it easy to misdiagnose as a caching or network issue rather than an application-level race.

**Suggested fix direction**

Store one meta row per session (`add_user_meta( $user_id, self::SESSION_META_KEY, $session_data, false )` for creation, `delete_user_meta()` for removal, `get_user_meta( $user_id, self::SESSION_META_KEY, false )` to read all) instead of one serialized array per user. This turns session creation/deletion into independent atomic operations and removes the race entirely.

chubes4:
Closes #239.

## Summary

- Store each MCP session in an independent `mcp_adapter_sessions` user-meta row instead of rewriting one serialized map containing every session.
- Target activity updates, cleanup, eviction, and deletion to one session row using its complete previous value.
- Lazily migrate the released serialized-map format to independent rows while preserving active session IDs.
- Add regression coverage for independent persistence, legacy migration, and sibling preservation across cleanup, update, and deletion.

## Why

Concurrent `initialize` requests currently perform an unlocked read-modify-write of the same user-meta value. Both requests can return valid session IDs, after which the later write removes the earlier session. The affected client then fails its first session-bearing request, commonly `tools/list`.

Independent metadata rows make session creation an append operation. Mutating one session no longer replaces a collection snapshot containing its siblings.

## Backward compatibility

- Public `SessionManager` methods and returned session shapes are unchanged.
- Existing `mcp_adapter_sessions` map values are persisted user state, so they are detected and migrated on first read rather than invalidated.
- Session limits, FIFO eviction, inactivity cleanup, and activity-write throttling retain their existing behavior.
- The user-meta representation is internal but changes from one collection row to multiple per-session rows. Integrations reading this private storage directly must use the public `SessionManager` API instead.

## How to test

1. Check out this branch on a fresh clone.
2. Run `npm ci && composer install --no-interaction`.
3. Start the test environment with `npm run wp-env:test -- start`.
4. Run `npm run test:php`.
5. Run `composer lint`.
6. Run `composer phpstan`.
7. Confirm `McpSessionManagerTest` covers four independent sessions, legacy migration, and targeted cleanup/update/delete without losing siblings.

## Verification

- `composer lint`: passed.
- `composer phpstan`: passed with no errors.
- `git diff --check`: passed.
- Transactional WordPress runtime proof: four sessions persisted, legacy migration succeeded, targeted cleanup/deletion preserved all siblings, and fixture state rolled back.
- Supplementary WPCOM sandbox proof using the same regression fixture:
  - Released collection storage: expected failure, 2 tests / 2 failures.
  - This change: passed, 2 tests / 22 assertions.

The WPCOM sandbox result is supplementary foreign-environment evidence. The primary test procedure above uses this repository's own tooling and is runnable from a fresh checkout.

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** OpenCode with OpenAI GPT-5.6-sol
- **Used for:** Reproduction analysis, implementation and regression-test drafting, review, and verification. Chris reviewed and owns the submitted change.

### DIFF
--- a/includes/Transport/Infrastructure/SessionManager.php
+++ b/includes/Transport/Infrastructure/SessionManager.php
@@ -21,7 +21,7 @@ use WP_Error;
 final class SessionManager {
 
 	/**
-	 * User meta key for storing sessions
+	 * User meta key for storing sessions.
 	 *
 	 * @var string
 	 */
@@ -79,21 +79,26 @@ final class SessionManager {
 				}
 			);
 
-			array_shift( $sessions );
+			$oldest_session_id = key( $sessions );
+			if ( is_string( $oldest_session_id ) ) {
+				self::delete_session( $user_id, $oldest_session_id );
+			}
 		}
 
 		// Create a new session
 		$session_id = wp_generate_uuid4();
 		$now        = time();
 
-		$sessions[ $session_id ] = array(
+		$session = array(
 			'created_at'    => $now,
 			'last_activity' => $now,
 			'client_params' => $params,
 		);
 
-		// Save sessions
-		update_user_meta( $user_id, self::SESSION_META_KEY, $sessions );
+		// Each session has its own row, so concurrent requests cannot overwrite siblings.
+		if ( false === add_user_meta( $user_id, self::SESSION_META_KEY, self::prepare_stored_session( $session_id, $session ) ) ) {
+			return false;
+		}
 
 		return $session_id;
 	}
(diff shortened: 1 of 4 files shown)


galatanovidiu:
Clowsing in favor of https://github.com/WordPress/mcp-adapter/pull/251

PR #251, opened 2026-07-22
Prevent concurrent MCP session overwrites

### LINKED ISSUE #239: Race condition in SessionManager: concurrent initialize calls for the same user can silently overwrite each other's session

galatanovidiu:
Closes https://github.com/WordPress/mcp-adapter/issues/239.

## Summary

- Protect updates to the existing `mcp_adapter_sessions` user-meta map with `update_user_meta( ..., $prev_value )` and bounded retries.
- Keep the current single-meta storage format and all existing public session behavior.
- Preserve an empty array after the final session is deleted.
- Report exhausted retries through the configured MCP error handler, with a standard error-log fallback.
- Add real multi-process concurrency coverage and failure-observability coverage.

## Why

Session mutations currently perform an unconditional read-modify-write. Two requests can read the same session map and then overwrite one another, dropping sessions.

Instead of changing the storage model to one user-meta row per session, as explored in #246, this uses WordPress's existing previous-value comparison as a small compare-and-swap loop. It does not add custom SQL, database locks, lock metadata, or a storage migration.

## Compatibility and tradeoff

The existing `mcp_adapter_sessions` key and serialized map remain unchanged. Existing `HttpSessionValidator` public method signatures are preserved, while optional trailing handler arguments on the final `SessionManager` class remain backward compatible.

WordPress ignores an empty `$prev_value`, so simultaneous first connections for a user, or simultaneous connections immediately after the map becomes empty, can still race. This is an intentional tradeoff: the common established-session case is protected without introducing a more complex locking or storage system. MCP protocol revisions that do not negotiate session IDs continue to bypass this compatibility storage.

Retry-exhaustion logs contain only the component, method, user ID, and attempt count. They do not include session IDs, client parameters, headers, or stored payloads.

## Validation

- `npm run test:php` — 1,011 tests, 3,920 assertions
- `vendor/bin/phpcs`
- `vendor/bin/phpstan analyze --memory-limit=1G --no-progress`
- Four independent PHP worker processes synchronize at the user-meta update boundary and verify that every concurrent session survives.

### Live concurrency testing

Tested against real deployments, changing only `SessionManager.php` between control and patched runs:

- With sessions already stored, the current code lost 7 of 18 and 9 of 18 sessions across two servers; patched runs lost none.
- With the meta row deleted before every round, the current code lost 20 of 36 sessions; the patch lost 0 of 114 sessions at up to 12 concurrent clients.
- Across all patched runs, 0 of 186 sessions were lost.

The known first-write limitation remains: WordPress ignores an empty `$prev_value`, so the initial write cannot be protected by this compare-and-swap mechanism. I could not reproduce a lost session in that case, so I can call this a good trade-off.

### DIFF
--- a/includes/Transport/Infrastructure/HttpRequestHandler.php
+++ b/includes/Transport/Infrastructure/HttpRequestHandler.php
@@ -198,7 +198,7 @@ class HttpRequestHandler {
 
 		// Validate session for all requests except initialize (router will handle initialize session creation)
 		if ( 'initialize' !== $method ) {
-			$session_validation = HttpSessionValidator::validate_session( $context );
+			$session_validation = HttpSessionValidator::validate_session_with_error_handler( $context, $this->transport_context->error_handler );
 			if ( true !== $session_validation ) {
 				return JsonRpcResponseBuilder::create_error_response( $request_id, $session_validation['error'] ?? $session_validation );
 			}
@@ -325,7 +325,7 @@ class HttpRequestHandler {
 	 * @return \WP_REST_Response Termination response.
 	 */
 	private function handle_session_termination( HttpRequestContext $context ): \WP_REST_Response {
-		$result = HttpSessionValidator::terminate_session( $context );
+		$result = HttpSessionValidator::terminate_session_with_error_handler( $context, $this->transport_context->error_handler );
 
 		if ( true !== $result ) {
 			$http_status = McpErrorFactory::get_http_status_for_error( $result );
--- a/includes/Transport/Infrastructure/HttpSessionValidator.php
+++ b/includes/Transport/Infrastructure/HttpSessionValidator.php
@@ -9,6 +9,7 @@ declare( strict_types=1 );
 
 namespace WP\MCP\Transport\Infrastructure;
 
+use WP\MCP\Infrastructure\ErrorHandling\Contracts\McpErrorHandlerInterface;
 use WP\MCP\Infrastructure\ErrorHandling\McpErrorFactory;
 
 /**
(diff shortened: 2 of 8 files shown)


gziolo:
This looks considerably more promising than the per-session-row alternative. Keeping the existing storage format while introducing bounded compare-and-swap retries is a thoughtful approach.

Two potential enhancements may be worth evaluating:

- As Copilot noted, clear the user-meta cache only after a conditional update fails. The first attempt could use the cached value, with a failed comparison triggering a fresh read and retry.
- Consider representing an empty session map with a non-empty internal sentinel, so the previous-value condition remains active. The missing-row bootstrap would still need an atomic initialization mechanism, but this could help extend compare-and-swap protection to the empty state.

Neither seems fundamental to the overall direction, but both may improve its completeness and efficiency.

gziolo (approved):

galatanovidiu:
Good point on the cache, I tried it and dropped it.

The win is small: `validate_session()` goes from 1 query to 0 on a warm cache.

But `update_meta_cache()` fills the cache with `wp_cache_add_multiple()`, so with a persistent object cache an empty session map can land in cache right after another request invalidated it. WordPress ignores an empty `$prev_value`, so the next write skips the comparison and overwrites the stored session. Same lost session this PR is fixing.

I can guard that with a re-read, but we would overcomplicate it for a small gain. Clearing before every read keeps `mutate_sessions()` correct without depending on how core caches meta.

On the sentinel: it changes the stored format, so I would rather keep that out of this PR.

PR #254, opened 2026-07-23
Inherit public ability exposure for MCP

gziolo:
## Summary

- resolve MCP exposure from an ability's stored metadata through `McpAbilityExposure::is_public()`
- inherit `meta.public` when no MCP-specific value is set, and keep explicit `meta.mcp.public` opt-ins and opt-outs authoritative
- route discovery, inspection, execution, resource and prompt exposure checks through the shared resolver
- cover the exposure precedence matrix and late `meta.public` changes
- update the README and guides to recommend the high-level public flag

## Why

[WordPress Core ticket #65568](https://core.trac.wordpress.org/ticket/65568) introduces `meta.public` in version 7.1 as the high-level client-exposure setting, with channel-specific settings taking precedence. MCP Adapter recognizes only `meta.mcp.public`, so abilities using the Core flag are not available through the default MCP server.

Exposure is resolved when the adapter reads a registered ability rather than while the ability is being registered. WordPress runs same-priority callbacks in registration order, so no `wp_register_ability_args` priority is guaranteed to run after every other callback, and `meta.public` is final only once registration has completed. Resolving on read also leaves `WP_Ability::get_meta()` unmodified, so a missing `meta.mcp.public` continues to mean the value is inherited.

## Impact

Abilities with `meta.public: true` become available through the default MCP server unless they set `meta.mcp.public: false`. An explicit `meta.mcp.public: true` continues to expose an otherwise private ability to MCP. A malformed `meta.mcp` resolves to not public. Authentication, capability checks, and each ability's permission callback are unchanged.

`McpAbilityExposure::is_public()` is the supported way to check MCP exposure. Reading `meta.mcp.public` directly does not account for inheritance.

## Validation

All CI checks pass.

Assisted with Codex.


includes/Core/McpAdapter.php
```diff
@@ -70,6 +73,38 @@ public static function instance(): self {
...
 
+	/**
+	 * Inherit the high-level public ability setting for MCP exposure.
+	 *
+	 * An explicit MCP-specific setting takes precedence. This mirrors the
+	 * Abilities API exposure cascade, where a channel-specific value overrides
+	 * the high-level public value.
+	 *
+	 * @since 0.6.0
+	 *
```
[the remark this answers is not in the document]
galatanovidiu: Good catch. This repo uses `@since n.e.x.t` for all unreleased code, so the fix here is to replace `0.6.0` with the placeholder, not bump the plugin version in this PR.

includes/Core/McpAdapter.php
```diff
@@ -70,6 +73,38 @@ public static function instance(): self {
...
+		}
+
+		$mcp_meta = $meta['mcp'] ?? array();
+
+		if ( ! is_array( $mcp_meta ) || isset( $mcp_meta['public'] ) ) {
+			return $args;
+		}
+
+		$mcp_meta['public']  = true;
+		$args['meta']['mcp'] = $mcp_meta;
```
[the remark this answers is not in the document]
galatanovidiu: I don’t think we should normalize malformed `meta.mcp` to an empty array here. The current chained `??` reads fail closed without warnings on PHP 7.4 and 8.4; this autofix instead turns `meta.public=true` plus malformed MCP metadata into MCP exposure and discards the invalid value. I’d keep this fail-closed, or validate and reject malformed metadata separately. This new branch also has no regression test, and Codecov is now failing on the two uncovered lines.

includes/Core/McpAdapter.php
```diff
@@ -58,6 +58,9 @@ public static function instance(): self {
 		if ( ! isset( self::$instance ) ) {
 			self::$instance = new self();
 
+			// Inherit the high-level ability exposure setting before abilities are registered.
+			add_filter( 'wp_register_ability_args', array( self::$instance, 'inherit_public_ability_exposure' ) );
```
galatanovidiu: This runs at priority 10 and materializes `meta.mcp.public` before the remaining `wp_register_ability_args` callbacks run. A priority-20 filter changing only `meta.public` can therefore leave final MCP exposure inverted; I reproduced both `true → false` (still exposed) and `false → true` (still hidden) on WordPress 7.0.2. I’d resolve effective exposure from the final metadata, or deliberately run this transform last, and add both ordering cases as regressions.

README.md
```diff
@@ -179,19 +180,20 @@ add_action( 'wp_abilities_api_init', function() {
...
 });
 
-// With the meta.mcp.public flag, the ability is exposed through the default MCP server.
+// With the meta.public flag, the ability is exposed through the default MCP server.
 // In the default server configuration, discover it via `discover-abilities`
 // and invoke it via `mcp-adapter/execute-ability` rather than expecting
 // it to appear as its own entry in `tools/list`.
-// Without the meta.mcp.public flag, abilities are only accessible
-// through custom MCP servers that explicitly list them.
+// To opt out of MCP while remaining public to other clients, explicitly set
```
galatanovidiu, line 192: This cross-client wording needs a WordPress version qualifier. Core only starts cascading `meta.public` into REST’s `show_in_rest` in WordPress 7.1. On supported WordPress 6.9/7.0, this adapter honors `meta.public` for MCP, but REST still needs `show_in_rest=true`; could we clarify that boundary here and in the other updated examples?

galatanovidiu (changes requested):
I found three things that need tightening before merge.

First, the inheritance filter runs at priority 10. A later `wp_register_ability_args` policy that changes only `meta.public` leaves the copied `meta.mcp.public` stale. I reproduced both directions on WordPress 7.0.2: `true → false` stayed exposed to MCP, and `false → true` stayed hidden.

Second, the docs describe `meta.public` as cross-client exposure across the plugin's supported versions. Core only adds that broader cascade in WordPress 7.1. On 6.9/7.0 the adapter makes it work for MCP, but REST still needs `show_in_rest=true`.

Third, the latest malformed-`meta.mcp` autofix changes invalid metadata from fail-closed to MCP-exposed when `meta.public=true`. The existing reads do not produce the claimed warnings on PHP 7.4 or 8.4, so I don't think we should silently broaden exposure here. The new branch is also missing regression coverage, and both Codecov checks currently fail at 84.61% patch coverage.

The core direction and authorization checks look sound. PHPCS, PHPStan, the PHP-version/WordPress-version test matrix, and Plugin Check all pass. I also replied directly to the existing Copilot threads for the `@since` convention and malformed metadata.

README.md
gziolo: We can add as much detail as necessary. From the MCP point of view, the WP version doesn't matter, but as you pointed out, setting `meta.public` to true doesn't mean the ability would automatically opt in to exposing it to the REST API. However, we might still want to polyfill that in the AI plugin.

includes/Core/McpAdapter.php
gziolo: Good catch. It feels like running this filter as the very last would do the trick. We need to ensure that all plugins can opt in/out from `meta.public` and `meta.mcp.public` for the ability, so resolving at that point should only happen if `meta.mcp.public` array key does not exist.

gziolo: Addressed in https://github.com/WordPress/mcp-adapter/pull/254/commits/eb162d7ffd83fcfb243d0ca15990769d7e33c5fd.

README.md
gziolo: Improved with https://github.com/WordPress/mcp-adapter/pull/254/commits/9b41e4a2a52dbe09dbaadd5e33457dae7d8fcc65.

galatanovidiu (approved):
Most of the previous review looks good now.

There is one remaining edge case that I think we should consider. `PHP_INT_MAX` is the last numeric priority, but WordPress runs callbacks with the same priority in registration order. A plugin registering later at `PHP_INT_MAX` can still change only `meta.public` after this callback, leaving `meta.mcp.public` stale. I reproduced both directions: opt-out stayed MCP-public, and opt-in stayed hidden.

I tested one possible approach locally: remove the registration-time inheritance filter and resolve exposure only when the adapter reads the final `WP_Ability` metadata. The prototype used one shared resolver like this:

```php
final class McpAbilityExposure {
	public static function is_public( array $meta ): bool {
		$mcp_meta = $meta['mcp'] ?? array();

		if ( ! is_array( $mcp_meta ) ) {
			return false;
		}

		if ( isset( $mcp_meta['public'] ) ) {
			return (bool) $mcp_meta['public'];
		}

		return true === ( $meta['public'] ?? false );
	}
}
```

I called this resolver from `McpAbilityHelperTrait` for discovery, info, and execution, and from `DefaultServerFactory` for resources and prompts. Because the resolver reads the stored ability after registration, every `wp_register_ability_args` callback has already run. An explicit non-null `meta.mcp.public` wins, a missing or null value inherits `meta.public`, and malformed `meta.mcp` remains hidden.

I tested both same-`PHP_INT_MAX` ordering cases, the full precedence matrix, and every default-server consumer listed above. The repository's full PHPUnit suite, PHPCS, and PHPStan passed with the prototype.

I don't think `WP_Ability::get_meta()` should change or return a generated `meta.mcp.public`. It should keep the actual registered metadata, since a missing MCP-specific value tells us that exposure is inherited from `meta.public`.

This is only a proposal, not a request to use this exact design. Keeping the current `PHP_INT_MAX` approach is also reasonable, and this edge case is not blocking my approval. I'll leave the choice between the two approaches to you. Whichever approach you choose, we should make sure the behavior is documented properly for people and AI agents using or maintaining MCP Adapter.

gziolo:
Thank you for testing this so thoroughly, including the same-priority ordering case. I agree with your reasoning. There is no priority that guarantees running after every other callback, so registration-time inheritance can always go stale. Resolving from the final ability metadata at read time is the only safe point. It also keeps `get_meta() `honest. A missing `meta.mcp.public` still tells us the exposure is inherited from `meta.public`.

The trade-off is that every read must go through the shared resolver. A direct `$meta['mcp']['public']` read anywhere would quietly reintroduce the bug. So the agreed direction is: centralize all read sites on one helper method, and document that it is the only supported way to check MCP exposure.

I'm away next week, so anyone is welcome to pick this up in the meantime, either here or in a follow-up PR. Otherwise I will update the PR when I return.

galatanovidiu:
Picked this up while Greg is away. Went with the read-time resolver from the last review round.

## The problem

Copying `meta.public` into `meta.mcp.public` during registration is a cache with no way to invalidate it. A later `wp_register_ability_args` callback that changes only `meta.public` leaves the copy behind.

`PHP_INT_MAX` does not fix this. WordPress runs same-priority callbacks in registration order, and `McpAdapter::instance()` is called when the plugin file loads (`mcp-adapter.php:57`), before `plugins_loaded`. So our callback is always first in the `PHP_INT_MAX` bucket, and any plugin that hooks there later always runs after us. That makes it the default outcome, not a rare race.

I reproduced both directions on WP 7.0.2, with a plugin that hooks `wp_register_ability_args` at `PHP_INT_MAX` on `plugins_loaded`:

- final `meta.public = false`: the ability stayed visible in `discover-abilities`, and `execute-ability` returned `{"success":true,"data":"ok"}`
- final `meta.public = true`: the ability stayed hidden

The first one is why I did not want to merge as is. A site policy revoked exposure and MCP still ran the ability.

## The fix

New `includes/Abilities/McpAbilityExposure::is_public()`, and the registration filter is removed. Exposure is resolved when the adapter reads the stored ability, which happens after every `wp_register_ability_args` callback has run, so the metadata is final by then.

The rules are unchanged: an explicit `meta.mcp.public` wins, missing or null inherits `meta.public`, malformed `meta.mcp` fails closed.

Three read sites go through it, and they were the only direct readers:

- `McpAbilityHelperTrait::check_ability_mcp_exposure()` and `::is_ability_mcp_public()`
- `DefaultServerFactory::discover_abilities_by_type()`

`WP_Ability::get_meta()` is untouched now, so a missing `meta.mcp.public` keeps meaning "inherited from `meta.public`", which was Greg's point.

## Tests

The three assertions that checked `get_meta()['mcp']['public']` had been rewritten now assert the resolved value instead. The discovery assertions were already correct.

`data_late_public_exposure_changes` also runs at `PHP_INT_MAX` now, not only at priority 20. That is the case the old tests missed: priority 20 loses to `PHP_INT_MAX` cleanly, so the suite passed while the bug was live. I checked that the new cases fail on the previous code and pass on this one.

Full suite green at 1019 tests, PHPCS and PHPStan L8 clean, and CI passes across the whole matrix from PHP 7.4 to 8.4 on WP 6.9, latest and trunk. Also verified against wp-env on WP 7.0.2 through the real endpoint: `discover-abilities`, `execute-ability`, `resources/list` and `prompts/list` all resolve inheritance correctly.

## Two notes

- `'public' => 1` resolves to not public, because the check is a strict `true ===`. Keeping it that way. Core already throws `InvalidArgumentException` for a non-bool `meta.show_in_rest` (`class-wp-ability.php:319`), so real booleans are the house style for ability meta, and anyone who needs `true` will write `true`.
- The docs note says core applies `meta.public` to REST in 7.1. I confirmed WP 7.0.2 has no `public` handling at all in `wp-includes/abilities-api/`, but I could not open the trac ticket to check the milestone. Worth a second pair of eyes on that line.

Issue #176, opened 2026-04-27
Approval interface for tool execution? (gap after wordpress-mcp deprecation)


ibrahimhajjaj:
Hey, opening this as a discussion / proposal rather than a bug.

  Was reading `wordpress-mcp` and `mcp-adapter` side by side this week and noticed a thing. `wordpress-mcp` has a real approval workflow, `user_confirmed`, `confirm_permanent_delete`, soft-trash, draft-by-default for content creation. That repo is being deprecated in favor of this one. I grepped `mcp-adapter` at HEAD for anything equivalent (`approval`, `confirm`, `consent`, `pending`, `user_confirmed`) and got zero hits.

  Which means if anyone shipped abilities gated on the old approval flow, they don't really have a migration path right now. That feels like a gap worth at least naming out loud.

  [The MCP spec's March 2026 ](https://blog.modelcontextprotocol.io/posts/2026-03-16-tool-annotations/)annotations post is pretty clear that `destructiveHint` and friends are advisory, not enforcement, so the spec itself isn't going to fill this. It's an adapter / plugin layer concern. 

  ## What I'd propose

  Not a queue, not a UI, not policy. Just a seam plugins can hook into. Three small pieces:

  1. A filter fired before tool execution that a plugin can short-circuit to return "pending approval" instead of running:

      ```php
      $decision = apply_filters(
          'mcp_adapter_pre_tool_execution',
          null, // null = proceed as today
          $tool_name,
          $arguments,
          $context // user_id, client_id, request_id
      );
      ```

      Plugin returns a decision object with state `pending` and the adapter responds with a structured pending payload instead of executing. If nothing hooks the filter, behavior is identical to today.

  2. A capability name, something like `mcp_approve_tool_call`, defined as a constant. The adapter doesn't enforce who has it, just gives plugins doing approval queues something to converge on instead of each picking their own.

  3. A documented state enum for the pending response so clients know what to poll for: `pending | approved | rejected | executed | expired`.

  That's the whole thing. Roughly a 100-line PR if you want me to put one up.

  ## Why this can't just live in a plugin

The enforcement (queue table, UI, role checks, state transitions, etc) absolutely should live in plugins, that's not in question. But plugins can't intercept tool execution from outside the adapter without something like this filter. Today the workarounds are wrapping `execute_callback` at the abilities-api layer, which works but doesn't see MCP-specific context (`client_id`, `request_id`), or forking the adapter, which nobody actually wants to do.

[...]

PR #256, opened 2026-07-25
Allow successful pre-tool-call completion

### LINKED ISSUE #255: Allow pre-tool-call middleware to complete a call successfully
## Problem

`mcp_adapter_pre_tool_call` currently supports two outcomes after permission checks:

- Return an argument array and execute the tool.
- Return `WP_Error` and complete the call with `isError: true`.

There is no equivalent way for middleware to complete the current call successfully without invoking the tool executor. Synchronous use cases such as challenge/response preconditions, idempotent replay, cache-backed responses, or delegation therefore need to replace the execution path rather than compose with the adapter's pre-call extension point.

The post-call filter cannot cover these cases because the tool side effect has already happened.

## Proposal

Allow pre-tool-call middleware to make a typed decision with three outcomes:

- `proceed( $args )`: execute once with the supplied arguments.
- `complete( CallToolResult $result )`: return a protocol-valid result immediately without executing the tool.
- `reject( WP_Error $error )`: retain the existing error short circuit.

The exact class and API can follow maintainer preference. A decision object would make the outcomes explicit while preserving the current array and `WP_Error` return values for compatibility.

`complete()` is synchronous. It returns a normal MCP `CallToolResult` for the current request. It introduces no polling state, approval vocabulary, persistence, server push, SSE, or transport-specific behavior. Domain-specific semantics remain in the returned content and in the plugin that owns the middleware.

## Expected behavior

- Tool permission checks still run before middleware can complete a call.
- Existing argument-array returns continue to modify arguments and execute the tool once.
- Existing `WP_Error` returns continue to produce `isError: true`.
- A successful completion returns `isError: false` and never calls `McpTool::execute()`.
- Invalid middleware return values fail deterministically rather than reaching the executor with an invalid argument type.
- Tests verify both the returned MCP result and the absence of tool side effects.

## Relationship to #176

#176 proposes an approval lifecycle with pending/approved/rejected states and discusses polling or elicitation. This request is the smaller generic execution primitive underneath synchronous response patterns. It does not define an approval protocol or solve asynchronous approval, so it does not require a server-to-client channel.

The existing hook already provides successful `proceed` and error `reject` behavior. Adding successful `complete` would make the pre-execution extension point symmetric without teaching MCP clients any adapter-specific lifecycle.

chubes4:
## Summary

- preserve `mcp_adapter_pre_tool_call` as an argument-array / `WP_Error` hook
- add a separate filter for registering successful pre-execution completion callbacks
- invoke completion callbacks in deterministic array order and stop on the first successful `CallToolResult`
- reject invalid hook and callback return values before invoking the tool executor
- keep permission checks ahead of argument filters and completion callbacks

This is a synchronous, transport-neutral execution primitive. It introduces no approval states, polling, persistence, SSE, server push, or domain-specific context semantics.

## Behavior

`mcp_adapter_pre_tool_call_completion_callbacks` filters an array of callbacks. Each registered callback receives the final filtered arguments, tool name, MCP tool, and server, then returns:

- `null` to continue to the next completion callback
- a successful `CallToolResult` to complete the request immediately

The handler, rather than `WP_Hook`, invokes those callbacks. The first completion result is terminal: remaining completion callbacks and `McpTool::execute()` are not invoked.

Existing `mcp_adapter_pre_tool_call` callbacks continue to receive only arrays and may return an argument array or `WP_Error`. A completion result can therefore never become the input to a later legacy array-typed callback.

## Verification

- `composer lint` passes.
- `composer phpstan` passes.
- PHP syntax and `git diff --check` pass.
- PHPUnit requires the repository's canonical WordPress test environment; the GitHub Actions PHP 7.4-8.4 and WordPress 6.9/latest/trunk matrix provides that verification.

Behavioral coverage verifies legacy array and `WP_Error` handling, completion after argument filtering, terminal first-completion semantics across multiple registered callbacks, permission-before-completion ordering, invalid returns, and absence of executor side effects.

## AI Assistance

OpenCode with Homeboy used `openai/gpt-5.6-terra` to draft the initial implementation and tests. OpenCode using `openai/gpt-5.6-sol` reviewed and refactored the implementation into the separate terminal completion pipeline, added composition coverage, and ran deterministic gates. Chris Huber reviewed and owns the submitted changes.

Closes #255.
Related to #176.


galatanovidiu:
I went through what a plugin can already do today, and then compared it with what this PR adds.

What we have today:

- A plugin that owns an ability can put the logic in the ability's own callback: return a cached result, skip work that already ran, hand the job to another service, replay an old result.
- `mcp_adapter_pre_tool_call` can change the arguments, or return a `WP_Error` to stop the call.
- `mcp_adapter_tool_call_result` runs after execution and can rewrite the result before it goes back.
- A plugin that needs to work in a different way can register its own tool, server or transport.

What this PR adds on top of that is one thing: a plugin can return a successful result from the pre-call hook, so the ability's code never runs. For a plugin that owns the ability, that is a shortcut for something it can already do inside the callback. The new part is for a plugin that does not own the ability. Plugin A can now say that plugin B's tool succeeded, and plugin B's code never runs.

My question is about that last case. Is someone building it? On which tools, and why can the logic not sit in the ability? Going through the cases in #255, each one looks like something the ability owner can do today. If I am wrong on one of them, tell me which, that is the interesting one.

That is also the part I am careful about. If a plugin can report success for a tool it does not own, then a bug or a hacked plugin can tell the client that a delete or a payment went through when nothing ran, and the client cannot tell the difference. A plugin can already do worse if it wants to, since it runs as PHP inside WordPress with no sandbox. But there is a difference between something a plugin can force its way into and something we hand out as a documented API. I would want a real use case before we take that on.

One more thing if we do go ahead. `mcp_adapter_pre_tool_call` is a normal WordPress filter, so `apply_filters()` passes what one callback returns into the next one. I tried two callbacks with the real `WP_Hook`: after `complete()`, an existing `function ( array $args ): array` callback throws a `TypeError` and the client gets an internal error, and an untyped callback can return an array that replaces the decision, so the tool runs anyway. `complete()` does not really stop anything, the last callback still wins.

To say where I am coming from. I would like the adapter to stay a thin layer that exposes abilities over MCP. It should adapt an ability to the protocol, not take over what the ability does. Once a plugin can decide what someone else's ability returns, we are past adapting. That is the line I am trying to hold, and it is why I keep asking who owns the ability in each of these cases.

chubes4:
For my use case, it does not belong in the ability layer, because it assumes the caller is an AI. Abilities are AI-agnostic. MCP is the AI communication layer. 

The exact use case for this is to have the first tool call return the required context and instruct the model to try again. This provides a deterministic guarantee that the agent has the information it needs to execute the task effectively without the need for abilities that exist purely to provide context for other abilities. It can reduce the total number of tools and the size of the initial context window using progressive disclosure. 

For example, having a `create_post` tool return content guidelines on the first call and execute on the second.

chubes4:
The corresponding Agents API proof is now open at https://github.com/Automattic/agents-api/pull/466. It documents and deterministically tests the same successful pre-tool completion sequence through Agents API's existing transport-neutral `pre_tool_mediator` / `replace_result` contract. This keeps model-loop ownership in Agents API while this PR remains the narrow hook for external MCP clients that bypass that loop.

galatanovidiu:
I see why the context is MCP-specific rather than part of the ability itself, but I still do not see a missing capability in the adapter.

`mcp_adapter_pre_tool_call` can already return a `WP_Error` before execution. On the first `create_post` call, the mediator can return the content guidelines, a receipt, and instructions to retry. The adapter turns that into a normal MCP `CallToolResult` with `isError: true`, the ability does not run, and the model receives the message so it can retry. When the receipt is present, the filter returns the arguments and the ability executes. That is the same sequence.

Since the post was not created on the first call, a tool error also seems like the accurate result. This PR changes the classification from an actionable tool error to success and adds `structuredContent`, but it does not enable the workflow.

The Agents API PR shows that successful replacement works in that provider loop. It does not show why an external MCP client cannot use the existing tool-error path. Is there a specific client that does not pass `isError: true` tool results to the model or prevents the model from retrying? If so, that is the missing case I would like to understand. Otherwise, if structured error data is the gap, I would rather improve the existing error-result mapping than add a global successful-execution bypass.

@JasonTheAdams @justlevine @gziolo @jorgefilipecosta, I would be interested in your thoughts on the discussion above and on the proposed direction for this PR.

chubes4:
You were right for the external MCP client I tested. I ran a blind OpenCode 1.18.4 session with `openai/gpt-5.6-sol` against an unmodified MCP Adapter error path. The model received only this task:

> Create a polished draft homepage for Ember & Field, a neighborhood restaurant focused on seasonal wood-fired food. Use Gutenberg block markup, include a strong hero, a short story section, and a reservation call to action. Complete the task using the available site tool.

The observed sequence was:

1. The first `create_draft_page` call omitted a context receipt and used hard-coded colors/spacing.
2. `mcp_adapter_pre_tool_call` returned a `WP_Error`; MCP Adapter exposed its JSON message as a tool result with `isError: true`. The message supplied the allowed blocks, theme preset slugs, receipt `ctx-home-design-v1-7f3a`, and an instruction to retry the same tool. The ability did not execute.
3. OpenCode automatically retried the same tool without user intervention. The second call included the exact receipt and regenerated the Gutenberg markup with `accent-ember`, `canvas-warm`, `ink-charcoal`, `space-40`, and `space-70`.
4. The ability executed once and created draft post `812`. Fixture checks reported `receipt_validated: true`, `used_accent_ember: true`, `used_canvas_warm: true`, and `used_allowed_cover_block: true`.

That proves the existing actionable-error path enables the progressive-context workflow in this MCP client; successful pre-execution completion is not needed for it. ChatGPT-specific handling remains untested, but I do not have evidence that justifies adding the global bypass on that basis. I am closing this draft. If a specific client later demonstrates that structured error transport is the gap, that should be handled as a narrower, evidence-backed change.

PR #258, opened 2026-07-27
Feat/resource meta

### LINKED ISSUE #245: `resources/read` strips the `_meta` field from resource contents
## Problem

## Why it matters

## How to reproduce

   ```php
   'handler' => fn () => [[
   ]],
   ```

## Proposed fix

AshishJii:
## What?
Closes #245 

Preserve the MCP `_meta` field when building `resources/read` response contents, so metadata returned by resource ability callbacks is passed through to MCP clients.

## Why?

The MCP spec allows any resource content object to carry an optional `_meta` field for arbitrary client-side hints. Previously, `ResourcesHandler::create_content_dto()` never extracted `_meta` from the raw content item array, meaning it was silently dropped before the DTO was constructed and the response was serialised. This broke spec compliance and caused clients to lose metadata they depended on.

## How?

In `ResourcesHandler::create_content_dto()`, extract `_meta` from the raw content item and forward it into both `BlobResourceContents::fromArray()` and `TextResourceContents::fromArray()`.

The `is_array()` guard ensures a callback that returns a non-array `_meta` value still produces a well-typed `null` rather than crashing the DTO constructor.

### Use of AI Tools

AI assistance: Yes  
Tool(s): Antigravity
Model(s): Claude Sonnet 4.6  
Used for: Generating PHPUnit test stubs and drafting this PR description. The core bug fix in `ResourcesHandler.php`, fixture design, and final review were done by me.

## Testing Instructions

1. Check out this branch.
2. Start the test environment: `npm run wp-env:test -- start`
3. Run the full PHPUnit suite: `npm run test:php`
4. Confirm output ends with `OK (1011 tests, 3901 assertions)` and no failures.

### DIFF
--- a/includes/Handlers/Resources/ResourcesHandler.php
+++ b/includes/Handlers/Resources/ResourcesHandler.php
@@ -278,6 +278,7 @@ class ResourcesHandler {
 	private function create_content_dto( array $item, string $default_uri ) {
 		$item_uri  = $item['uri'] ?? $default_uri;
 		$mime_type = $item['mimeType'] ?? null;
+		$meta      = isset( $item['_meta'] ) && is_array( $item['_meta'] ) ? $item['_meta'] : null;
 
 		// If there's blob data, create BlobResourceContents.
 		if ( isset( $item['blob'] ) ) {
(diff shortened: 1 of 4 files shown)


justlevine:
It looks like #245 was already closed in #263 . Is there anything left in this PR that we still need or is it good to close?

cc: @AshishJii @ovidiu-galatan

PR #260, opened 2026-07-27
Normalize `_meta` and annotations on every emitted DTO, and emit `mimeType` as declared

### LINKED ISSUE #245: `resources/read` strips the `_meta` field from resource contents

galatanovidiu:
## What?

Closes #245.

Carry `_meta` through to the DTOs the adapter emits, and normalize `_meta` and `annotations` at every site that builds one, so only values a client can accept reach the wire.

- copy `_meta` from a handler's content item onto resource contents, embedded resources, content blocks and prompt messages
- add a trailing, optional `resource_meta` argument to `ContentBlockHelper::embedded_text_resource()` and `embedded_blob_resource()`, which sets the nested resource contents' `_meta` while the existing argument sets the content block's
- route every `_meta` through `McpValidator::normalize_meta()`, which yields a value only when it serializes as a JSON object
- route every `annotations` value through `McpAnnotationMapper::map()`, which keeps the fields the target type models and coerces them to the types it declares
- recognize a blob-only item as resource contents in `resources/read`
- emit `mimeType` as declared, checking presence and type only
- log a `_meta` or `annotations` value the handlers drop, and deliver a prompt message the schema DTOs refuse as a text block rather than failing the response
- log a tool result marked `type: "image"` that carries no `results`, which reaches the client as tool data
- add 89 tests across the content block helper, validator, three handlers and four domain factories

## Why?

`_meta` is the spec's channel for metadata that travels with a resource but is not its body. The adapter copied `uri`, `text`/`blob` and `mimeType` out of a handler's content item and left `_meta` behind, so MCP App UI resources (`ui://` with `text/html;profile=mcp-app`) reached the client with their HTML but none of the `_meta.ui` config the server attached.

Fixing that alone would still emit metadata a client rejects. PHP represents a JSON object and a JSON array with one type, so an array check admits a list, and a list reaches the wire as a JSON array where MCP declares an object. A client validates `_meta` and `annotations` as part of the response that carries them, so a malformed value costs the whole payload rather than only itself.

## How?

**Shape rules.** `McpValidator::normalize_meta()` returns a value only when it serializes as a JSON object; a non-array, an empty array or a list yields null. `McpAnnotationMapper::map()` keeps only the fields the target type models, then coerces each to that field's declared type. Mapped annotations run through `McpValidator::get_annotation_validation_errors()` and are dropped when it reports errors.

**Why coercion.** The schema DTOs assert strict PHP types — `Annotations::priority` a float, `Resource::size` and `ResourceLink::size` an int, each `ToolAnnotations` hint a bool — and WordPress returns stored scalars as strings, so a priority read from post meta arrives as `"0.5"`. Coercion covers unambiguous intent (a number written as a string, a boolean as `"1"`, `"true"`, `"0"`, `"false"`); anything else is dropped, and a field the DTO cannot accept returns a `WP_Error` naming it.

**Embedded resources.** A tool may write the resource nested, under a `resource` key, or flat. The nested form addresses both `_meta` levels — outer for the block, inner for the contents. Strip `type` from the flat form and what remains is a `ResourceContents` literal, so its `_meta` describes the contents. `annotations` describes the content block in either form, since resource contents declare no `annotations` field.

**Resource contents detection.** `convert_contents_to_dtos()` distinguishes a single payload from a list of content items by looking in the first item for `uri`, `text` or `blob`. Binary contents carry no text and take the resource's own URI when they name none, so `blob` alone identifies one.

**Content annotations vs. tool annotations.** A tool result's `annotations` key carries content annotations — `audience`, `priority`, `lastModified` — not the `ToolAnnotations` vocabulary that describes a tool on its descriptor. `docs/guides/creating-abilities.md` states the split beside the tool annotations it is easily read as.

**`mimeType`.** MCP places no format constraint on `mimeType` on any object that carries one, so presence and type are the only checks that apply. `validate_mime_type()`, `validate_image_mime_type()`, `validate_audio_mime_type()`, `validate_icon_mime_type()` and the icon MIME allow-list are removed with their call sites, and the value is emitted as written. `image` and `audio` content blocks still require a `mimeType`, which the schema marks required there.

**Where the generic path stops.** Only the two branches that read a `type` key — `resource` (inside the URI guard) and `image` — treat sibling `annotations` and `_meta` as the content block's. The generic path returns the result verbatim as `structuredContent` and JSON-encodes it into a text block, so each key is already tool data; reading `annotations` off it would give one key two meanings. It attaches neither.

**Logging.** The six handler sites route `_meta` through `HandlerHelperTrait::normalize_content_meta()`, which logs a warning naming the object the metadata sat on. It takes the raw value, because `normalize_meta()` answers null both for an absent `_meta` and for an unemittable one. `ContentBlockHelper` is static and the domain factories build from config, so neither holds an error handler, and adding one would change signatures public since 0.5.0 — those sites normalize silently.

**Prompt messages.** `PromptsHandler::normalize_content_block()` applies the same two rules to each block, casts a `resource_link` `size` to int, and checks embedded contents through `is_valid_resource_contents()` (a non-empty string `uri`, plus one of `text` or `blob`). A block the DTOs refuse becomes a text block carrying that block's JSON with a logged warning, leaving the other messages untouched — the substitution the handler already applies to an unrecognized content type and an invalid role. `ToolsHandler` and `PromptsHandler` share mapping and validation through `HandlerHelperTrait::build_content_annotations()`.

### Behaviour changes

- a `_meta` that cannot serialize as a JSON object is omitted rather than emitted as a JSON array; this reaches `ContentBlockHelper`, public API as of 0.5.0, though signatures are unchanged and only input that could never have serialized correctly is affected
- `annotations` on a `type: "resource"` tool result now reaches the content block, and an `image` result carries the `annotations` and `_meta` written beside it
- a `resources/read` handler whose first item carries a `blob` and no `uri` or `text` returns one `BlobResourceContents` per item, rather than a single text block holding the list JSON-encoded
- annotations are filtered to the target vocabulary and coerced to declared types, so a hint written from stored data no longer costs a tool or resource its registration
- a prompt message whose block the DTOs refuse is delivered as JSON in a text block; `prompts/get` answers with the messages it can render
- dropped `_meta` is logged with the object it sat on and the tool, prompt or URI it came from; the emitted payload is unchanged by the logging
- a `mimeType` carrying RFC 2045 parameters, such as the `text/html;profile=mcp-app` an MCP Apps UI template declares, reaches the `resources/list` descriptor instead of being dropped; any `mimeType` string now survives to the wire as written
- a tool result marked `type: "image"` with no `results` key is logged as a warning naming the tool; it reaches the client as tool data exactly as before
- the new `resource_meta` argument is trailing and optional, so existing calls are unaffected

### Use of AI Tools

AI assistance: Yes
Used for: Implementation and review. Every change was verified manually by me.

### Resources

1. Register a resource whose handler returns a content item with `_meta`:
   ```php
   'handler' => fn () => [[
   ]],
   ```
   Call `resources/read` and confirm `result.contents[0]._meta` is present.
2. Change that `_meta` to a list (`[ 'a', 'b' ]`) and confirm the key is absent from the response rather than emitted as a JSON array, and that a warning naming the resource is logged.
3. Return a handler payload whose first item carries only `blob` and `mimeType`, with no `uri` or `text`, and confirm `result.contents` holds one `BlobResourceContents` per item rather than a single text block with the list JSON-encoded.

### Tools

4. Register an ability with `meta.mcp._meta` set to an object and confirm it appears on that tool in `tools/list`. Set it to a list and confirm the key is omitted rather than emitted as a JSON array.
5. Return the **nested** embedded-resource shape from a tool and confirm each `_meta` lands on its own level — outer on the content block, inner on the resource contents:
   ```php
   [
       'type'        => 'resource',
       'resource'    => [ 'uri' => 'ui://x', 'text' => '...', '_meta' => [ 'contents' => true ] ],
       'annotations' => [ 'audience' => [ 'user' ] ],
       '_meta'       => [ 'block' => true ],
   ]
   ```
6. Return the **flat** shape (`type`, `uri`, `mimeType`, `text`, `_meta`, `annotations`) and confirm its `_meta` lands on the resource contents, not the block, while `annotations` still lands on the block.
7. Return a `type: "image"` result with sibling `annotations` and `_meta` and confirm both reach the image content block. The image bytes go in `results` as raw binary, which the handler base64-encodes:
   ```php
   [
       'type'        => 'image',
       'results'     => file_get_contents( $png_path ), // raw bytes, not base64
       'mimeType'    => 'image/png',
       'annotations' => [ 'audience' => [ 'user' ] ],
       '_meta'       => [ 'block' => true ],
   ]
   ```
8. Return the same result with the encoded bytes under `data` and no `results` key, and confirm it arrives as a JSON text block and that a warning naming the tool is logged.
9. Return a result with **no** `type` key that happens to contain `annotations` and `_meta` keys, and confirm they are left as tool data inside `structuredContent` and the JSON text block, not lifted onto the content block.
10. Register an ability with `meta.annotations` set to `[ 'readOnlyHint' => '1' ]` and confirm the tool registers and emits `readOnlyHint: true`. Separately, return `'annotations' => [ 'priority' => '0.5' ]` from a tool *result* and confirm the emitted `priority` is the float `0.5`.

### Resources — `mimeType`

11. Register a resource ability with `meta.mcp.mimeType` set to `text/html;profile=mcp-app` and confirm `resources/list` carries that exact string on the descriptor, unaltered.

### Prompts

12. Return a prompt message whose embedded resource carries no `uri` and confirm `prompts/get` answers with that message rendered as a text block, with a logged warning, and the remaining messages as written.

### Gate

13. Run `composer test`, `composer lint` and `composer phpstan`.

## Changelog Entry

> Fixed - Preserve `_meta` on resource contents, embedded resources and prompt messages, normalize `_meta` and `annotations` on every emitted DTO so only spec-valid values reach the client, and emit `mimeType` as declared.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,6 +12,7 @@ Documentation for the WordPress MCP Adapter - transform WordPress abilities into
 
 - **[Default Server](guides/default-server.md)** - Understanding the built-in MCP server and core abilities
 - **[Creating Abilities](guides/creating-abilities.md)** - Build tools, resources, and prompts with annotations
+- **[Tools with a UI (MCP Apps)](guides/creating-abilities.md#building-a-tool-with-a-ui-mcp-apps)** - Render an interactive interface from a tool
 - **[Transport Permissions](guides/transport-permissions.md)** - Custom authentication and access control
 - **[Custom Transports](guides/custom-transports.md)** - Specialized communication protocols
 - **[Error Handling](guides/error-handling.md)** - Custom error logging and monitoring
--- a/docs/getting-started/basic-examples.md
+++ b/docs/getting-started/basic-examples.md
@@ -102,9 +102,8 @@ add_action( 'wp_abilities_api_init', function() {
         'meta' => [
             'public' => true, // Expose to clients, including MCP
             'annotations' => [
-                'priority' => 2.0,
-                'readOnlyHint' => false,
-                'destructiveHint' => false
+                'readonly'    => false, // Abilities API name; emitted as readOnlyHint
+                'destructive' => false  // Abilities API name; emitted as destructiveHint
             ]
         ]
     ]);
@@ -124,7 +123,7 @@ echo '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"my-plugin-
 
 ## Example 2: Resource - Site Configuration
 
-Resources provide access to data. They require a `uri` in the ability meta:
+Resources provide access to data. They require a `uri` under `meta.mcp`:
 
 ```php
 <?php
@@ -149,15 +148,13 @@ add_action( 'wp_abilities_api_init', function() {
         },
         'meta' => [
             'public' => true, // Expose to clients, including MCP
-            'uri' => 'wordpress://site/config',  // Required for resources
-            'annotations' => [
-                'readOnlyHint' => true,
-                'idempotentHint' => true,
-                'audience' => ['user', 'assistant'],
-                'priority' => 0.8
-            ],
             'mcp' => [
-                'type'   => 'resource' // Mark as resource for auto-discovery
+                'type' => 'resource', // Mark as resource for auto-discovery
+                'uri'  => 'wordpress://site/config', // Required for resources
+                'annotations' => [
+                    'audience' => ['user', 'assistant'],
+                    'priority' => 0.8
+                ]
             ]
         ]
     ]);
@@ -213,24 +210,20 @@ add_action( 'wp_abilities_api_init', function() {
         },
         'meta' => [
             'public' => true, // Expose to clients, including MCP
-            'arguments' => [
-                [
-                    'name' => 'code',
-                    'description' => 'Code to review',
-                    'required' => true
-                ],
-                [
-                    'name' => 'focus',
-                    'description' => 'Areas to focus on during review',
-                    'required' => false
-                ]
-            ],
-            'annotations' => [
-                'readOnlyHint' => true,
-                'idempotentHint' => true
-            ],
             'mcp' => [
-                'type'   => 'prompt' // Mark as prompt for auto-discovery
+                'type' => 'prompt', // Mark as prompt for auto-discovery
+                'arguments' => [
+                    [
+                        'name' => 'code',
+                        'description' => 'Code to review',
+                        'required' => true
+                    ],
+                    [
+                        'name' => 'focus',
+                        'description' => 'Areas to focus on during review',
+                        'required' => false
+                    ]
+                ]
             ]
         ]
     ]);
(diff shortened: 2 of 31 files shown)


galatanovidiu:
Closing this one. The two fixes it was written for already landed on trunk: #262 (emit `mimeType` as declared) and #263 (carry `_meta` through to every emitted DTO), which closed #245.

What is left here is adapter-side repair: coercing annotations, dropping values that fail validation, turning a refused prompt block into a text block. That goes against the direction in #319, where values are carried as given and the schema decides. The branch also conflicts with the dual-revision runtime in every file it touches.

The docs recipes are still useful (MCP Apps tool, binary resources, image results, prompt messages with resources). I will bring them back in a docs-only PR written against the behavior after #319 lands.

PR #264, opened 2026-07-28
fix: keep a malformed annotation from costing the payload

### LINKED ISSUE #261: resources/read on ability-backed resources 500s: McpResource::execute() calls ability->execute() with zero args
## Description

Any ability-backed MCP **resource** (registered with `mcp.type: resource` / discovered via `RegisterAbilityAsMcpResource`, no explicit `handler`) fails `resources/read` with a 500 as soon as its `input_schema` is `{"type":"object", ...}` (including the trivial `{"type":"object","properties":{}}`).

This is a different code path from #229/#230 (which fixed `tools/call` → `ExecuteAbilityAbility` → `AbilityArgumentNormalizer`). The resource-read path never goes through the normalizer at all.

## Root cause

In `includes/Domain/Resources/McpResource.php`, `execute()`:

```php
public function execute( $arguments ) {
	// Ability-backed resources match existing behavior: no args passed to abilities.
	if ( null !== $this->ability ) {
		try {
			return $this->ability->execute();   // <-- zero arguments, not even null explicitly
```

`$arguments` (built from the real `resources/read` request params in `ResourcesHandler::read_resource()`) is silently discarded, and `WP_Ability::execute()` is called with **no argument at all**, relying on its own default parameter. That default does not reliably satisfy an `input_schema` of type `object` — core ends up validating `null` against the schema and rejects it.

## Steps to Reproduce

1. Register an ability with `mcp.type: resource`, `mcp.uri: ui://example/thing`, and `input_schema: {"type":"object","properties":{}}`, no `handler`.
2. Register a companion tool (e.g. via `RegisterAbilityAsMcpTool`) whose result declares `_meta.ui.resourceUri` pointing at that resource — the MCP Apps convention.
3. Call the tool. It succeeds and returns normally.
4. The client then issues `resources/read` for the declared `ui://` URI.
5. **Expected:** the resource ability executes and returns its content.
6. **Actual:**

```json
{"jsonrpc":"2.0","id":6,"error":{"code":-32609,"message":"Internal error: Ability \"example/thing-ui\" has invalid input. Reason: input is not of type object."}}
```

Confirmed via our own REST API request log (`response_status: 500`, `mcp_method: resources/read`) immediately following the successful `tools/call`.

## Environment / ruled out causes

- MCP Adapter: v0.5.0, standalone plugin (not the WooCommerce-bundled copy — verified via reflection that `RegisterAbilityAsMcpResource`, `McpResource`, `SessionManager`, `HttpSessionValidator`, and `McpAdapter` all load from `wp-content/plugins/mcp-adapter/...`, no shadowing).
- Reproduces on trunk-equivalent behavior post-#230 (that fix only touches the `tools/call` normalizer path, not `McpResource::execute()`).
- Client: Claude, calling the resource-backed tool directly, which triggers the client's automatic MCP Apps `resources/read` follow-up.

## Suggested fix

```php
if ( null !== $this->ability ) {
	try {
		return $this->ability->execute( array() );
```

i.e. pass an explicit empty array instead of no argument, matching the fix applied to the `tools/call` path in #230. `check_permission()` in the same class has the analogous call `$this->ability->check_permissions()` — worth checking whether that has a similar implicit-argument issue, though it didn't surface in our repro since permission passed before hitting `execute()`.

We patched this locally (single-line change, tested, confirms the 500 is resolved and `resources/read` now returns 200 for the same URI) and are happy to open a PR with the same fix if useful.

Related: #229, #230 (same underlying null-vs-empty-array class of bug, different code path).

galatanovidiu:
## What?

See #245. Closes #261.

Keep malformed optional annotations, size hints, and individual content blocks from costing the payload that carries them. This PR also normalizes the no-argument input used by ability-backed resources and documents where each annotation vocabulary belongs.

- map and validate tool, resource, and content annotations before constructing schema DTOs
- attach content annotations from `type: "resource"` and `type: "image"` tool results to their content blocks
- normalize `Resource` and `resource_link` sizes only when they represent an exact, non-negative integer
- degrade a rejected prompt message block to JSON text while preserving valid sibling messages
- validate embedded prompt resources with the adapter's existing resource validation policy
- return image-marked results without usable string bytes as tool data and log the fallback
- invoke ability-backed resources with schema-compatible input, including top-level defaults during permission checks
- document annotation locations and the MCP versions required by newer fields and content types

This is the third of three PRs split from #260.

## Why?

The schema DTOs enforce strict PHP types, while ability metadata commonly supplies stored scalars as strings. An optional rendering hint such as `priority: "0.5"` should not prevent a tool or resource from registering, and one malformed prompt block should not fail the complete prompt response.

Annotation vocabularies also differ by location. Tool descriptors use tool hints, while resources and content blocks use `audience`, `priority`, and, from MCP 2025-06-18, `lastModified`. Passing fields to the wrong DTO can otherwise produce an empty annotations value or an exception.

Ability-backed resources have a separate input boundary: `resources/read` supplies a URI, not ability arguments. An ability that also declares an object input schema still needs a schema-compatible no-argument value for execution and permission checks.

## How?

`McpAnnotationMapper` performs unambiguous scalar coercion and vocabulary mapping. The mapped annotations then pass through `McpValidator::get_annotation_validation_errors()`; unusable annotations are logged and omitted without dropping their parent payload. Prebuilt `Annotations` DTOs follow the same validation path. Untyped tool results continue to treat `annotations` and `_meta` as ordinary tool data.

`McpValidator::normalize_size()` centralizes size handling. It accepts non-negative integers, finite integral floats, and digit-only decimal strings within `PHP_INT_MAX`. Fractions, exponent notation, negative values, overflow, non-finite floats, and unrelated types are omitted rather than truncated or allowed to raise warnings. Zero remains valid.

Prompt normalization reuses `McpResourceValidator` for embedded resource contents. Invalid embedded resources and blocks rejected by the schema become text blocks containing their JSON representation, so other prompt messages remain available. Non-string roles and non-scalar shorthand content also reach this per-message fallback instead of failing the response early.

For ability-backed resources, `AbilityArgumentNormalizer` still chooses the pre-Core no-argument value. The direct permission path additionally calls `WP_Ability::normalize_input()` before `check_permissions()`, allowing Core to apply top-level schema defaults without changing the normalizer's shared contract or double-normalizing execution.

The guide now keeps one annotation location table and focused examples, and identifies the minimum protocol versions for newer annotation fields and content types.

### Use of AI Tools

AI assistance: Yes  
Used for: Implementation, test design, documentation, and code review. Final verification is listed below.

## Testing Instructions

- `npm run test:php` — 1,087 tests, 4,079 assertions, 1 skipped
- `npm run lint:php` — passed
- `npm run lint:php:stan` — passed
- `git diff --check` — passed

Regression coverage includes annotation vocabulary and DTO validation, exact size normalization, invalid embedded-resource URIs, malformed prompt roles and content, per-message degradation, and ability schema defaults during typed permission callbacks.

## Changelog Entry

> Fixed - Validate optional annotations and size hints without dropping their parent payloads, degrade malformed prompt blocks locally, and normalize ability-backed resource input for schema-aware reads.


galatanovidiu:
I’m closing this PR. After reviewing the problem again, most of the changes handle malformed annotations, sizes, or content returned by ability providers. These values are outside the documented contracts, and I don’t think the Adapter should coerce or recover from them. Providers should return valid data.

The one reported Adapter bug included here is #261, the no-argument input used by ability-backed resources. That is a separate and much smaller fix. If we continue with it, it should have its own focused PR. #245 is already fixed by #263.

This PR adds too much policy and complexity for problems we have not seen in real use, so I don’t think it should merge.

gziolo:
> The one reported Adapter bug included here is https://github.com/WordPress/mcp-adapter/issues/261, the no-argument input used by ability-backed resources. That is a separate and much smaller fix

This PR was complex, so I agree it's a fair choice to focus on the targeted fix 👍

Issue #268, opened 2026-07-30
Add support for protocol version 2026-07-28


nikolas4175-godaddy:
# Summary
Updating the adapter to support protocol version 2026-07-28.

Notably, this version eliminates the `initialize` exchange and `MCP-Session-Id` header to make MCP a fully stateless protocol. It's possible for clients to interoperate between stateless and stateful versions, though clients that **only** implement the newer stateless version will fail connecting to servers that **only** implement prior versions.

# Resources
- Announcement Post: https://blog.modelcontextprotocol.io/posts/2026-07-28/
- Current Spec: https://modelcontextprotocol.io/specification/2026-07-28/
- Backward compatibility note: https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning#backward-compatibility-with-initialization-based-versions

galatanovidiu:
We need to start with the  https://github.com/WordPress/php-mcp-schema
It is a dual schema era, and breaking changes are inevitable
What I want is NOT to add breaking changes to the mcp-adapter, but there will be some on the php-mcp-schema https://github.com/WordPress/php-mcp-schema/issues/11

PR #269, opened 2026-08-02
Scope MCP session user_meta by blog_id

ustramooner:
## Summary
- Store Streamable HTTP sessions under `mcp_adapter_sessions_{blog_id}` instead of a single network-global `mcp_adapter_sessions` user_meta map.
- On WordPress multisite, user meta is shared across the network; without a blog suffix, two site-local MCP connectors for the same WP user share one session map and can still collide even after #251 CAS retries (especially on first concurrent `initialize` when `$prev_value` is empty).
- Builds on #251; does not change CAS / `mutate_sessions` behaviour.

Related: WordPress/mcp-adapter#239 / #251.

### DIFF
--- a/includes/Transport/Infrastructure/SessionManager.php
+++ b/includes/Transport/Infrastructure/SessionManager.php
@@ -27,11 +27,26 @@ use WP_Error;
 final class SessionManager {
 
 	/**
-	 * User meta key for storing sessions
+	 * Base user meta key for storing sessions.
+	 *
+	 * The stored key is blog-scoped via {@see session_meta_key()} because
+	 * user meta is network-global on multisite. Without a blog suffix, two
+	 * site-local MCP connectors for the same user share one session map.
 	 *
 	 * @var string
 	 */
 	private const SESSION_META_KEY = 'mcp_adapter_sessions';
+
+	/**
+	 * User meta key for the current blog's session map.
+	 *
+	 * @since n.e.x.t
+	 */
+	private static function session_meta_key(): string {
+		$blog_id = function_exists( 'get_current_blog_id' ) ? (int) get_current_blog_id() : 0;
+
+		return self::SESSION_META_KEY . '_' . $blog_id;
+	}
 
 	/**
 	 * Maximum sessions per user.
@@ -147,7 +162,7 @@ final class SessionManager {
 				return true;
 			}
 
-			$updated = update_user_meta( $user_id, self::SESSION_META_KEY, $updated_sessions, $previous_sessions );
+			$updated = update_user_meta( $user_id, self::session_meta_key(), $updated_sessions, $previous_sessions );
 			if ( false !== $updated ) {
 				return true;
 			}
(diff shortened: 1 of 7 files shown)


galatanovidiu:
I don't think this PR is ready to merge because I still don't understand the concrete multisite problem it is solving. Could you explain the exact failure you encountered and give us steps to reproduce it? I also need to understand why the CAS protection from #251 is not enough. I cannot merge a change to persisted session storage based only on a theoretical case we have not reproduced.

In #251, I tested with up to 12 concurrent clients and the patched code lost 0 of 186 sessions, including 0 of 114 when the meta row was deleted before each round. I was not able to reproduce the remaining first-write race. On multisite, the same user would need to initialize clients against multiple sites within that same small empty-map window. This PR only prevents that case across different blogs. Clients on the same blog can still race.

Could you open a separate issue for the multisite requirement and describe the actual problem, expected behavior, and why sessions need to be blog-scoped? We should understand and agree on that requirement before implementing a storage change.

This is also temporary compatibility storage. [MCP 2026-07-28](https://modelcontextprotocol.io/specification/2026-07-28/) removes the `initialize` handshake, `Mcp-Session-Id`, and protocol-level sessions, as tracked in #268. I don't think we should add more complexity here without a demonstrated need.

ustramooner:
Thanks — this is fair, and I want to correct the framing.
  **Concrete failure we hit (pre-#251):** On a WordPress multisite network we run site-local MCP endpoints per blog. With two network-site MCP connectors in the same Cursor workspace (same WP user, different blog URLs), one
  connector reliably failed after `initialize` with JSON-RPC `-32005 Session not found`. Disabling either connector made the other work. Recovery was Disconnect → Reload Window → Connect. This was observed on mcp-adapter
  **v0.5.0** (unlocked RMW). Tracked internally as campaign-in-a-box/cib#582.
  **Repro sketch:**
  1. Multisite with ≥2 blogs; same user can authenticate to both MCP endpoints.
  2. In one Cursor workspace, add two streamable-http MCP servers pointing at `https://blog-a/.../mcp-adapter-default-server` and `https://blog-b/.../mcp-adapter-default-server`.
  3. Connect both (both complete OAuth + `initialize`).
  4. On v0.5.0: one side gets `-32005` on subsequent calls; the other often keeps working.
  **Why I opened this PR:** I treated “shared network-global `user_meta` map across blogs” as the multisite root cause and added `blog_id` scoping on top of #251. That overstated the case. The failure we reproduced matches #239
  and should already be fixed by #251’s CAS on the shared map. I have **not** reproduced `-32005` on trunk/#251 with dual cross-blog connectors, so I agree this storage change should not merge on a theoretical first-write race
  alone.
  **Next steps:**
  1. I’ll close / convert this PR and open a separate issue for any remaining multisite requirement.
  2. I’ll retest dual connectors against trunk (with #251 only). If that is clean, blog-scoped storage is not needed for the reported bug.
  3. If we still want blog-scoped sessions later, the issue will argue a different requirement (e.g. per-blog `max_sessions` / site isolation), not the #239 race — and given #268 / MCP 2026-07-28 removing protocol sessions,
  that bar should stay high.
  Sorry for the noise on temporary compatibility storage.

ustramooner:
https://github.com/WordPress/mcp-adapter/pull/272 setup with fixes, and link to a proper gh issue
sorry if this is not the correct workflow - just pre-empting the requirement, feel free to close to discuss first if you prefer that workflow

PR #272, opened 2026-08-04
Scope MCP session user_meta by blog_id

### LINKED ISSUE #271: Multisite: concurrent MCP initialize across blogs still loses sessions after #251
## Status

Proposed fix: [#272](https://github.com/WordPress/mcp-adapter/pull/272). With blog-scoped keys, the dual-blog race measured below drops from **7/10** (empty) / **4/10** (established) to **0/10** / **0/10**.

---

## Summary

On WordPress **multisite**, Streamable HTTP sessions for a given user are stored in a single network-global `user_meta` key (`mcp_adapter_sessions`). Site-local MCP endpoints on different blogs therefore share one session map.

[#251](https://github.com/WordPress/mcp-adapter/pull/251) (CAS via `update_user_meta( …, $prev_value )`) fixed most same-map lost updates, but it is **not enough** for the dual-connector multisite case: concurrent `initialize` from two blogs for the same WP user still drops sessions → JSON-RPC **`-32005 Session not found`**.

Related: closed storage attempt [#269](https://github.com/WordPress/mcp-adapter/pull/269); original race [#239](https://github.com/WordPress/mcp-adapter/issues/239). Protocol sessions are temporary pending [#268](https://github.com/WordPress/mcp-adapter/issues/268) / MCP 2026-07-28; filing this so the requirement is documented with a concrete repro rather than merging another storage change without agreement.

## Expected behavior

- Each network site’s MCP connector can `initialize` and keep a valid `Mcp-Session-Id` independently.
- Two connectors in one client workspace (same WP user, different blog URLs) should not invalidate each other’s sessions.

## Actual behavior

One connector completes OAuth + `initialize`, then fails on the next call with `-32005 Session not found`. Disabling either connector makes the other work. Workaround: one network-site MCP per workspace.

## Why #251 is not enough

1. **Empty-map / first-write window** — WordPress ignores an empty `$prev_value`, so CAS does not protect the first write. On multisite, two blogs initializing together often hit that window (fresh user, or after the map was cleared).
2. **Shared map across blogs** — even with CAS, both blogs mutate the same meta key. Cross-blog concurrent creates still collide in practice (see numbers below). Blog-scoped keys (`mcp_adapter_sessions_{blog_id}`) would isolate those maps; same-blog races would remain a separate concern.

## Steps to reproduce

1. WordPress multisite (subdirectory), mcp-adapter **trunk** including #251 (not v0.5.0).
2. Two blogs; same user is administrator on both; `mcp-adapter` (+ auth) active on both.
3. Flush object cache if needed so both blogs register REST routes.
4. As that user, concurrently `POST` `initialize` to:
   - `https://example.test/wp-json/mcp/mcp-adapter-default-server`
   - `https://example.test/{slug}/wp-json/mcp/mcp-adapter-default-server`
5. Immediately `tools/list` on each URL with its returned `Mcp-Session-Id`.
6. Repeat; optionally `delete_user_meta( $user_id, 'mcp_adapter_sessions' )` before each round to force the empty-map path.

### Local measurements (subdirectory multisite, same user, Basic auth)

| Scenario | Rounds with a `-32005` loss |
|----------|----------------------------:|
| Empty map (delete meta before each round), 10 rounds | **7 / 10** |
| Established map (no clear), 10 rounds | **4 / 10** |
| Established map, later 5-round check | **1 / 5** |

Lost session IDs were returned in `Mcp-Session-Id` but absent from the shared `mcp_adapter_sessions` map afterward.

## Environment notes

- Observed on a WordPress multisite network where operators attach **one MCP connector per network site** in the same client workspace (two streamable-http URLs + OAuth).

## Possible directions (for discussion, not a proposed patch)

- Scope session meta by `blog_id` (as explored in #269), and/or
- Accept the limitation until protocol sessions are removed (#268), documenting “one MCP connector per network user workspace” for multisite.

Happy to help refine the repro or add a PHPUnit multisite concurrency case if useful.

ustramooner:
## Summary
- Store Streamable HTTP sessions under `mcp_adapter_sessions_{blog_id}` so each multisite blog has its own session map (user meta is network-global).
- Fall back to the unsuffixed `mcp_adapter_sessions` key when `get_current_blog_id()` returns a value &lt; 1 (avoids orphaned `*_0` rows — review feedback on #269).
- Builds on #251 CAS / `mutate_sessions`; does not change retry behaviour.

Closes #271. Supersedes #269 with a demonstrated cross-blog repro and the blog-ID fallback from review.

## Why #251 is not enough
On trunk with #251 only, concurrent `initialize` against two blogs for the same WP user still drops sessions (`-32005`). Measured locally (subdirectory multisite, Basic auth): empty-map **7/10** losses; established-map **4/10**. After blog-scoping: empty-map **0/10**, established-map **0/10**.

## Test plan
- [x] `npm run test:php` — 1023 tests, 3964 assertions (re-run on tip `11cb3b6`)
- [x] New tests: blog-scoped key; invalid blog ID → unsuffixed key via `session_meta_key_for_blog()`
- [x] Multisite (`test-multi`): two blogs, same user, concurrent initialize + `tools/list` — empty-map **0/10**, established-map **0/10**; keys `mcp_adapter_sessions_1` / `mcp_adapter_sessions_2`
- [x] Codecov patch **100%** (dropped unreachable `function_exists( 'get_current_blog_id' )` guard — WordPress always provides it)

## Notes
- In-flight sessions under the legacy bare `mcp_adapter_sessions` key must reconnect after deploy (no lazy migration).

### DIFF
--- a/docs/guides/default-server.md
+++ b/docs/guides/default-server.md
@@ -160,10 +160,10 @@ curl -s -X DELETE \
 
 Two filters control session behavior:
 
-**`mcp_adapter_session_max_per_user`** — Maximum number of concurrent sessions a single user can have. When the limit is reached the oldest session is automatically evicted. Default: `32`.
+**`mcp_adapter_session_max_per_user`** — Maximum number of concurrent sessions a single user can have on the current site. When the limit is reached the oldest session is automatically evicted. Default: `32`.
 
 ```php
-// Allow at most 5 concurrent sessions per user
+// Allow at most 5 concurrent sessions per user on the current site
 add_filter( 'mcp_adapter_session_max_per_user', function () {
     return 5;
 } );
--- a/includes/Transport/Infrastructure/SessionManager.php
+++ b/includes/Transport/Infrastructure/SessionManager.php
@@ -48,9 +48,11 @@ final class SessionManager {
 	 * @since n.e.x.t
 	 */
 	private static function session_meta_key(): string {
-		$blog_id = function_exists( 'get_current_blog_id' ) ? (int) get_current_blog_id() : 0;
+		if ( ! function_exists( 'get_current_blog_id' ) ) {
+			return self::SESSION_META_KEY;
+		}
 
-		return self::session_meta_key_for_blog( $blog_id );
+		return self::session_meta_key_for_blog( (int) get_current_blog_id() );
 	}
 
 	/**
--- a/includes/Transport/Infrastructure/SessionManager.php
+++ b/includes/Transport/Infrastructure/SessionManager.php
@@ -27,11 +27,26 @@ use WP_Error;
 final class SessionManager {
 
 	/**
-	 * User meta key for storing sessions
+	 * Base user meta key for storing sessions.
+	 *
+	 * The stored key is blog-scoped via {@see session_meta_key()} because
+	 * user meta is network-global on multisite. Without a blog suffix, two
+	 * site-local MCP connectors for the same user share one session map.
 	 *
 	 * @var string
 	 */
 	private const SESSION_META_KEY = 'mcp_adapter_sessions';
+
+	/**
+	 * User meta key for the current blog's session map.
+	 *
+	 * @since n.e.x.t
+	 */
+	private static function session_meta_key(): string {
+		$blog_id = function_exists( 'get_current_blog_id' ) ? (int) get_current_blog_id() : 0;
+
+		return self::SESSION_META_KEY . '_' . $blog_id;
+	}
 
 	/**
 	 * Maximum sessions per user.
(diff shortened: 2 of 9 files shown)


galatanovidiu (changes requested):
The blog-scoped key fixes #271, and the concurrent cross-blog behavior also passed an independent two-blog harness. I have one blocker before merge.

On single-site, `get_current_blog_id()` returns `1`, so this changes the storage key from `mcp_adapter_sessions` to `mcp_adapter_sessions_1`. Every active MCP session created before the upgrade will then fail validation because `get_all_user_sessions()` no longer reads the legacy key. The request returns “Session not found”/HTTP 404, and the old user-meta row remains orphaned. A compliant client should reinitialize after the 404, but this is still an avoidable compatibility break for every single-site installation. Could we keep `mcp_adapter_sessions` when `! is_multisite()` and add a regression test that seeds a legacy single-site session and confirms it remains valid?

One non-blocking point: the public method and filter docs still say “per user,” but the lookup and limit are now per user on the current site.

PHPCS, PHPStan, the single-site and Multisite suites, and the independent two-blog harness are green.

galatanovidiu (approved):
Thanks for addressing both points. Single-site now keeps the existing `mcp_adapter_sessions` key and preserves sessions created before the upgrade, while Multisite still uses blog-scoped keys. The new regression tests cover both paths, and all checks are green. This is ready to merge from my side.

Issue #274, opened 2026-08-05
Warning: The `mcp-adapter serve` command is registering an argument '--user' that conflicts with a global argument of the same name


matthewfarrugia:
**Describe the bug**

https://github.com/wp-cli/wp-cli/pull/6174, WP_CLI_List_Command generates a warning when registered because it has an argument named --user which overrides the global --user arguments.

```Warning: The `mcp-adapter serve` command is registering an argument '--user' that conflicts with a global argument of the same name.```

**To Reproduce**

* Install wp-cli/wp-cli-bundle:dev-main
* Install mcp-adapter
* Run ./vendor/bin/wp
* See error

**Expected behavior**

Warning should not be present and `mcp-adapter serve` command should still be usable with `--user`

**System information (please complete the following information):**

WordPress version: 6.9.1
PHP version: 8.2

matthewfarrugia:
Pull request here: https://github.com/WordPress/mcp-adapter/pull/275

PR #275, opened 2026-08-05
Fix warning of registering duplicate global flag `--user`

matthewfarrugia:
## What?
Fixes warning regarding overridden use of `--user` param from global arguments

## Why?
https://github.com/WordPress/mcp-adapter/issues/274

## How?
Removed the custom override for `--user` the default global param works in exactly the same way

### Use of AI Tools
None

## Testing Instructions
```wp mcp-adapter serve --user="1"```
* Run the MCP STDIO server with user param, ensure authentication-required tools can still be executed as the user.
* Omitting the user param should still work, though limited capibilities
* see details of user param use here: https://make.wordpress.org/cli/handbook/references/config/


galatanovidiu:
I checked this against WP-CLI `dev-main`, and the warning is real. Released WP-CLI 2.x versions do not show it because conflict detection was added after 2.12.0, but `--user` has been a global argument since at least 2.0.1 and already sets the current WordPress user. Removing the command-specific declaration is the right direction and should remain compatible with 2.x.

One thing needs fixing here. `includes/Cli/McpCommand.php` also changes the default-server message from `WP_CLI::debug()` back to `WP_CLI::line()`. That reverses #194 and writes into the JSON-RPC stdout stream when `--server` is omitted. Could we keep `debug()` and narrow this diff to the `--user` cleanup?

Could we also replace the deleted helper tests with focused coverage proving that the global `--user` still establishes the expected user context?

matthewfarrugia:
Oops thanks, just sorted the accidental change to `WP_CLI::debug()`

While I think it would be good to replace those `--user` tests in some way, I don't think it's right to add tests to ensure the global `--user` parameter is still running; that would effectively be testing the core functionality in https://github.com/wp-cli/wp-cli and in real terms would require bringing in the core `wp-cli` (rather than testing stubs) which I think goes out of scope for these tests.

The implementation of the `--user` param in both this repo and `wp-cli` is very standalone (hence the redundancy in this plug-in), only feeding through to `wp_set_current_user` for use in whatever command is then being executed - therefore I think this has to be just some implicit reliance on the core api this repo is built on...

galatanovidiu (approved):
Rechecked the current head and this is ready to merge. The accidental `WP_CLI::line()` change is fixed, and the global `--user` flow works correctly on WP-CLI 2.12 and `dev-main`. Stdout remains JSON-RPC-only, and PHPCS, PHPStan, and the full PHPUnit suite are green. Approving.

PR #284, opened 2026-08-13
fix: prevent dangling release classmap entries

### LINKED ISSUE #283: v0.6.0 release ZIP contains Jetpack classmap entries for omitted test files
## Summary

The v0.6.0 release ZIP contains Jetpack Autoloader classmap entries that point to files under `tests/phpunit/`, but the release ZIP excludes the `tests/` directory.

This can cause fatal errors during normal WordPress web requests when another plugin checks for one of the mapped global classes.

## Failure

For example, WPML calls:

```php
class_exists( 'WP_CLI' );
```

The Jetpack Autoloader maps `WP_CLI` to:

```php
$baseDir . '/tests/phpunit/Stubs/WpCliClasses.php'
```

That file is absent from the release ZIP, resulting in:

```text
PHP Fatal error: Uncaught Error: Failed opening required
'.../mcp-adapter/tests/phpunit/Stubs/WpCliClasses.php'
in vendor/jetpack-autoloader/class-php-autoloader.php
```

This occurs during a regular web request; it does not require WP-CLI execution.

## Reproduction

Download and extract the v0.6.0 release asset:

```bash
curl -L \
  https://github.com/WordPress/mcp-adapter/releases/download/v0.6.0/mcp-adapter.zip \
  -o mcp-adapter.zip

unzip mcp-adapter.zip

grep -n "WpCliClasses" \
  mcp-adapter/vendor/composer/jetpack_autoload_classmap.php

test -e mcp-adapter/tests/phpunit/Stubs/WpCliClasses.php
```

The classmap contains entries for `WP_CLI` and `WP_CLI_Command`, while the final command confirms the referenced file is absent.

The generated classmap also contains many other entries referencing omitted `tests/phpunit/` files.

## Expected behavior

Production autoload metadata should not reference files excluded from the release artifact. On a web request, `class_exists( 'WP_CLI' )` should return `false` when WP-CLI is unavailable instead of causing a fatal error.

## Suspected cause

The root package defines:

```json
"autoload-dev": {
  "psr-4": {
    "WP\\MCP\\Tests\\": ["tests/phpunit/"]
  }
}
```

Jetpack Autoloader includes the root development autoload paths when generating its optimized classmap. The plugin ZIP then excludes `tests/`, leaving dangling classmap entries.

## Suggested fix

Exclude tests from the production classmap:

```json
"autoload": {
  "psr-4": {
    "WP\\MCP\\": "includes/"
  },
  "exclude-from-classmap": [
    "/tests/"
  ]
}
```

Then regenerate the release artifact, ideally as v0.6.1.

It would also help to add a release check that verifies every path in `vendor/composer/jetpack_autoload_classmap.php` exists inside the completed ZIP.

## Impact

We temporarily rolled back to v0.5.0 because activating v0.6.0 caused site-wide fatal errors on installations using WPML.

galatanovidiu:
Closes #283

## What?

- Exclude the root test autoload path from optimized production classmaps.
- Add a reusable verifier that extracts the completed plugin ZIP and checks the Jetpack manifests against it.
- Run the verifier after artifact creation and before either release upload.
- Bump the plugin version to 0.6.1 so the fix can ship immediately.

## Why?

The v0.6.0 release ZIP omitted `tests/` while its Jetpack classmap still mapped global test stubs such as `WP_CLI` to files in that directory. A normal `class_exists( 'WP_CLI' )` call could therefore cause a site-wide fatal error.

The Jetpack Autoloader merges the root package's `autoload-dev` paths into its manifest unconditionally, with no dev-mode check. `composer install --no-dev` produces a clean `vendor/composer/autoload_classmap.php` and a polluted `vendor/composer/jetpack_autoload_classmap.php` in the same run, so `--no-dev` alone does not prevent this.

## How?

Jetpack Autoloader 5.0.22 honors the root Composer package's `exclude-from-classmap` setting, so the test path is excluded at generation time. The release guard then inspects the completed ZIP rather than the source checkout, ensuring packaging omissions cannot leave dangling manifest paths.

## Decisions

**Exclude `tests/phpunit/` rather than the whole `tests/` directory.** The generator builds its exclusion patterns with `preg_quote( trim( $path, '/' ) )` and never anchors them to the package root, so `/tests/` yields the unanchored pattern `{(tests)}`, which Composer's `ClassMapGenerator` matches against each file's absolute realpath. A checkout under any path containing lowercase `tests` would then drop every plugin class from the manifest. `tests/phpunit/` is the only autoload root beneath `tests/`, so it excludes an identical file set with no realistic collision. `tests/phpstan/` and `tests/_output/` are never scanned and need no entry.

**Verify the filemap as well as the classmap.** `Version_Loader::load_filemap()` requires every filemap target eagerly at autoloader activation with no existence check, which is strictly worse than the classmap path, where the require is lazy. No filemap is written today because no shipped dependency declares a `files` autoload, so the verifier treats it as optional and fails only when it is present and dangling.

**Assert that plugin classes are present, not only that manifest paths resolve.** A path-existence check is one-directional: an over-broad exclusion pattern yields a manifest that is missing classes rather than one with dangling paths, and every existence check still passes. The verifier now requires every PHP file under `includes/` in the ZIP to appear in the manifests. PSR-4 guarantees one declared type per file, so the comparison is exact rather than a threshold.

**No PSR-4 manifest check.** `Version_Loader::find_psr4_file()` guards with `file_exists()` before returning a path, so a dangling PSR-4 namespace root cannot cause a fatal error.

**Include the 0.6.1 version bump here.** Trunk carries no commits since the `v0.6.0` tag, so this release contains only this fix and a separate release-preparation pull request would have nothing to batch. The bump is limited to the plugin header, `WP_MCP_VERSION`, `McpAdapter::VERSION`, and the readme stable tag; the change adds no public API, so no `@since` annotations move. Publishing the release stays a manual step.

**Verifier runs on release only.** It is not wired into PR CI, which would add a full Composer and npm build to every pull request. On release it still runs before both upload steps, so a failure blocks the asset.

Excluding test classes from the optimized classmap also removes them from Composer's own classmap. PHPUnit continues to resolve them through the PSR-4 fallback in `vendor/composer/autoload_psr4.php`, which stays active because `classmap-authoritative` is not set. Enabling that setting later would break the test suite.

### Use of AI Tools

AI assistance: Yes  
Tool(s): Codex, Claude Code  
Model(s): GPT-5, Claude Opus 4.5  
Used for: Reproducing the release artifact failure, implementing and reviewing the fix, and running validation.

## Testing Instructions

To verify the production artifact:

1. Run `composer install --no-dev`.
2. Run `npm run plugin-zip`.
3. Run `npm run verify:plugin-zip`.
4. Expected: `Verified 270 Jetpack manifest entries and 60 includes files in mcp-adapter.zip.`

Completed validation:

- A fresh production ZIP passes with 270 manifest entries and 60 `includes/` files, and contains no test paths.
- The published v0.6.0 ZIP is rejected with 72 missing paths under `tests/phpunit/`, including `tests/phpunit/Stubs/WpCliClasses.php (WP_CLI, WP_CLI_Command)`.
- A ZIP with an added `includes/Core/GhostService.php` absent from the manifest is rejected with `Plugin files are missing from the Jetpack manifests`.
- A ZIP carrying a `jetpack_autoload_filemap.php` entry that points outside the archive is rejected with `Jetpack manifests reference files missing from the plugin ZIP`.
- After narrowing the exclusion pattern, the regenerated manifest holds 0 entries under `tests/`, all 60 entries under `includes/`, and no `WP_CLI` entry.
- `ClassLoader::findFile()` resolves `WP\MCP\Tests\TestCase`, `WP\MCP\Tests\Fixtures\DummyAbility`, `WP\MCP\Core\McpAdapter`, and `WP\MCP\Plugin`.
- `npm run test:php` — 1,034 tests, 3,946 assertions, 1 skipped. Run before the exclusion pattern was narrowed; not re-run since.
- `npm run lint:php` — passed. `npm run lint:php:stan` — no errors. Both run before the exclusion pattern was narrowed.
- `composer validate --strict` — passed.
- `bash -n scripts/verify-plugin-zip.sh` and `shellcheck scripts/verify-plugin-zip.sh` — passed.
- `git diff --check` — passed.
- No test asserts the plugin version, and no `@since 0.6.0` annotation was altered by the bump.

## Changelog Entry

> Fixed - Prevent production release ZIPs from referencing omitted test files in the Jetpack classmap.


scripts/verify-plugin-zip.sh
justlevine, line 1: This feels unnecessary and fragile (in a "who watches the watchmen?" sort of way). Jetpack is an implementation detail.

Can we punt this to a followup PR and discuss there? (Two other possible approaches that are probably more scalable is to either update the plugin-check.yml's composer step to build the production version or to just (finally) backport the playground preview workflow from wordpress/ai - either one should catch Composer fatals whether they come from Jetpack Autoloader or anything else. without adding anything else we need to actively maintain/code-review)

justlevine (changes requested):
Plugin changes built locally (`npm ci & composer install --no-dev --optimize-autoloader` then `npm run plugin-zip`) and tested on playground with WP_DEBUG on, no warnings or errors - LGTM.

I _did not_ test or review the script, and recommend we punt that to a separate PR.

Image

scripts/verify-plugin-zip.sh
justlevine: (Regardless of the [how](https://github.com/WordPress/mcp-adapter/pull/284#discussion_r3772552772), i'd recommend we validate on PRs (test.yml or plugin-check) and not just as a release step i.e. _before_ we merge in something that breaks but the build and not just catch it in release)

justlevine (approved):

PR #285, opened 2026-08-13
docs: rebuild readme.txt for the WordPress.org listing

abhi3315:
## What?

See #178

Rebuilds `readme.txt` for the WordPress.org listing, following [`WordPress/ai/readme.txt`](https://github.com/WordPress/ai/blob/trunk/readme.txt). Also removes two packaging entries for files that aren't in the repo.

## Why?

A few things would be wrong on the .org page as it stands. The Changelog tab would be empty, because `== Changelog ==` just links out to GitHub releases. `== Installation ==` tells people to download from a GitHub release, which won't be right once the plugin is in the directory. There's no FAQ. And `Contributors:` only lists `wordpressdotorg`.

Separately, `package.json` `files` lists `CHANGELOG.md` and `LICENSE`. Neither file exists, and `.gitattributes` export-ignores `CHANGELOG.md` at the same time, so the two disagree.

## How?

- Added `For Developers`, `Frequently Asked Questions` and `Upgrade Notice` sections. The full changelog is now inline, 0.6.1 back to 0.1.0, with the breaking changes repeated under `Upgrade Notice`.
- `Installation` now starts with the plugin directory. Composer and Jetpack Autoloader notes moved into `For Developers`.
- `Stable tag: 0.6.1`, to match `mcp-adapter.php`.
- Removed the `CHANGELOG.md` entries from `package.json` and `.gitattributes`, and changed `LICENSE` to `LICENSE.md`, which is the real filename.
- No Screenshots section. There's no admin UI in trunk yet (#184).

**Contributors** is now `wordpressdotorg, ovidiu-galatan`. @gziolo asked for `galatanovidiu` in #178, but that's the GitHub handle, and this field needs a WordPress.org username. `profiles.wordpress.org/galatanovidiu/` 404s and `ovidiu-galatan` works. The parser drops names it can't find, so the GitHub handle would have credited nobody. Happy to change it if I've picked the wrong person.

### Use of AI Tools

AI assistance: Yes
Tool(s): Claude Code
Model(s): Claude Opus 5
Used for: Writing the readme.txt text and turning the existing release notes into the inline changelog. I checked everything against the repo and against the .org readme parser, and I'm responsible for what's here.

## Testing Instructions

1. Put the [raw readme](https://raw.githubusercontent.com/abhi3315/mcp-adapter/fix/readme-txt-wporg-listing/readme.txt) through the [readme validator](https://wordpress.org/plugins/developers/readme-validator/).
2. Run `npm run plugin-zip` and check that `readme.txt` and `LICENSE.md` are both in the zip.
3. Check `Stable tag` matches `Version` in `mcp-adapter.php`. Both should be `0.6.1`.
4. Plugin Check runs in CI.

Every section is under the parser's limits. The biggest is Changelog, at 906 words out of 5,000.

## Changelog Entry

> Changed - Rebuilt `readme.txt` for the WordPress.org listing with the full changelog inline, and removed `package.json` and `.gitattributes` entries for files that aren't in the repo.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -29,44 +29,25 @@ The MCP Adapter bridges WordPress's Abilities API with the [Model Context Protoc
 
 1. Go to **Plugins > Add New** in your WordPress admin, search for "MCP Adapter", and click **Install Now**.
 2. Activate the plugin through the 'Plugins' screen in WordPress.
-3. Register the abilities you want to expose, or install a plugin that registers them for you. Permission checks run against the current user before anything is executed.
+3. Register the abilities you want to expose, or install a plugin that registers them for you.
 
 You can also upload the plugin ZIP via **Plugins > Add New > Upload Plugin**, or install it with WP-CLI:
 
 `wp plugin install mcp-adapter --activate`
 
-WordPress 6.9 or newer is required, as the Abilities API ships in WordPress core from that release.
-
-== For Developers ==
-
-MCP Adapter is built to be extended, and can also be consumed directly as a library.
-
-**Use it as a Composer dependency:**
-
-`composer require wordpress/mcp-adapter`
-
-If you vendor the library inside your own plugin, load it through the [Jetpack Autoloader](https://github.com/Automattic/jetpack-autoloader) so that every plugin on the site resolves the same `WP\MCP` implementation. Running an independently namespace-prefixed copy alongside the plugin is not supported.
-
-**Extend the adapter:**
-
-* **Custom Transports** – Implement `McpTransportInterface` to add transport protocols beyond the built-in HTTP and STDIO support.
-* **Custom Error Handlers** – Implement `McpErrorHandlerInterface` for per-server error handling strategies.
-* **Observability Handlers** – Record events and timings into your own metrics pipeline.
-* **Multiple Servers** – Use `create_server()` to run several MCP servers with independent configurations.
-* **Hooks** – Filters and actions throughout the codebase, all prefixed `mcp_adapter_`.
-
-**Get started:**
-
-1. Read the [developer documentation](https://github.com/WordPress/mcp-adapter/tree/trunk/docs) for architecture, getting started guides, and troubleshooting.
-2. Read the [Contributing Guide](https://github.com/WordPress/mcp-adapter/blob/trunk/CONTRIBUTING.md) for development setup.
-3. Join the conversation in [#core-ai on WordPress Slack](https://wordpress.slack.com/archives/C08TJ8BPULS).
-4. Browse the [GitHub repository](https://github.com/WordPress/mcp-adapter) and open issues or pull requests.
-
 == Frequently Asked Questions ==
 
-= What do I need to run this? =
+= What is MCP? =
 
-WordPress 6.9 or newer and PHP 7.4 or newer. The Abilities API is included in WordPress core from 6.9, so no additional plugin is required.
+Model Context Protocol is an open standard for connecting AI agents to external systems. Instead of every AI client needing its own custom integration, a system exposes what it can do as MCP tools, resources and prompts, and any MCP-compatible client can use them. See [modelcontextprotocol.io](https://modelcontextprotocol.io) for the specification.
+
+= What is MCP Adapter? =
+
+MCP Adapter is the server and transport layer. It does not provide any AI features itself.
+
+What it exposes comes from the WordPress Abilities API. An ability is a single capability registered by WordPress core or by another plugin, such as reading posts or updating a setting. MCP Adapter takes the abilities registered on your site, makes them available to MCP clients, and handles protocol negotiation, permissions and error handling.
+
+That means you also need abilities on the site for this to be useful, from core or from a plugin that registers its own.
 
 = How do AI agents connect to my site? =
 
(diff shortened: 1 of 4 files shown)


readme.txt
```diff
@@ -27,12 +27,203 @@ The MCP Adapter bridges WordPress's Abilities API with the [Model Context Protoc
...
+- `_meta` is preserved on resource contents, embedded resources, content blocks, and prompt messages.
+- Malformed `_meta` is omitted without discarding the payload it accompanies.
+- `mimeType` is emitted exactly as declared, including values with parameters such as `text/html;profile=mcp-app`.
+- Blob-only resource contents are handled correctly.
+- Resource URI schemes are matched case-insensitively, so clients can read resources even when they normalise the scheme to lowercase.
+- Empty arguments are normalised for schema-defining abilities, allowing valid zero-argument tool calls to execute.
+- `wp mcp-adapter serve` keeps JSON-RPC stdout clean when selecting the default server.
+- WP-CLI's global `--user` argument is used instead of registering a conflicting local option.
+
+= 0.5.0 - 2026-04-15 =
```
justlevine: IMO better to put this and everything earlier into CHANGELOG.md to keep this file from getting unwieldly. Once we hit 1.0 (i'm assuming the submission itself), we'd want _everything_ pre-release in CHANGELOG.md and out of this file.

readme.txt
```diff
@@ -27,12 +27,203 @@ The MCP Adapter bridges WordPress's Abilities API with the [Model Context Protoc
...
+If you vendor the library inside your own plugin, load it through the [Jetpack Autoloader](https://github.com/Automattic/jetpack-autoloader) so that every plugin on the site resolves the same `WP\MCP` implementation. Running an independently namespace-prefixed copy alongside the plugin is not supported.
+
+**Extend the adapter:**
+
+* **Custom Transports** – Implement `McpTransportInterface` to add transport protocols beyond the built-in HTTP and STDIO support.
+* **Custom Error Handlers** – Implement `McpErrorHandlerInterface` for per-server error handling strategies.
+* **Observability Handlers** – Record events and timings into your own metrics pipeline.
+* **Multiple Servers** – Use `create_server()` to run several MCP servers with independent configurations.
+* **Hooks** – Filters and actions throughout the codebase, all prefixed `mcp_adapter_`.
+
```
justlevine: I'd strip all of this from here

1. The audience of this file are end users installing it from W.org
2. Once it's on w.org Composer installation should be a last resort.
3. Single source of truth, as part of the docs cleanup we want _fewer_ places repeating the same things.

readme.txt
```diff
@@ -27,12 +27,203 @@ The MCP Adapter bridges WordPress's Abilities API with the [Model Context Protoc
...
+2. Read the [Contributing Guide](https://github.com/WordPress/mcp-adapter/blob/trunk/CONTRIBUTING.md) for development setup.
+3. Join the conversation in [#core-ai on WordPress Slack](https://wordpress.slack.com/archives/C08TJ8BPULS).
+4. Browse the [GitHub repository](https://github.com/WordPress/mcp-adapter) and open issues or pull requests.
+
+== Frequently Asked Questions ==
+
+= What do I need to run this? =
+
+WordPress 6.9 or newer and PHP 7.4 or newer. The Abilities API is included in WordPress core from 6.9, so no additional plugin is required.
+
```
justlevine: Isn't this redundant? We have Requires headers...

readme.txt
```diff
@@ -27,12 +27,203 @@ The MCP Adapter bridges WordPress's Abilities API with the [Model Context Protoc
...
+= Does every registered ability get exposed automatically? =
+
+No, exposure is opt-in. An ability is exposed through the default MCP server when `meta.mcp.public` is true, or when it isn't set and the general `meta.public` flag is true. Setting `meta.mcp.public` to false always wins. Permission callbacks and capability checks still run against the current user before anything executes.
+
+Note that since 0.6.0, an ability marked `meta.public` for reasons unrelated to MCP will be exposed unless it opts out. If you register public abilities, check whether you want them reachable over MCP.
+
+= Can another plugin bundle MCP Adapter as a library? =
+
+Yes, via `composer require wordpress/mcp-adapter`. Load it through the Jetpack Autoloader so that all plugins on the site resolve the same `WP\MCP` implementation. Running an independently namespace-prefixed copy alongside the plugin is not supported.
+
```
justlevine: Same as above, it's possible but we want to bury this possibility as much as possible in the docs.

readme.txt
```diff
@@ -27,12 +27,203 @@ The MCP Adapter bridges WordPress's Abilities API with the [Model Context Protoc
...
+* **Hooks** – Filters and actions throughout the codebase, all prefixed `mcp_adapter_`.
+
+**Get started:**
+
+1. Read the [developer documentation](https://github.com/WordPress/mcp-adapter/tree/trunk/docs) for architecture, getting started guides, and troubleshooting.
+2. Read the [Contributing Guide](https://github.com/WordPress/mcp-adapter/blob/trunk/CONTRIBUTING.md) for development setup.
+3. Join the conversation in [#core-ai on WordPress Slack](https://wordpress.slack.com/archives/C08TJ8BPULS).
+4. Browse the [GitHub repository](https://github.com/WordPress/mcp-adapter) and open issues or pull requests.
+
+== Frequently Asked Questions ==
```
justlevine: I would add two more (imo the most important ones):

1. What is MCP?
2. What is MCP _Adapter_? (This is where we stress that it's just a server + transport, and any consumable functionality (e.g. abilities) come from somewhere else.

abhi3315: Agreed, those are the important ones. Adding both.

readme.txt
abhi3315: Will cut it.

readme.txt
abhi3315: Will drop it.

Turns out it wasn't rendering as its own section anyway.

readme.txt
abhi3315: Yeah, fair. Will keep 0.6.1 and 0.6.0 here and move the rest into CHANGELOG.md.

@gziolo asked for the full changelog inline over in #178, so I'll mention it there.

readme.txt
abhi3315: Dropping it. 

Want to keep one bit though, that the Abilities API ships in core from 6.9 so you don't need the separate plugin. That's not in any header and 0.6.0 dropped support for the standalone one. I'll fold it into the "What is MCP Adapter?" answer unless you'd rather it sat somewhere else.

justlevine: IWhy would anyone think you need a separate plugin? That sounds like an AI Slop decision-log piece of information, so id rather it sat _nowhere_. The plugin requires WordPress 6.9+ - that's all that the user needs to know, IMO, not that before WordPress 6.9 this plugin won't work _and_ the Abilities API doesn't exist...

abhi3315: RIght, make sense! I addressed you suggestions, please take a look again!

justlevine (approved):
This looks great, Thanks @abhi3315 🙇

PR #288, opened 2026-08-16
feat: deprecate usage of MCP Adapter as bundled dependency

matiasbenedetto:
## What?

- Part of #178 

Deprecates usage of MCP Adapter as a bundled library, in favor of a canonical plugin.

## Why?

Prevents library conflicts in preparation of release on w.org

## How?

- Adds a class_exists() check into the `Autoloader::autoload()` to check if `Core\McpAdapter` is already loaded. If it is, then logs a `_doing_it_wrong()` and `admin_notice` telling the user that they're using an old plugin that is bundling MCP Adapter as a library.
- Relocates the plugin constant definitions into `Plugin`, and then uses a missing `MCP_ADAPTER_VERSION` constant as the indicator that `Core\McpAdapter` is being autoloaded via a library instead of via the plugin.

> [!NOTE]
> The PR currently uses `x.y.z` as a placeholder for the next release version. This will need to be replaced before release.

### Use of AI Tools

- Original version of this PR was created by @matiasbenedetto 's Claude
- Manually refactored by @justlevine, who used OpenCode (glm-5.3) to backfill the `AutoloaderTest` integration test.

## Screenshots or screencast

image

## Changelog Entry

> Changed -  Usage of MCP Adapter as a bundled library has been deprecated in favor of using the canonical MCP Adapter plugin. See the [vx.y.z migration guide](migration/vx.y.z.md) for instructions on how to migrate away from a bundled copy of MCP Adapter.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fplugins.php%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.5%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fmcp-adapter%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-288-ce4a145d7c4810c71f679dab741d380c5566c08f-mcp-adapter.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


stormrockwell:
From my understanding, we don't want to recommend packaging the MCP adapter in composer require, regardless of the method since support may drop in the future

matiasbenedetto:
Thanks @stormrockwell — you were right, and #289 made it clearer what the guide should have said. I've updated it.

**What changed**

Two things in the migration guide were still teaching the old way of doing things:

1. **`Requires Plugins` is now the answer.** The guide used to say "check `class_exists()` on `plugins_loaded`". It now leads with adding `Requires Plugins: mcp-adapter` to your plugin header, like #289 does. WordPress won't let your plugin activate without MCP Adapter, and it loads MCP Adapter first, so there's nothing to wire up. The `class_exists()` check is still there, but only as a safety net for older WordPress or if you'd rather fail softly than block activation.

2. **The guide now helps you get *rid* of Jetpack Autoloader.** It used to explain how to set it up. There's now a step for removing it: check whether any other package you bundle still needs it, remove it, switch back to `vendor/autoload.php`, and delete the leftover generated files (`composer remove` doesn't always clean up `vendor/autoload_packages.php`, and a stale one works locally then breaks in production).

   One ordering note: removing Jetpack Autoloader comes *after* dropping the bundled copy, not before. If you pull it while you're still shipping a copy, a stale copy can win against the active plugin.

I also fixed a spot where this guide contradicted #289 about `WP_MCP_VERSION`. Now it matches: check the class to see if MCP Adapter is there, check the constant if you need a minimum version. `Requires Plugins` can't express a version constraint.

**About #289**

Same direction, and yours goes further — good. Two practical notes:

- **They'll conflict.** I tested the merge: `README.md` and `docs/getting-started/installation.md`. I added a one-line pointer to each, and those lines sit inside blocks #289 deletes. Merge #289 first and I'll rebase, or tell me to drop the pointers now.
- **The code changes here don't overlap #289 at all** — it's docs-only. Worth noting your new troubleshooting line, that "The Composer autoloader was not found" only affects source checkouts, is only fully true once this PR lands. Right now a Composer-installed copy triggers it too.

**One thing I'd like a call on**

My guide documents installing via `wp-plugin/mcp-adapter` and `wpackagist-plugin/mcp-adapter`. #289 leaves Composer out of the install docs entirely, and its testing steps expect zero `wpackagist` hits.

I added it because the #178 checklist says "prioritize usage via plugin (and wpackagist/wp-packages for Composer)". But those only work once we're listed on .org, so leaving them out for now is also reasonable. Happy to drop that section if you'd rather keep Composer out of the docs completely.

Last thing: if it's easier to review, I can split this into a code-only PR (the crash fix, conflicts with nothing) and a docs follow-up rebased on #289. Just say the word.

matiasbenedetto:
Rebased on trunk now that #289 has merged. Conflicts are gone.

**What I changed to fit the new docs**

- **Dropped the wpackagist / wp-packages section.** This was the open question from my last comment. #289 kept Composer out of the install docs, so leaving it in here would have made this guide the only place still suggesting a Composer install. The guide now just links to the Installation Guide.
- **Moved the pointer links.** They used to sit in the Composer sections #289 deleted. They're now one line each in "As a dependency of your plugin" (README) and "Depending on MCP Adapter from your own plugin" (installation.md), aimed only at people who already bundle a copy.
- **Added a CHANGELOG entry**, since `CHANGELOG.md` landed while this PR was open.

The guide is now purely about moving *off* a bundled copy. It doesn't teach installing via Composer, and it doesn't teach setting up the Jetpack Autoloader — it explains how to remove it.

**Check that this still holds**

Your verification grep from #289 passes:

```console
$ grep -rn -i "composer require\|wpackagist\|jetpack" README.md readme.txt docs/ \
    | grep -v "docs/guides/testing.md" | grep -v "docs/migration/composer-to-plugin.md"
docs/troubleshooting/common-issues.md:84:# Check the autoloader exists (Jetpack Autoloader, not autoload.php)
readme.txt:72: ... (0.6.1 changelog entry)
readme.txt:93: ... (0.6.0 changelog entry)
```

The `common-issues.md` line is the one you expected. The two `readme.txt` hits are changelog entries from #285, not install advice. Zero `wpackagist` anywhere.

Tests, `phpcs`, and `phpstan` are green against the new trunk.

The code changes are unchanged from before, and still don't overlap anything in #289. Offer stands to split them into their own PR if that's easier to review.

matiasbenedetto:
Added a `_doing_it_wrong()` notice for non-canonical usage, and reworked how it detects that.

## What it does

The MCP Adapter plugin is the canonical copy. It owns the `WP\MCP\` namespace, the hooks, the REST routes, and the default server ID — all global, so a second copy doesn't coexist with the plugin, it *replaces* it. `McpAdapter::instance()` now says so when the code running isn't the plugin's own.

Developer-facing only: it needs `WP_DEBUG`, and nothing about runtime behavior changes.

## How it detects it

My first attempt just checked whether `WP_MCP_DIR` was defined. That turned out to answer the wrong question — the bootstrap defines that constant whether or not its own classes are the ones executing. So it was silent in the case that matters most: the plugin is active, but Jetpack Autoloader resolved `WP\MCP\` to a newer copy bundled inside some other plugin, which then takes over site-wide.

I confirmed that against wp-env before changing it. With the constant pointing at one directory and the class loaded from another, the old check reported "canonical".

It now compares where the class was **actually loaded from** against the plugin directory, via `ReflectionClass::getFileName()`. Both paths go through `realpath()` first, since symlinked plugin directories are normal in local and Bedrock-style setups and would otherwise look foreign. The directory is compared with a trailing slash so a sibling like `mcp-adapter-fork/` can't match on prefix alone.

That also let me drop the old message's claim that the code came from "a copy bundled inside another plugin" — we'd only established the bootstrap hadn't run. It could equally have been an mu-plugin or a custom loader.

## Evidence

Front end of a wp-env site with `WP_DEBUG` on. Both scenarios use a fake `acme-widgets` plugin that bundles MCP Adapter.

**The plugin is not active, and something else supplied the classes.**

Notice: the MCP Adapter plugin is not active

**The plugin is active, but a bundled copy is what's running.** This is the case the constant check missed. Note that the message names the file, so you can tell which plugin is shipping it:

Notice: code loaded from a bundled copy instead

**Control — plugin active, no bundled copy.** Clean page, no notice:

```console
$ wp plugin activate mcp-adapter
Success: Activated 1 of 1 plugins.
$ curl -s http://localhost:8889/ | grep -c "called <strong>incorrectly"
0
```

There's also a unit test for this. `WP_UnitTestCase` fails on an unexpected `_doing_it_wrong()`, so a regression that made normal installs noisy would break the suite.

`is_loaded_from()` is unit-tested directly for the plugin directory, a foreign directory, and the shared-prefix case. 1042 tests, `phpcs` and `phpstan` clean.

## Two things for reviewers

**The version string is a guess.** `_doing_it_wrong()` needs one and I used `0.6.2`. Wants changing if the next release isn't a patch.

**This is firmer than anything currently written down.** #172 called bundling-through-Jetpack-Autoloader supported, and this now nags about it in every debug environment — including for a site owner running someone else's plugin that bundles a copy. It matches where #289 and this PR's review landed, but it is a policy call rather than a bug fix, so I'd rather it be reviewed as one than slip through. Happy to drop it or soften it to the inactive-plugin case only.

<sub>Screenshots are hosted on an `evidence/pr-288` branch of my fork, since GitHub has no API for attaching images to comments. Nothing in this PR's diff; delete any time.</sub>

matiasbenedetto:
Fixed the `Requires Plugins` documentation in 3336d26.

The migration guide no longer says that WordPress installs, activates, or load-orders MCP Adapter. It now describes the actual behavior: WordPress surfaces the dependency relationship and prevents the dependent plugin from being activated while MCP Adapter is missing or inactive.

The migration step also now says explicitly that `Requires Plugins` does not control plugin load order, so integrations should wait until `plugins_loaded` before checking for or using the `WP\MCP\` classes. The existing example already uses that safe timing and remains unchanged.

matiasbenedetto:
Tested this against a real bundler — `seo-by-rank-math`, which vendors `wordpress/mcp-adapter` — and found the check couldn't fire in the case it was written for. Fixed, and the fix is now verified end to end against the unmodified plugin.

## The gap

Rank Math 1.0.276 bundles **v0.4.1**, declared `type: library`, loaded through the stock Composer autoloader — no Jetpack Autoloader anywhere in its `vendor/`. From `rank-math.php:262`:

```php
include __DIR__ . '/vendor/autoload.php';
if ( class_exists( 'WP\MCP\Core\McpAdapter' ) && function_exists( 'wp_get_abilities' ) ) {
	\WP\MCP\Core\McpAdapter::instance();
}
```

With both plugins active it works — but only because `active_plugins` happens to be alphabetical, so `mcp-adapter` loads first and its classes win. Reordering that option is enough to flip it:

```
WP_MCP_DIR            = .../plugins/mcp-adapter/     ← plugin bootstrap ran
WP_MCP_VERSION        = 0.6.1                         ← installed plugin
running class         = .../seo-by-rank-math/vendor/wordpress/mcp-adapter/includes/Core/McpAdapter.php
running VERSION const = 0.4.1                         ← two minor versions old
```

Two minor versions of adapter code running against a 0.6.1 plugin, and **nothing reported it**. The check lived in `McpAdapter::instance()`, and when a bundled copy wins the race, `McpAdapter` *is* that copy — v0.4.1 has no such method, so it cannot report its own takeover.

## The fix

The takeover check moves to `mcp-adapter.php`. That file always runs while the plugin is active, whichever copy the autoloader resolved, so it is the one place the check can be relied on.

Three things it has to respect:

- **Self-contained.** It must not call into the loaded classes. The method it would call does not exist in whichever version won, which would turn a notice into a fatal on exactly the sites being warned.
- **A closure, not a named function.** A named function in that file reintroduces the compile-time redeclare fatal this PR opens by fixing.
- **Disjoint from the other check.** `McpAdapter::instance()` now only handles "no plugin active, something else supplied the classes". The two conditions are mutually exclusive, so exactly one notice is ever raised.

The message also names both versions now, since that is the whole diagnosis in one line.

## Evidence

**Unmodified Rank Math 1.0.276 + this branch, Rank Math loading first.** No patching, no simulation — the stock plugin from the .org directory:

Notice naming Rank Math's bundled v0.4.1 taking over from the 0.6.1 plugin

> Notice: Function WP\MCP\Core\McpAdapter was called **incorrectly**. MCP Adapter 0.4.1 was loaded from `/var/www/html/wp-content/plugins/seo-by-rank-math/vendor/wordpress/mcp-adapter/includes/Core/McpAdapter.php`, so it is running instead of the MCP Adapter plugin (version 0.6.1) installed on this site. Whichever plugin ships that copy should depend on the MCP Adapter plugin with the "Requires Plugins" header rather than bundling its own, since the bundled copy replaces the installed plugin site-wide.

**Default plugin order** — the plugin's own copy wins, no notice:

```console
$ curl -s http://localhost:8888/ | grep -c "called <strong>incorrectly"
0
```

That second result matters as much as the first: a real third-party bundler on a normally-ordered site produces no false positive.

1040 tests, `phpcs` and `phpstan` clean.

## Notes

**Coverage changed shape.** Rather than duplicate the path logic into the bootstrap, I removed it from `McpAdapter`, so it now exists in one place. The three unit tests for the old private helpers are replaced by one asserting a normal install resolves inside `WP_MCP_DIR` — the false-positive guard. The Rank Math check above is the real coverage for the takeover path.

**One inherent limit, worth stating.** This cannot fire when an old bundled copy wins *and* the plugin is not installed, because then no current code runs at all. It protects sites that have the plugin, which is the population we can actually reach.

**Possibly worth raising with Rank Math.** Their bundled copy is two minor versions behind and they are not using Jetpack Autoloader, so which copy wins is decided by plugin load order. Happy to open something on their side if that is useful.

The earlier caveats still stand: the `0.6.2` version string in `_doing_it_wrong()` is a guess, and the notice as a whole is a policy call rather than a bug fix.

<sub>Screenshot hosted on an `evidence/pr-288` branch of my fork, since GitHub has no API for attaching images to comments. Not part of this PR's diff.</sub>

matiasbenedetto:
Follow-up: the check now also reports a bundled copy that is installed but **not** currently running.

## Why

My last comment noted that Rank Math works today only because `active_plugins` happens to be alphabetical, so `mcp-adapter` loads first and wins. That is not a property either plugin controls — deactivating and reactivating a plugin is enough to reorder autoloader registration, and then the bundled v0.4.1 takes over.

Reporting only the takeover means staying silent right up until the moment it happens. That is the worst possible time to start telling someone.

## How

Registered Composer autoloaders can be asked where they *would* resolve a class, without loading it:

```php
$loader->findFile( 'WP\MCP\Core\McpAdapter' )
```

Rank Math's `ClassLoader` is registered even when its copy loses the race, so this finds the dormant copy with no filesystem scanning — no globbing over `wp-content/plugins/*/vendor/`. `findFile()` is duck-typed rather than checked against `Composer\Autoload\ClassLoader`, so scoped and prefixed Composer builds are covered too.

## Evidence

**Stock Rank Math 1.0.276, default plugin order** — the installed plugin is running correctly, and this is the state that was silent before:

Notice reporting a dormant bundled copy inside seo-by-rank-math

> Notice: Function WP\MCP\Core\McpAdapter was called **incorrectly**. Another copy of MCP Adapter is installed at `/var/www/html/wp-content/plugins/seo-by-rank-math/vendor/wordpress/mcp-adapter/includes/Core/McpAdapter.php`. Only one copy can run per request, and which one wins is decided by plugin load order, so this site may start running that copy instead of the MCP Adapter plugin (version 0.6.1) without anything changing. Whichever plugin ships it should depend on the MCP Adapter plugin with the "Requires Plugins" header rather than bundling its own.

All three states against the same stock plugin:

| Situation | Result |
|---|---|
| Default order, plugin wins | Reports the dormant copy — **previously silent** |
| Rank Math first, its copy wins | Reports the takeover, naming both versions |
| Rank Math deactivated | Silent |

The two messages are mutually exclusive: a takeover returns before the dormant scan runs, so the middle row emits exactly one notice, not two. I checked that explicitly rather than assuming it.

1040 tests, `phpcs` and `phpstan` clean.

## Two things reviewers should weigh

**This widens the blast radius, deliberately.** Every `WP_DEBUG` site running Rank Math now gets a notice, even though nothing is broken for them today. That is the intent — the risk is real and currently invisible — but it makes the policy question from my earlier comment sharper rather than softer. If that is too aggressive for now, the dormant case is the part to drop; the takeover case stands on its own.

**It does not find every copy.** Only copies reachable through an autoloader that is registered by `init` and exposes `findFile()`. A plugin that `require`s its files directly will not show up. Catching those needs a filesystem scan, which I think belongs in a Site Health test rather than a per-request check — happy to open that as a follow-up issue.

The earlier caveats still stand: the `0.6.2` version string in `_doing_it_wrong()` is a guess, and the notice as a whole is a policy call rather than a bug fix.

<sub>Screenshot hosted on an `evidence/pr-288` branch of my fork, since GitHub has no API for attaching images to comments. Not part of this PR's diff.</sub>

justlevine:
Apologies for the force-push, had to correct a bad merge-conflict

mcp-adapter.php
```diff
@@ -56,3 +71,132 @@ function constants(): void {
 if ( class_exists( Plugin::class ) ) {
 	Plugin::instance();
 }
+
+/*
```
justlevine: This is the wrong place for this. The composer autoloader entrypoint is via `Plugin::instance()` so this file gets bypassed.

docs/README.md
```diff
@@ -24,6 +24,7 @@ Documentation for the WordPress MCP Adapter - transform WordPress abilities into
 
 ## Migration guides
 
+- **[Composer library to canonical plugin](migration/composer-to-plugin.md)** - Moving off a bundled copy of MCP Adapter
```
justlevine: This should be updated to handle ALL v0.X migration requirements.

includes/Core/McpAdapter.php
```diff
@@ -58,6 +58,8 @@ public static function instance(): self {
 		if ( ! isset( self::$instance ) ) {
 			self::$instance = new self();
 
+			self::schedule_canonical_plugin_check();
```
justlevine: By moving it into Autoloader.php we don't need to do this twice. https://github.com/WordPress/mcp-adapter/pull/288#discussion_r3856687412 .

To log a deprecation message, we can instead check for whether the version constant is created by the `Plugin::instance()` or if they're loading this class directly.

justlevine (approved):
I desloped the AI changes and refactored the detection in 0251fa7bf7d46274aac34acea163cef1a95b5045

The current approach moves duplication detection into the `Autoloader`, logging a `doing_it_wrong` and admin notice if a bundled `McpAdapter` is conflicting with the canonical plugin, and adds a deprecation notice if `McpAdapter` is _only_ loaded via a library.

It works by checking the  `WP_MCP_VERSION` constant that is now defined in `Plugin`:  if `McpAdapter` is initialized without that constant set, it means they're autoloading and calling `McpAdapter` directly.

docs/getting-started/installation.md
```diff
@@ -49,8 +49,10 @@ The best way to ensure that MCP Adapter is installed and activated is to include
 
 ### As a Composer Library
 
-While not recommended, there are some situations where you may want to bundle a copy of MCP Adapter with your plugin. In that case, you can add it as a Composer dependency:
-
+> [!NOTE]
+> Bundling MCP Adapter as a Composer library is not recommended, as it can lead to conflicts with the MCP Adapter plugin or other plugins that may be bundling their own copy of MCP Adapter. It is strongly recommended to use the plugin dependency method above instead.
```
jeffpaul, line 53: Should put something like this in at the top of the CHANGELOG.md entry for the release, at the top of the changelog section in readme.txt, and as an upgrade notice in readme.txt to try and flag to folks more broadly about this concern/issue.

mcp-adapter.php
```diff
@@ -29,22 +29,6 @@
...
-function constants(): void {
-	/**
-	 * Shortcut constant to the path of this file.
-	 */
-	define( 'WP_MCP_DIR', plugin_dir_path( __FILE__ ) );
-
-	/**
-	 * Version of the plugin.
-	 */
-	define( 'WP_MCP_VERSION', '0.6.1' );
```
jeffpaul, line 44: Is see this is removed here, but where is the version constant added elsewhere so this is defined within the plugin and bumped with new releases?

justlevine: @jeffpaul it's now in `Plugin::constants()`, so it can consume `Core\McpAdapter::VERSION`, instead of being another source of truth that needs manual bumping.

docs/getting-started/installation.md
justlevine: Added the `readme.txt` upgrade notice in https://github.com/WordPress/mcp-adapter/pull/288/commits/981213725e2526067a401a091bb9652b35ef8103 . `CHANGELOG.md` already has it, and it looks like @ovidiu-galatan handles the `readme.txt` changelog manually before cutting the release.

@jeffpaul lmk if there's anything else you'd like to me change before merge.

jeffpaul: I believe the Upgrade Notice has a max of 300 characters, which is likely why the Plugin Check run is failing.  So you'll want to shorten that significantly in readme.txt to clear the check.

jeffpaul (approved):

docs/getting-started/installation.md
justlevine: Good catch, thanks! Trimmed in https://github.com/WordPress/mcp-adapter/pull/288/commits/ce4a145d7c4810c71f679dab741d380c5566c08f . Once the CI is passing, I'll resolve this thread and merge.

justlevine: image

PR #287, opened 2026-08-14
Replace MCP DTOs with a descriptor-backed dual-revision runtime

galatanovidiu:
## What?

See [WordPress/php-mcp-schema#14](https://github.com/WordPress/php-mcp-schema/pull/14).

This replaces the Adapter's concrete, revision-specific DTO surface with revision-neutral arrays validated and serialized by descriptor-backed schema catalogs. It then uses that boundary to support both MCP `2025-11-25` and `2026-07-28`, including stateless multi-round tool, resource, and prompt execution.

This is an intentional breaking change for PHP integrations that consumed DTO-typed public payloads. Prompt-builder results, initialize/list filter payloads, component getters, content blocks, and error-factory results now use arrays. No DTO compatibility facade or dual-shape return is retained.

## Why?

The previous DTO graph coupled Adapter execution code to one protocol revision and duplicated exact wire-shape ownership that belongs in `php-mcp-schema`. That made a second revision require parallel codecs and made JSON object/list identity easy to lose at transport and serialization boundaries.

The new split keeps exact revision schemas, hydration, validation, and JSON identity in `php-mcp-schema`. The Adapter retains protocol negotiation, sessions and headers, execution classification, MRTR policy, observability, and WordPress Ability normalization.

### Revision-neutral execution and descriptor-backed output

- Converts stored components, handlers, public hooks, content blocks, and errors to revision-neutral arrays.
- Selects one immutable schema catalog per request and narrows at revision-specific wire encoders.
- Hydrates every protocol result before emission; output validation is inherent and always enabled.
- Keeps registration-time deep validation behind its existing opt-in filter.
- Drops only an invalid component from list results, with logging and `_doing_it_wrong()`, instead of disabling the server.

### JSON identity and transport boundaries

- Adds a shared identity-preserving JSON-RPC decoder for HTTP, STDIO, and custom transports.
- Distinguishes JSON objects from lists for batches, tool arguments, empty maps, and nested values.
- Keeps callback-facing parameters as associative arrays.
- Treats JSON object key order as insignificant while retaining exact checks for content, escaping, number formatting, and `{}` versus `[]`.

### Exact dual-revision runtime

- Supports exactly `2025-11-25` and `2026-07-28`.
- Keeps legacy initialize negotiation, sessions, and ping behavior.
- Resolves modern requests independently from required header/body protocol metadata and does not touch legacy session state.
- Applies per-era method tables, not-found policy, typed protocol errors, HTTP statuses, result metadata, and cache fields.
[...]

PR #299, opened 2026-08-24
Fix resources/read 500 on ability-backed resources with an object input schema

### LINKED ISSUE #261: resources/read on ability-backed resources 500s: McpResource::execute() calls ability->execute() with zero args

shoemoney:
## What?
Closes #261

`resources/read` returns a 500 for any ability-backed resource whose ability declares an object `input_schema`. This routes the empty argument set through `AbilityArgumentNormalizer` so the ability receives `[]` instead of being called with zero arguments.

## Why?
`McpResource::execute()` and `check_permission()` called the backing ability with no arguments at all. For an ability with `'input_schema' => array( 'type' => 'object' )`, core then validates `null` against that schema and the read fails:

```
Internal error: Ability "..." has invalid input. Reason: input is not of type object.
```

Full credit to @lhero-org for the root-cause analysis in #261, which identified the exact failing call and ruled out the sibling `tools/call` path already fixed in #230.

## How?
`McpTool` and `McpPrompt` already normalize their arguments through `AbilityArgumentNormalizer`; `McpResource` was the one component that did not. This applies the same pattern in both `execute()` and `check_permission()`:

- Abilities with an object schema now receive an empty array.
- Abilities with no schema still receive `null`, unchanged legacy behavior.
- A schema with a top-level `default` still has that default applied by core.

Scope note: the `resources/read` protocol params (`uri`) are intentionally *not* forwarded as ability input. They are protocol-level, and the added test asserts the ability receives exactly `[]`, so a regression that forwarded `uri` would fail.

## Testing Instructions
1. Register an ability whose `input_schema` is `array( 'type' => 'object' )` and expose it as an MCP resource via `meta.mcp.type = 'resource'`.
2. Call `resources/read` for that resource's URI.
3. Before this change: `Internal error: Ability "..." has invalid input. Reason: input is not of type object.` After: the resource reads normally.

Automated coverage: `test_ability_backed_resource_with_object_input_schema_executes` in `tests/phpunit/Unit/Resources/McpResourceTest.php`.

```
composer test -- --filter McpResourceTest
```

## Changelog Entry

> Fixed - `resources/read` returning a 500 for ability-backed resources whose ability declares an object input schema.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fplugins.php%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.5%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fmcp-adapter%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-299-6d3ba8bfabc6e06e3e53f3d76f9ebaab5b3b7e9d-mcp-adapter.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Domain/Resources/McpResource.php
+++ b/includes/Domain/Resources/McpResource.php
@@ -11,6 +11,7 @@ declare( strict_types=1 );
 namespace WP\MCP\Domain\Resources;
 
 use WP\MCP\Domain\Contracts\McpComponentInterface;
+use WP\MCP\Domain\Utils\AbilityArgumentNormalizer;
 use WP\MCP\Domain\Utils\McpValidator;
 use WP\MCP\Infrastructure\ErrorHandling\Contracts\McpErrorHandlerInterface;
 use WP\MCP\Infrastructure\Observability\FailureReason;
@@ -269,10 +270,15 @@ final class McpResource implements McpComponentInterface {
 	 * @return mixed
 	 */
 	public function execute( $arguments ) {
-		// Ability-backed resources match existing behavior: no args passed to abilities.
+		// Ability-backed resources receive no ability input: the resources/read
+		// params (uri) are protocol-level, not ability arguments. The empty
+		// argument set is normalized so abilities with an object input schema
+		// receive an empty array instead of null, matching McpTool and McpPrompt.
 		if ( null !== $this->ability ) {
+			$args = AbilityArgumentNormalizer::normalize( $this->ability, array() );
+
 			try {
-				return $this->ability->execute();
+				return $this->ability->execute( $args );
 			} catch ( \Throwable $throwable ) {
 				return new WP_Error(
 					'mcp_execution_failed',
@@ -305,10 +311,15 @@ final class McpResource implements McpComponentInterface {
 	 * @return bool|\WP_Error
 	 */
 	public function check_permission( $arguments ) {
-		// Ability-backed resources match existing behavior: no args passed to abilities.
+		// Ability-backed resources receive no ability input: the resources/read
+		// params (uri) are protocol-level, not ability arguments. The empty
+		// argument set is normalized so abilities with an object input schema
+		// receive an empty array instead of null, matching McpTool and McpPrompt.
 		if ( null !== $this->ability ) {
+			$args = AbilityArgumentNormalizer::normalize( $this->ability, array() );
+
 			try {
-				return $this->ability->check_permissions();
+				return $this->ability->check_permissions( $args );
 			} catch ( \Throwable $throwable ) {
 				return new WP_Error(
 					'mcp_permission_check_failed',
--- a/includes/Domain/Resources/McpResource.php
+++ b/includes/Domain/Resources/McpResource.php
@@ -270,10 +270,7 @@ final class McpResource implements McpComponentInterface {
 	 * @return mixed
 	 */
 	public function execute( $arguments ) {
-		// Ability-backed resources receive no ability input: the resources/read
-		// params (uri) are protocol-level, not ability arguments. The empty
-		// argument set is normalized so abilities with an object input schema
-		// receive an empty array instead of null, matching McpTool and McpPrompt.
+		// resources/read params are protocol-level, so the ability gets no input.
 		if ( null !== $this->ability ) {
 			$args = AbilityArgumentNormalizer::normalize( $this->ability, array() );
 
@@ -311,10 +308,7 @@ final class McpResource implements McpComponentInterface {
 	 * @return bool|\WP_Error
 	 */
 	public function check_permission( $arguments ) {
-		// Ability-backed resources receive no ability input: the resources/read
-		// params (uri) are protocol-level, not ability arguments. The empty
-		// argument set is normalized so abilities with an object input schema
-		// receive an empty array instead of null, matching McpTool and McpPrompt.
+		// resources/read params are protocol-level, so the ability gets no input.
 		if ( null !== $this->ability ) {
 			$args = AbilityArgumentNormalizer::normalize( $this->ability, array() );
 
--- a/tests/phpunit/Unit/Resources/McpResourceTest.php
+++ b/tests/phpunit/Unit/Resources/McpResourceTest.php
@@ -56,10 +56,10 @@ final class McpResourceTest extends TestCase {
 				'category'            => 'test',
 				'input_schema'        => array( 'type' => 'object' ),
 				'execute_callback'    => static function ( $input ) {
-					return is_array( $input ) ? 'object schema content' : 'unexpected input';
+					return array() === $input ? 'object schema content' : 'unexpected input';
 				},
 				'permission_callback' => static function ( $input ) {
-					return is_array( $input );
+					return array() === $input;
 				},
 				'meta'                => array(
 					'mcp' => array(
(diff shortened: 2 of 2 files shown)


shoemoney:
The Run Plugin Check failure is unrelated to this change: it flags readme.txt with outdated_tested_upto_header ("Tested up to: 7.0 < 7.1") after the WordPress 7.1 release. This PR does not touch readme.txt, and the same check passed on PR #296 before the release. Every new run will hit this until the header is bumped.

shoemoney:
The Plugin Check failure is the repo-wide outdated_tested_upto_header error in readme.txt (Tested up to: 7.0 < 7.1), not from this diff. PRs #300 and #287 currently fail the same check. Happy to bump the header here if you want it in this PR, otherwise leaving it for a repo-level fix.

shoemoney:
CI note: Run Plugin Check failure is pre-existing on main/trunk and unrelated to this PR's changed files. Fails identically on trunk (07c9912e) with outdated_tested_upto_header (Tested up to: 7.0 < 7.1 in readme.txt) plus reserved contributors warning, unrelated to McpResource.php changes. No code fix required from this PR; rebase/label will clear it.

justlevine:
@shoemoney , you mind updating your PR description to use the repository's [Pull request template](https://github.com/WordPress/mcp-adapter/blob/trunk/.github/PULL_REQUEST_TEMPLATE.md)?

includes/Domain/Resources/McpResource.php
```diff
@@ -269,10 +270,15 @@ public function get_protocol_dto(): ResourceDto {
 	 * @return mixed
 	 */
 	public function execute( $arguments ) {
-		// Ability-backed resources match existing behavior: no args passed to abilities.
+		// Ability-backed resources receive no ability input: the resources/read
+		// params (uri) are protocol-level, not ability arguments. The empty
+		// argument set is normalized so abilities with an object input schema
+		// receive an empty array instead of null, matching McpTool and McpPrompt.
```
justlevine: This reads like AI slop...

includes/Domain/Resources/McpResource.php
```diff
@@ -305,10 +311,15 @@ public function execute( $arguments ) {
 	 * @return bool|\WP_Error
 	 */
 	public function check_permission( $arguments ) {
-		// Ability-backed resources match existing behavior: no args passed to abilities.
+		// Ability-backed resources receive no ability input: the resources/read
+		// params (uri) are protocol-level, not ability arguments. The empty
+		// argument set is normalized so abilities with an object input schema
+		// receive an empty array instead of null, matching McpTool and McpPrompt.
```
justlevine: This also reads like AI slop

shoemoney:
Wow sorry about that fixing.

shoemoney:
The red `Run Plugin Check` here is not from this diff. It reports two findings in `readme.txt`, which this PR does not touch:

```
ERROR   outdated_tested_upto_header — "Tested up to: 7.0" < 7.1
WARNING readme_reserved_contributors — "wordpressdotorg" is reserved
```

The same check has failed on `trunk` on every run since 21 Aug, so any open PR inherits it. #307 bumps the header to 7.1, which should clear the error for all of them.

Everything else here is green: PHPUnit across PHP 7.4–8.4 on WP 6.9 / latest / trunk, PHPCS, PHPStan, and codecov.

shoemoney:
Updated, thanks for the nudge.

- Description now follows the repository template, including the AI-tools disclosure.
- Both comment blocks you flagged are down to a single line each (d7ade87). They were restating the diff rather than explaining it.

That commit also picks up the two Copilot notes: the callbacks now assert the ability receives exactly `[]` — so a regression that forwarded `uri` would fail instead of passing — and the `wp_unregister_ability()` moved into a `finally` so a failed assertion can't leak state into the rest of the suite.

PR #300, opened 2026-08-24
Add SSE streaming to the HTTP transport GET endpoint

### LINKED ISSUE #280: Add real HTTP transport support
Right now, the server is always STDIO - the only difference is whether it connects to a local instance or a remote one. 

This breaks when you try to connect Claude Desktop in Chat / Cowork mode because those work in a sandbox / cloud environment and can't access the STDIO MCP on your computer.

sarthak-19:
## Summary

Fixes #280.

The HTTP transport's `GET` endpoint has always returned `405 Method Not Allowed`, so the only documented way to reach a WordPress MCP server was some form of STDIO on the client side: `wp mcp-adapter serve` running locally, or the `@automattic/mcp-wordpress-remote` proxy, which is also a locally-spawned process that just forwards to HTTP behind the scenes. Either way, the client needs to be able to launch a local process — which doesn't work for Claude in Chat/Cowork mode, since that runs in a hosted sandbox.

This PR completes the existing `HttpTransport`'s Streamable HTTP support by implementing the `GET` (SSE) endpoint, so clients that support remote/custom MCP connectors can point straight at the site's REST endpoint over HTTPS (with an Application Password), without any local proxy.

- `GET` now validates the session and protocol version exactly like `POST` does, then opens a real `text/event-stream` response with periodic keep-alive comments.
- The stream is held open for a bounded, filterable duration (30s by default) and then closes; compliant SSE/EventSource clients reconnect automatically. This is a deliberate choice for a synchronous PHP-FPM/mod_php request model: an indefinitely-held connection would tie up a worker per client, which the MCP Streamable HTTP spec doesn't require (a server may omit `GET`/SSE support entirely and stay compliant).
- New filters: `mcp_adapter_enable_http_sse_stream` (opt back out to the old `405` behavior), `mcp_adapter_sse_stream_duration`, `mcp_adapter_sse_ping_interval`.
- The actual byte streaming runs through a `rest_pre_serve_request` filter scoped to the specific request, so it only fires during a real `WP_REST_Server` dispatch — calling the transport/handler directly (as the existing unit tests do) never blocks.
- Docs updated with a "connecting directly over HTTP" section and a `curl` example; updated the stale "SSE not yet implemented" notes elsewhere.

Out of scope: OAuth-based authorization for browser-driven connector setups. That's a much larger, security-sensitive feature (essentially an OAuth provider for WordPress) and deserves its own design discussion rather than being bundled into this fix. This PR only completes the transport-level piece; sites can already authenticate remote HTTP access today via Application Passwords.

## Tested on

- [x] `composer test` (full PHPUnit suite via `wp-env`) — all passing.
- [x] `composer lint` / `composer phpstan` — no new warnings or errors.
- [x] Manually verified end-to-end against a real WordPress install in `wp-env`: initialized a session over `POST`, opened the `GET` SSE stream with `curl -N`, confirmed `Content-Type: text/event-stream`, the initial `stream-open` comment, a `ping` keep-alive at the configured interval, and correct `400`/`404` responses for missing/invalid sessions.
- [x] Verified the SSE feature can be disabled via `mcp_adapter_enable_http_sse_stream` to restore the previous `405` behavior.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fplugins.php%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.5%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fmcp-adapter%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-300-50595e45b985c396196219b0ac09b7b0ee7cd4ce-mcp-adapter.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file, per [the Ke
 
 ## [Unreleased] - TBD
 
+### Added
+- The HTTP transport's `GET` endpoint now opens a real SSE stream (per the MCP Streamable HTTP transport) instead of returning `405`, so clients that can only reach a WordPress site over HTTPS — including remote/cloud MCP clients that cannot spawn a local STDIO process — can connect directly using an [Application Password](https://make.wordpress.org/core/2020/11/05/application-passwords-integration-guide/), with no local proxy required. The stream is held open for a bounded duration (30 seconds by default) before closing so a single connection can't tie up a PHP-FPM worker indefinitely; compliant clients reconnect automatically. New filters: `mcp_adapter_enable_http_sse_stream`, `mcp_adapter_sse_stream_duration`, `mcp_adapter_sse_ping_interval`. See the [CLI Usage guide](docs/guides/cli-usage.md#connecting-directly-over-http-remote-clients).
+
 ## [0.6.1] - 2026-08-13
 
 ### Fixed
--- a/docs/guides/cli-usage.md
+++ b/docs/guides/cli-usage.md
@@ -146,6 +146,40 @@ The [`@automattic/mcp-wordpress-remote`](https://www.npmjs.com/package/@automatt
 
 For more information, see the [@automattic/mcp-wordpress-remote](https://www.npmjs.com/package/@automattic/mcp-wordpress-remote) package documentation.
 
+#### Connecting directly over HTTP (remote clients)
+
+Both options above are STDIO from the client's point of view: `wp mcp-adapter serve` runs locally, and so does the `@automattic/mcp-wordpress-remote` proxy — it just forwards to HTTP behind the scenes. Neither works for a client that runs somewhere it can't launch local processes, such as Claude in Chat or Cowork mode, which runs in a hosted sandbox.
+
+For those clients, point them at the site's MCP REST endpoint directly — no local process required:
+
+```
+https://your-site.example/wp-json/mcp/mcp-adapter-default-server
+```
+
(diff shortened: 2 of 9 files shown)


galatanovidiu:
Thanks for the work on this, but I'm going to close it. We won't add the SSE stream to the GET endpoint.

Two reasons:

- MCP 2026-07-28 removes the HTTP GET endpoint and the `Mcp-Session-Id` header (SEP-2567, SEP-2575). Server-to-client notifications move to `subscriptions/listen` over POST. This code would only apply to the legacy revision.
- On 2025-11-25, returning 405 on GET is already spec-compliant. The adapter runs request-per-process, so it never sends server-initiated messages. The stream would only carry pings while holding a PHP worker for 30 seconds, and clients reconnect right after.

On #280: the adapter already speaks Streamable HTTP over POST, so remote clients can connect today. As far as I can tell, the real blocker for Claude Chat/Cowork is authentication, not transport. Their connector UI only offers OAuth. That is a separate discussion, and I'll keep #280 open for it.

PR #310, opened 2026-09-01
Support exact MCP 2025-11-25 and 2026-07-28 revisions

### LINKED ISSUE #268: Add support for protocol version 2026-07-28

galatanovidiu:
## What?

Closes #268.

This adds exact MCP `2025-11-25` and `2026-07-28` support to the Adapter through the revision-selected runtime introduced by [WordPress/php-mcp-schema#15](https://github.com/WordPress/php-mcp-schema/pull/15).

- keep the 2025 initialization, session, notification, and ping lifecycle
- add the 2026 sessionless `server/discover` lifecycle and per-request protocol metadata
- replace the former generated DTO integration with exact-revision schema records
- route HTTP and STDIO requests through one revision-aware wire boundary
- project tools, resources, and prompts only into revisions that can represent them
- add raw-wire, architecture, and projection coverage for both revisions
- document the direct-consumer migration and the intentional protocol differences

This PR does not include the MRTR work built on top of this branch.

## Why?

MCP `2026-07-28` changes both message shapes and lifecycle semantics. The Adapter cannot correctly support it by extending the existing single-revision DTO path or by treating the new revision as a small negotiation alias.

The schema package now owns exact wire shape, validation, hydration, serialization, and message availability. The Adapter remains responsible for negotiation, transport, lifecycle, component registration, execution, and WordPress integration. Keeping that boundary allows both revisions to coexist without duplicating the Adapter or branching ordinary Ability registration by protocol version.

The schema dependency PR is still draft, so this PR is also opened as draft.

## How?

- Select one exact schema per request through `McpSchemaProvider` and carry it in `McpRequestContext`; there is no implicit schema at protocol-facing boundaries.
- Use `McpWireOrchestrator` as the single decode, validation, dispatch, and response-encoding boundary for HTTP and STDIO.
- Keep Ability registration revision-neutral, cache validated component projections per revision, and omit a component only from revisions whose canonical schema cannot represent it.
- Preserve the stateful 2025 HTTP flow while enforcing 2026 body metadata and protocol headers without server-side sessions.
- Retain Adapter-owned semantic checks while removing duplicated DTO-shape validators and the validation-disable switch.
- Reject batches before dispatch.
- Remove non-canonical `tools/list/all`; expose `initialize`, `notifications/initialized`, and `ping` only in 2025, and `server/discover` only in 2026.
- Keep existing Ability declarations, callbacks, Adapter execution filters, and `create_server()` usage unchanged. Direct schema/DTO consumers and custom protocol integrations must follow the included migration guide.
- Pin the schema runtime to the reviewed commit until the dual-revision schema package is released.

### Use of AI Tools

AI assistance: Yes  
Tool(s): Codex  
Used for: implementation and test support, documentation, and pull request description drafting.

## Testing Instructions

The following validation was completed on this branch:

- The GitHub `Test PHP 8.4 WP latest with coverage` job resolved `latest` to WordPress 7.1 and passed; a local WordPress 7.1 Plugin Check rerun reported no errors.
- `npm run test:php` — 481 tests and 1,928 assertions passed; one existing test was skipped.
- `npm run lint:php` — passed.
- `npm run lint:php:stan` — passed with no errors.
- `composer validate --strict` — metadata was valid; Composer emitted the expected warning about the intentional exact commit-ref dependency.
- PHP syntax checks and `git diff --check` — passed.

The PHPUnit suite includes `DualRevisionWireCorpusTest`, which exercises positive and negative raw HTTP and STDIO behavior for both revisions.

To reproduce the main gates:

```bash
npm run wp-env start
npm run test:php
npm run lint:php
npm run lint:php:stan
```

## Changelog Entry

> Changed - Support exact MCP 2025-11-25 and 2026-07-28 revisions through the revision-selected php-mcp-schema runtime.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fplugins.php%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.5%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fmcp-adapter%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-310-8a87020802b1414828a841695cd3c03eeebd827a-mcp-adapter.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


dugyen:
I reproduced the testing instructions locally at `2418a976` (the same commit the Playground-preview artifact link in the PR description points to), on macOS with PHP 8.5.8 / Composer 2.9.2. Docker wasn't available on this machine, so `wp-env` couldn't be used directly — instead I stood up an isolated WordPress 7.1 test environment (fresh WP core download, a throwaway MariaDB instance, `wp-phpunit/wp-phpunit`) to run the real PHPUnit suite outside of Docker, and ran the rest via Composer directly.

**Results — all green:**

| Check | Result |
|---|---|
| `composer validate --strict` | Valid (only the expected warning about the intentional commit-ref dependency on `wordpress/php-mcp-schema`) |
| PHP syntax (`php -l`) across `includes/`, `tests/` | Clean |
| `git diff --check` vs `trunk` | Clean |
| `composer run-script lint` (phpcs) | 16/16, no violations |
| `composer run-script phpstan` | 61 files analysed, no errors |
| `tests/artifact/verify.sh` (plugin-zip smoke) | Passed — sha256-verified zip, 319 files (305 PHP), no dev-only paths (`tests/`, `composer.json`, `vendor/phpunit`, etc.) leaked into the artifact. Smoke output confirms `server/discover` reports both `["2026-07-28","2025-11-25"]` with the expected per-revision method sets. |
| PHPUnit (unit + integration, `--strict-coverage`) | **475 tests, 1782 assertions, 1 skipped**, 0 failures |

One thing worth flagging: the PR description's own CI run recorded **411 tests / 2,188 assertions**, vs. **475 tests / 1,782 assertions** here — same single skip either way, and nothing failed in my run. My `composer install` pulled `wordpress/php-mcp-schema` at `dev-feature/dual-revision-schema-runtime` HEAD as resolved right now, and since that dependency is pinned to a moving branch ref on a still-draft PR ([php-mcp-schema#15](https://github.com/WordPress/php-mcp-schema/pull/15)), the schema-driven data providers in `DualRevisionWireCorpusTest` / `ComponentCompatibilityTest` etc. likely picked up more revision/case combinations than were present when the PR's CI job ran. Worth double-checking the `composer.lock` schema commit is still the one you intend to ship before merge, given it's exact-pinned but the upstream branch keeps moving.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

dugyen:
I did a deeper pass on the diff (`includes/` vs `trunk`, ~8.5k lines across 37 files) looking for correctness bugs and cleanup opportunities, verifying each candidate by reading the actual code/behavior on both sides of the diff rather than taking it at face value. Findings below, most severe first.

## Correctness

**1. `tool_not_found()` / `prompt_not_found()` lost their dedicated error codes** — [`includes/Infrastructure/ErrorHandling/McpErrorFactory.php:107-122`](https://github.com/WordPress/mcp-adapter/blob/2418a976b09d317058e771b1ef1ad2719f1cebdc/includes/Infrastructure/ErrorHandling/McpErrorFactory.php#L107-L122)
`resource_not_found()`, right above them, was correctly made revision-aware (`Schemas::V2026_07_28 === $revision ? INVALID_PARAMS : RESOURCE_NOT_FOUND`). `tool_not_found()` and `prompt_not_found()` weren't — they now unconditionally return `INVALID_PARAMS` (-32602), where trunk used `TOOL_NOT_FOUND`/`PROMPT_NOT_FOUND` (-32003/-32004). Since `McpWireOrchestrator::http_response_status()` only special-cases `INVALID_PARAMS` → HTTP 400 for the 2026-07-28 revision (lines 146-148), a **2025-11-25** client calling `tools/call`/`prompts/get` with an unknown name now gets **HTTP 200** with a JSON-RPC error body instead of the previous 404. Looks like the revision-branch was applied to `resource_not_found` but forgotten on the other two.

**2. Header-name folding breaks `x-mcp-header` params whose name contains `_`** — [`includes/Transport/Infrastructure/HttpRequestContext.php:83`](https://github.com/WordPress/mcp-adapter/blob/2418a976b09d317058e771b1ef1ad2719f1cebdc/includes/Transport/Infrastructure/HttpRequestContext.php#L83)
Incoming header names are normalized via `str_replace('_', '-', strtolower($name))`. But `McpTool::scan_header_annotations()`'s validity regex (`^[!#$%&'*+.^_\`|~0-9A-Za-z-]+$`) explicitly permits `_` in a declared header name. A tool declaring `"x-mcp-header": "my_field"` expects header `Mcp-Param-my_field`; the lookup in `McpWireOrchestrator` (~line 709) tries `mcp-param-my_field` and `mcp_param_my_field`, but the stored key is `mcp-param-my-field` (the `_` inside the name got folded too). Every such header-mapped parameter fails with a spurious `header_mismatch` even when the client sends exactly what the schema asks for.

**3. `fromArray()` no longer fails fast on malformed component data** — `includes/Domain/Tools/McpTool.php` (and the `McpResource`/`McpPrompt` equivalents)
Trunk wrapped `ToolAnnotations::fromArray()`/`ToolDto::fromArray()` in try/catch and returned `WP_Error` synchronously on bad `annotations`/`execution` shapes. That's gone — `fromArray()` now only returns `WP_Error` for a missing `name`/`handler`; malformed `annotations` (e.g. `readOnlyHint` as a string instead of bool) is stored as-is and construction "succeeds". The bad shape is only caught later, lazily, in `McpComponentRegistry::has_any_projection()`, which just logs a `warning` and drops the component from registration. Callers that check `is_wp_error($tool)` per the documented `self|WP_Error` contract get a false negative — the tool just silently vanishes from `tools/list`.

**4. `RegisterAbilityAsMcpResource::make()` / `RegisterAbilityAsMcpPrompt::make()` silently changed return shape** — [`includes/Domain/Resources/RegisterAbilityAsMcpResource.php:78`](https://github.com/WordPress/mcp-adapter/blob/2418a976b09d317058e771b1ef1ad2719f1cebdc/includes/Domain/Resources/RegisterAbilityAsMcpResource.php#L78)
Public static factory, return type changed from a `Resource`/`Prompt` DTO object to a plain array. No in-repo caller is left to catch this (only the sibling `build()` is used internally), and it isn't mentioned in `docs/migration/dual-revision-schema-runtime.md` despite that doc's stated purpose of covering direct-consumer migration. Any external plugin still calling `::make()` and then `->getUri()` on the result will now fatal.

**5. Malformed-notification exceptions are swallowed with zero logging** — [`includes/Transport/Infrastructure/McpWireOrchestrator.php:238-249`](https://github.com/WordPress/mcp-adapter/blob/2418a976b09d317058e771b1ef1ad2719f1cebdc/includes/Transport/Infrastructure/McpWireOrchestrator.php#L238-L249)
`catch (\Throwable $throwable)` around `hydrate_inbound()` for notifications discards `$throwable` entirely — there's no `error_handler->log()` call anywhere in this 872-line file. A non-compliant client's malformed `notifications/initialized` is silently accepted with no server-side trace to diagnose it by.

**6. Deleted validators' `pattern`/`format`/length checks have no schema-driven replacement at all** — the removed `McpToolValidator`/`McpResourceValidator`/`McpPromptValidator` enforced MCP name format, resource URI format, and base64 content validity via regex/format checks. `vendor/wordpress/php-mcp-schema`'s `SchemaInterpreter` implements only `type`/`enum`/`const`/`minimum`/`maximum`/`maxItems`/`required` — grepping it turns up zero handling of `pattern`, `format`, `minLength`, or `maxLength`. This isn't just the old `mcp_adapter_validation_enabled` switch being removed (that's documented as intentional) — the underlying capability to re-express these checks doesn't exist in the new validator at all, even for a future opt-in.

## Cleanup

**7. Not-found/HTTP-status revision quirk duplicated across two files** — `McpErrorFactory::resource_not_found()` ([:100](includes/Infrastructure/ErrorHandling/McpErrorFactory.php#L100)) is the one method in an otherwise revision-neutral factory that branches on `$revision`; `McpWireOrchestrator::http_response_status()` ([:146-148](includes/Transport/Infrastructure/McpWireOrchestrator.php#L146-L148)) carries a second, disconnected special-case for the same `(2026-07-28, INVALID_PARAMS)` pair. One protocol difference, two unlinked ad hoc branches — a third revision would need matching edits in both places with nothing enforcing they stay in sync (see finding #1 for what happens when they drift).

**8. Message decoded and revision-negotiated twice per HTTP request** — `requires_2025_11_25_http_session()` and `process()` (`McpWireOrchestrator.php` ~104/169/185) each independently call `decoder->to_associative()` (a full recursive walk of the JSON-RPC payload, including `params`/`arguments`) and `select_revision()` on the identical message. Hoisting the first call's result into `process()` would halve this per-request cost.

**9. `is_available_for()` triplicated instead of living on the shared trait** — byte-identical in `McpTool`, `McpResource`, and `McpPrompt` (all three already `use RevisionProjectionTrait`): `try { $this->get_protocol_record($schema); } catch (\Throwable) { return false; } return true;`. Could be one default trait method.

**10. Four independent hand-rolled JSON-tree-walkers** — `McpWireOrchestrator::to_object()`/`copy_json_value()`, `McpRequestContext::copy_value()`/`copy_object()`/`copy_array()`, `HandlerHelperTrait::callback_value()`, and `JsonRpcRequestDecoder::to_associative()` all independently convert between `stdClass`/array/list shapes. `McpRequestContext::copy_object()` already duplicates its own sibling `copy_value()`'s stdClass branch rather than calling it — the pattern is already drifting apart within one new file, let alone across four.

---

Happy to help fix any of these if useful — #1 and #2 in particular look like straightforward, low-risk patches.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

galatanovidiu:
Thanks @dugyen for taking the time to verify this. Went through all ten.

Finding 2 is real. WordPress canonicalizes header names to underscores in `WP_REST_Request::canonicalize_header_name()`, `HttpRequestContext` folds them back to hyphens, and the lookup in `McpWireOrchestrator::validate_tool_parameter_headers()` kept the underscore from the annotation name, so the keys never matched. Fixed in dbcf171, with an integration test for an underscore header name.

Finding 9 is done in 0ffc6f2. `is_available_for()` lives on `RevisionProjectionTrait` now.

Finding 4: `make()` on the resource and prompt registrars goes away in #319, which stacks on this PR and documents the replacement (`McpResource::fromAbility()` / `McpPrompt::fromAbility()`, or `build()`) in the migration doc.

The rest I don't plan to change:

- Finding 1 is intentional. The 2025-11-25 spec uses `-32602` for unknown tools and prompts and only gives resources their own code (`-32002`). `-32003` and `-32004` were adapter-private codes. This is in the changelog under "Changed" and in `docs/guides/error-handling.md`.
- Finding 3 is by design. Validity is per revision now, so `fromArray()` can't fail at construction for a shape one schema accepts and the other rejects. The registry logs a warning and omits the component from that revision. The changelog covers it, and #319 continues in that direction.
- Finding 5: a notification gets no response by spec. Logging it would be nice to have, not a correctness issue.
- Finding 6 is about `php-mcp-schema`, not this PR. `format` is an annotation in JSON Schema unless you opt in, so the interpreter is correct not to enforce it.
- Findings 7 and 8 encode two different spec rules and one extra walk over a small payload. Not worth the coupling a shared helper would add.
- Finding 10: the four walkers do three different jobs (deep copy, array to object for hydration, object to array for callbacks). Only the last pair overlaps, and merging it would make the handlers depend on the transport decoder.

On the first comment: the schema dependency is pinned to a commit hash (`#5e08fc16`), so `composer install` can't drift. The moving-branch explanation for the test count doesn't hold.

dugyen:
Tested this against `0ffc6f2` (branch head at time of testing). Ran the full local suite outside `wp-env` (PHP 8.5.7, a disposable WP 7.1 core checkout, MySQL 5.7):

- Unit: 444 tests, 1643 assertions — OK (1 skipped)
- Integration: 44 tests, 350 assertions — OK
- Combined: 488 tests, 1993 assertions — OK
- `composer phpstan`: 0 errors across 61 files
- `composer lint`: clean

Also specifically exercised the underscore/hyphen header-folding fix from `dbcf171` — ran `test_http_2026_matches_underscore_tool_parameter_headers` directly and confirmed the valid/mismatch/missing cases all resolve as expected (200 / 400 HEADER_MISMATCH / 400 HEADER_MISMATCH).

All green, no issues found. 🚀

includes/Transport/Infrastructure/McpWireOrchestrator.php
```diff
@@ -0,0 +1,895 @@
...
+	 */
+	private function validate_tool_parameter_headers( array $params, McpRequestContext $context, array $headers, $id ): ?array {
+		$name = $params['name'] ?? null;
+		if ( ! is_string( $name ) ) {
+			return null;
+		}
+
+		$tool = $this->transport_context->mcp_server->get_mcp_tool( $name );
+		if ( ! $tool || ! $tool->is_available_for( $context->schema() ) ) {
+			return null;
```
gziolo, line 716: This lookup uses the raw tool name, while `ToolsHandler::call_tool()` trims it before execution. A padded name can therefore miss this lookup and still resolve to a tool later. The reported reproduction used `name: " regional-tool "`, a matching base64-encoded `Mcp-Name`, and a present `region` argument: the handler ran without `Mcp-Param-Region`, whereas the unpadded name returned 400. A focused test comparing the two forms would help verify that validation and execution resolve the same tool and enforce the same required headers.

includes/Domain/Tools/McpTool.php
```diff
@@ -414,4 +433,76 @@ public function get_adapter_meta(): array {
...
+	 * @param mixed $node Schema node.
+	 * @param bool $property_schema Whether this node is a property schema.
+	 * @param bool $reachable Whether its path contains only properties keys.
+	 * @param list<string> $path Property path.
+	 * @param array<string, true> $names Case-insensitive header names.
+	 * @param array<int, array{name: string, path: list<string>}> $annotations Valid annotations.
+	 */
+	private function scan_header_annotations( $node, bool $property_schema, bool $reachable, array $path, array &$names, array &$annotations ): void {
+		if ( ! is_array( $node ) ) {
+			return;
```
gziolo, line 467: The scan stops at `stdClass` nodes, and the `properties` branch below also requires an array. These object representations are accepted by schema projection, so it would be useful to check that an annotation advertised by `tools/list` is also enforced during a call. The reported case used `stdClass` for `inputSchema.properties` or a child property schema and executed without the required header. Tests for both representations should verify that a missing header prevents the handler from running.

includes/Handlers/Resources/ResourcesHandler.php
```diff
@@ -86,48 +86,55 @@ public function list_resources(): ListResourcesResult {
...
-		// Extract parameters using helper method.
-		$request_params = $this->extract_params( $params );
-
-		if ( ! isset( $request_params['uri'] ) ) {
-			return McpErrorFactory::missing_parameter( $request_id, 'uri' );
+	public function read_resource( ReadResourceRequest $request, McpRequestContext $request_context ): array {
+		$params         = $request->getParams();
+		$request_id     = $request->getId();
+		$uri            = trim( $params->getUri() );
+		$request_params = array( 'uri' => $params->getUri() );
```
gziolo, line 118: This rebuilds the callback parameters from four known fields, although the schema record retains additional request fields. On trunk, those fields reached permission callbacks, callable handlers, and resource filters. The reported reproduction added `customParam` to `resources/read` and found it missing from callback input on both revisions. It would be useful to verify that the conversion preserves the full parameter set expected by existing integrations, with coverage for permission checks, execution, and filters.

includes/Transport/Infrastructure/McpWireOrchestrator.php
```diff
@@ -0,0 +1,895 @@
...
+	/** @param mixed $body_value Body value. Compare a decoded primitive header value to the body value. */
+	private function header_value_matches( string $header_value, $body_value ): bool {
+		if ( is_bool( $body_value ) ) {
+			return ( $body_value ? 'true' : 'false' ) === $header_value;
+		}
+		if ( is_int( $body_value ) ) {
+			if ( $body_value > 9007199254740991 || $body_value < -9007199254740991 ) {
+				return false;
+			}
+			return is_numeric( $header_value ) && (float) $header_value === (float) $body_value;
```
gziolo, line 789: A useful edge case here is an integer-schema parameter encoded as `42.0` with header `"42"`. The schema runtime accepts the value as an integer, but PHP decodes it as a float, so it misses this branch. The reported result was 400, while body value `42` succeeded with the same header. A wire test that preserves the `42.0` representation would help verify that header comparison follows the schema’s numeric semantics while retaining the existing range checks.

includes/Transport/Infrastructure/McpWireOrchestrator.php
```diff
@@ -0,0 +1,895 @@
...
+
+		$id           = $generic['id'] ?? null;
+		$safe_id      = is_string( $id ) || is_int( $id ) ? $id : null;
+		$method       = isset( $generic['method'] ) && is_string( $generic['method'] ) ? $generic['method'] : null;
+		$notification = null !== $method && ! array_key_exists( 'id', $generic );
+		if ( '2.0' !== ( $generic['jsonrpc'] ?? null ) || null === $method ) {
+			return $this->failure( McpErrorFactory::invalid_request( $safe_id, 'Invalid JSON-RPC request envelope' ), $method, $safe_id, $notification );
+		}
+		if ( ! $notification && ! is_string( $id ) && ! is_int( $id ) ) {
+			return $this->failure( McpErrorFactory::invalid_request( null, 'Request id must be a string or integer' ), $method, null, false );
```
gziolo, line 182: A raw request containing `id: 1.0` is worth checking here. The schema runtime accepts it as an integer, but PHP decodes it as a float, so this check rejects it with `-32600` and `id: null`. The mismatch was reported for both 2026 HTTP and 2025 STDIO requests. Raw-JSON tests on both transports would help verify that this validation and the safe-ID handling agree with the schema’s integer semantics.

includes/Transport/Infrastructure/JsonRpcRequestDecoder.php
```diff
@@ -0,0 +1,187 @@
...
+
+		try {
+			$value = json_decode( $json, false, self::MAX_DEPTH, JSON_THROW_ON_ERROR );
+		} catch ( \JsonException $exception ) {
+			// phpcs:ignore WordPress.Security.EscapeOutput.ExceptionNotEscaped -- Exception text is protocol validation data, not HTML output.
+			throw new \InvalidArgumentException( 'Invalid JSON: ' . $exception->getMessage(), 0, $exception );
+		}
+
+		if ( ! $value instanceof \stdClass ) {
+			throw new \UnexpectedValueException( 'The JSON-RPC payload must be one JSON object; batches are not supported.' );
```
gziolo, line 39: This unconditional rejection needs checking against the legacy versions accepted during initialization. `McpVersionNegotiator::LEGACY_PROTOCOL_VERSIONS` includes `2025-03-26`, whose [specification requires receiving JSON-RPC batches](https://modelcontextprotocol.io/specification/2025-03-26/basic#batching), but batches are rejected here before revision dispatch. A test using a session negotiated to that revision would establish the compatibility gap and help determine whether batching needs support or the advertised and negotiated version set needs narrowing.

gziolo (commented):
Thank you for this work. This review was assisted by Codex and Claude Code. The following summarizes the remaining findings and recommendations reported during those reviews, cross-checked against `aeb8d68`.

The overall assessment supports the design and its alignment with WordPress/php-mcp-schema#15. The six reported correctness findings are attached inline, with recommendations for fixes and regression tests.

Additional compatibility and documentation recommendations:

- List-shaped results: 2025 responses now drop `structuredContent` for results such as `['a', 'b', 'c']`, while 2026 retains it. This is spec-correct but changes trunk's behavior. Document the change and test empty and nonempty lists.
- Deferred factory validation: malformed annotations or schemas no longer necessarily produce an immediate `WP_Error`. Explain the changed `is_wp_error()` contract and replacement checks in the migration guide.
- Missing protocol headers: rejection is documented and tested for `2025-06-18`; add an explicit `2025-11-25` case.
- Supported versions: discovery lists two versions, while unsupported-version errors list five. Align these with the supported negotiation behavior.

The coverage review also recommends retaining dedicated assertions for prompt/resource mapping and name filters, transformed tool input/output handling, session update retry exhaustion, image-block `_meta`, empty capability-object serialization, and ability callback exception codes.

docs/migration/dual-revision-schema-runtime.md
```diff
@@ -0,0 +1,152 @@
...
+    Tool::class,
+    array(
+        'name'        => 'weather',
+        'inputSchema' => array( 'type' => 'object' ),
+    )
+);
+```
+
+Replace complete serialization methods with `jsonSerialize()` or direct JSON
+encoding. Records also provide named getters, `get()`, and `has()`.
```
gziolo: It would help to mention that `jsonSerialize()` returns a `stdClass`, rather than an array. An integration that replaces `toArray()` with `jsonSerialize()` but keeps `$data['name']` access will fail. A short example showing object access and reconstruction through the selected schema would make the migration clearer, especially for filters that modify serialized records.

docs/migration/dual-revision-schema-runtime.md
```diff
@@ -0,0 +1,152 @@
...
+
+Protocol-facing Adapter internals now receive validated records and an exact
+request context:
+
+| Surface                              | Current contract                                                                                                            |
+| ------------------------------------ | --------------------------------------------------------------------------------------------------------------------------- |
+| Custom HTTP transports               | Delegate raw `WP_REST_Request` objects through `HttpRequestHandler` and `HttpRequestContext`.                               |
+| Other custom transports              | Decode and process raw JSON through `McpWireOrchestrator`.                                                                  |
+| `RequestRouter::route_request()`     | Accepts a generated request `Record`, `McpRequestContext`, and transport name. Do not pass raw method and parameter arrays. |
+| Method handlers                      | Accept the exact generated request record and `McpRequestContext`; return logical arrays for final schema projection.       |
```
gziolo: The initialization exception would be useful to mention here. `InitializeHandler::handle()` returns an `InitializeResult` record, and the `mcp_adapter_initialize_response` filter must return that record type too. The current wording suggests every handler returns an array, which could lead someone migrating an initialization filter to return the wrong type.

includes/Transport/Infrastructure/McpWireOrchestrator.php
galatanovidiu: Nice catch. Fixed in a8d209d: `ToolsHandler::call_tool()` and `ResourcesHandler::read_resource()` now use the name and URI exactly as sent, same as the orchestrator lookup and both SDKs. A padded name is not found on either revision, so it can no longer skip the `Mcp-Param-*` check. Test: `test_http_padded_names_and_uris_are_not_trimmed` in `DualRevisionWireCorpusTest`. It fails on the old code with the tool executing, and passes now with tool-not-found.

includes/Domain/Tools/McpTool.php
galatanovidiu: Fixed in bbbb6e6. `scan_header_annotations()` converts a `stdClass` node to an array at entry and before the `properties` branch, so both representations declare the same mirrors. Input schemas are still stored as given. Tests: `test_header_annotations_are_found_in_stdclass_schema_nodes` (properties object and child object, with the path) and `test_http_2026_validates_parameter_headers_declared_in_stdclass_schema_nodes` (missing header returns 400 and the handler does not run).

includes/Handlers/Resources/ResourcesHandler.php
galatanovidiu: I'd keep the four protocol keys here (`uri`, `_meta`, `inputResponses`, `requestState`). It is input whitelisting rather than a lost feature: ability-backed resources get no arguments anyway, and I could not find a consumer that reads extra keys. It is a change from trunk, so abf20d4 adds it to the migration guide and changelog, and `test_http_resource_read_forwards_only_protocol_parameters` checks that an unknown key reaches neither the permission callback, `mcp_adapter_pre_resource_read`, nor the handler on both revisions. If you know an integration that depends on extra keys, I'm happy to revisit.

includes/Transport/Infrastructure/McpWireOrchestrator.php
galatanovidiu: Fixed in a865b6a. Each header annotation now carries the declared type, and `header_value_matches()` branches on it like the TS SDK: an `integer` declaration accepts an int or float body, gates the header on `^-?\d+(\.\d+)?$`, and compares numerically. The safe-integer range check stays. Other declarations compare the body's decimal string. Test: `test_http_2026_compares_integer_parameter_headers_numerically`, raw JSON so `42.0` survives. It covers `42.0` vs `42`, `42` vs `42.0`, rejects ` 42 `, `+42`, `4.2e1`, and checks the string-declared case still mismatches on `42.0`.

includes/Transport/Infrastructure/McpWireOrchestrator.php
galatanovidiu: Fixed in 3f9d87b. I put the normalization in `JsonRpcRequestDecoder::decode()` rather than in `process()`: an integral float id becomes an int there, so both transports and the HTTP session preflight errors echo `1` instead of rejecting or returning `id: null`. `1.5` and out-of-range values are still rejected. Test: `test_http_and_stdio_accept_integral_float_request_ids`, raw JSON on HTTP 2026 and STDIO 2025, plus a 2025 request without a session to check the preflight id.

includes/Transport/Infrastructure/JsonRpcRequestDecoder.php
galatanovidiu: Fair point. I went with narrowing the version set (da186f9): `2025-03-26` is no longer in `LEGACY_PROTOCOL_VERSIONS`, a proposal for it gets `2025-11-25` as before, and `server/discover` and the unsupported-version error both list `SUPPORTED_PROTOCOL_VERSIONS`. Trunk never advertised `2025-03-26` either, so nothing that worked before stops working. `test_http_unknown_initialize_proposal_receives_2025_11_25` now covers the `2025-03-26` proposal, and the unsupported-version error asserts the same list as discovery.

galatanovidiu:
All four notes are addressed on the branch:

- List-shaped results: documented in the migration guide and changelog (ee1e71c). `test_http_list_shaped_structured_content_is_revision_specific` covers an empty and a non-empty list on both revisions.
- Deferred factory validation: the migration guide now explains what `fromArray()` still returns as `WP_Error` and how schema problems surface through `is_available_for()` and `get_projection_error()` (f5131e7).
- Missing protocol header: `test_http_2025_11_25_session_requires_protocol_version_header` (ee1e71c).
- Supported versions: discovery and the unsupported-version error share one list (da186f9, see the batches thread).

Coverage backfill is in eea9477: name and URI filters, schema wrappers on execute and permission, session update retry exhaustion, image-block `_meta`, capability objects on the wire, and ability callback exceptions on both revisions. One thing I learned there: for ability-backed tools the error code is core's `ability_callback_exception`, because `WP_Ability` catches before the adapter does.

dugyen:
Re-tested at `f5131e7`.

**Local (WP 7.0.2, PHP 8.5.7, fresh core download since Docker wasn't available here):** full suite green — 502 tests, 2,156 assertions, 1 intentional skip, 0 failures. Individually re-ran `DualRevisionWireCorpusTest` (48 tests / 439 assertions) and `ComponentCompatibilityTest` (11 tests / 120 assertions), including the two tests named in the comment above (`test_http_list_shaped_structured_content_is_revision_specific`, `test_http_2025_11_25_session_requires_protocol_version_header`) — all pass.

**CI at the same commit tells a different story for WP 6.9:** all 7 "WP 6.9" matrix jobs fail (PHP 7.4 through 8.5), while every WP 7.0/latest/trunk cell passes. It's one new assertion from the `eea9477` backfill:

```
ComponentCompatibilityTest::test_ability_tool_exceptions_and_metadata_remain_available
tests/phpunit/Unit/ComponentCompatibilityTest.php:551
Failed asserting that two strings are identical.
-'ability_callback_exception'
+'mcp_execution_failed'
```

The test assumes `WP_Ability::execute()` catches the callback exception and tags it `ability_callback_exception` before the adapter's own catch runs — true on WP 7.0's Abilities API (which is why every non-6.9 cell is green), but not on WP 6.9's older Abilities API, where the adapter's own handler runs and returns `mcp_execution_failed` instead. Same split hits the `permission_error` assertion one line below.

Three options, least to most churn: assert on the error message instead of the exact code (or branch the expected code); skip the exact-code assertions on WP < 7.0 the way the suite already skips the multisite meta-key test; or drop WP 6.9 from the supported matrix if the adapter now targets the newer Abilities API only.

Everything else is clean: PHPCS, static analysis, Plugin Check, and the plugin-zip build are all green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

dugyen:
@gziolo the new commits since your last pass (`ee1e71c`, `da186f9`, `eea9477`, `f5131e7`) look ready for another review — could you take another look when you have a chance? I don't have write access on this repo to re-request you directly. (Re-test notes above.)

README.md
```diff
@@ -10,11 +10,11 @@ The official WordPress package for MCP integration that exposes WordPress abilit
 
 ## Overview
 
-MCP Adapter bridges WordPress's [Abilities API](https://developer.wordpress.org/news/2025/11/introducing-the-wordpress-abilities-api/) with the [MCP specification](https://modelcontextprotocol.io/specification/2025-11-25/), giving AI agents a standardized way to interact with WordPress functionality. It includes HTTP and STDIO transport support, comprehensive error handling, and an extensible architecture for custom integrations.
+MCP Adapter bridges WordPress's [Abilities API](https://developer.wordpress.org/news/2025/11/introducing-the-wordpress-abilities-api/) with exact MCP [`2025-11-25`](https://modelcontextprotocol.io/specification/2025-11-25/) and [`2026-07-28`](https://modelcontextprotocol.io/specification/2026-07-28/) revisions (legacy `2025-06-18`, `2025-03-26`, and `2024-11-05` clients are served through the `2025-11-25` schema), giving AI agents a standardized way to interact with WordPress functionality. It includes HTTP and STDIO transport support, comprehensive error handling, and an extensible architecture for custom integrations.
```
justlevine: This reads like a decision log, and we want to not need to update this readme every time the version we ship changes. Recommend reverting, but changing the URL to autolink to the latest.


```suggestion
MCP Adapter bridges WordPress's [Abilities API](https://developer.wordpress.org/news/2025/11/introducing-the-wordpress-abilities-api/) with the [MCP specification](https://modelcontextprotocol.io/specification/), giving AI agents a standardized way to interact with WordPress functionality. It includes HTTP and STDIO transport support, comprehensive error handling, and an extensible architecture for custom integrations.

```

README.md
```diff
@@ -36,6 +36,7 @@ Connect via WP-CLI over STDIO, or point an HTTP client at `/wp-json/mcp/mcp-adap
 
 ## Migration
 
+- [Dual-revision schema runtime migration](docs/migration/dual-revision-schema-runtime.md)
```
justlevine: The docs themselves should be folded in as well.

```suggestion
- [Migration Guide: vx.y.z](docs/migration/vx.y.z.md)
```

tests/phpunit/Unit/ComponentCompatibilityTest.php
```diff
@@ -0,0 +1,555 @@
...
+	public function handle( array $arguments ): array {
+		return $arguments;
+	}
+}
+
+/** Protects component behavior that remains independent of schema representation. */
+final class ComponentCompatibilityTest extends TestCase {
+
+	public function set_up(): void {
+		parent::set_up();
```
justlevine, line 61: IMO - using the yoast polyfills from the start prevents us from dealing with edge cases later.

```suggestion
	public function setUp(): void {
		parent::setUp();
```

docs/migration/dual-revision-schema-runtime.md
justlevine, line 1: Per https://github.com/WordPress/mcp-adapter/pull/310#discussion_r4018434380 I would merge this into vx.y.x.md instead of sending folks to migrate using two different docs.

docs/README.md
```diff
@@ -25,6 +25,7 @@ Documentation for the WordPress MCP Adapter - transform WordPress abilities into
 ## Migration guides
 
 - **[Migrating to vx.y.z](migration/vx.y.z.md)**
+- **[Dual-revision schema runtime](migration/dual-revision-schema-runtime.md)**
```
justlevine: https://github.com/WordPress/mcp-adapter/pull/310#discussion_r4018434380
```suggestion
- **[Migrating to vx.y.z](migration/vx.y.z.md)**
```

docs/architecture/overview.md
justlevine, line 1: This is a big departure from the previous file, @galatanovidiu wanted to confirm you reviewed it?

docs/architecture/overview.md
```diff
@@ -1,481 +1,169 @@
...
+npm run lint:php
+npm run lint:php:stan
+npm run plugin-zip
+```
+
+The PHPUnit corpus includes conforming raw HTTP and STDIO lifecycle, tool,
+resource, and prompt positives for both revisions; cross-revision method
+negatives; header, metadata, numeric, and unsupported-version failures;
+ordinary Ability execution; hook preservation; isolated component projections;
+and forbidden-symbol scans.
```
justlevine: This I'm assuming is doc-slop (doesn't belong _here_, and already exists in CONTRIBUTING.md)
```suggestion
```

docs/guides/error-handling.md
```diff
@@ -1,240 +1,85 @@
...
-    // Validation failed, $validation_result contains error response
-    return new WP_REST_Response($validation_result, 400);
-}
-// Validation passed
-```
\ No newline at end of file
+`JsonRpcRequestDecoder` performs one identity-preserving JSON decode and rejects
+malformed JSON, batches, excessive depth, integers outside PHP's native range,
+and non-finite numbers. `McpWireOrchestrator` then checks method availability,
+revision metadata, transport headers, schema hydration, and response encoding.
```
justlevine: As this is AI generated, I didn't read the contents, but I did catch the line-breaks. This is a doc file there's no 80ch limit (and those limits hurt tokenization).

```suggestion

Request IDs may be strings, integers, or `null`. Protocol-facing code must pass the selected revision to `resource_not_found()` because MCP 2025 uses `-32002` and MCP 2026 uses `-32602`. Missing tools and prompts use standard Invalid Params (`-32602`) in both supported revisions.

## Handler boundary

Handlers receive a validated generated request record and `WP\MCP\Core\McpRequestContext`. They return logical result data or an error array. The selected schema constructs the final result and JSON-RPC response records.

Custom HTTP transports should delegate to `HttpRequestHandler`. Other transports should call `McpWireOrchestrator::decode()` and `process()` before serializing the returned record. Do not route unvalidated method and parameter arrays directly.

## HTTP status

The built-in HTTP path uses the selected revision when mapping protocol errors. For example, Invalid Params remains a JSON-RPC response with HTTP 200 in the 2025 revision and uses HTTP 400 in the 2026 revision. Custom HTTP transports that delegate to `HttpRequestHandler` inherit the same behavior.

## JSON-RPC validation

`JsonRpcRequestDecoder` performs one identity-preserving JSON decode and rejects malformed JSON, batches, excessive depth, integers outside PHP's native range, and non-finite numbers. `McpWireOrchestrator` then checks method availability, revision metadata, transport headers, schema hydration, and response encoding.
```

docs/guides/custom-transports.md
```diff
@@ -1,186 +1,121 @@
...
 
-This guide covers how to implement custom transport layers for the MCP Adapter when the built-in `HttpTransport` doesn't meet your specific needs.
+Use a custom transport when the built-in HTTP and WP-CLI STDIO transports do
+not fit the connection you need. For authentication or authorization changes on
+the built-in HTTP endpoint, use [transport permissions](transport-permissions.md)
+instead.
 
-## Built-in Transports
+Custom transports implement `McpTransportInterface`. REST transports implement
+`McpRestTransportInterface`:
```
justlevine:
```suggestion
Use a custom transport when the built-in HTTP and WP-CLI STDIO transports do not fit the connection you need. For authentication or authorization changes on the built-in HTTP endpoint, use [transport permissions](transport-permissions.md) instead.

Custom transports implement `McpTransportInterface`. REST transports implement `McpRestTransportInterface`:

```

docs/guides/custom-transports.md
```diff
@@ -1,186 +1,121 @@
...
 }
 ```
 
-## Creating Custom Transports
+## Custom REST transport
 
-### Basic Example: API Key Transport
+Delegate MCP processing to `HttpRequestHandler`. It owns the exact 2025 session
+lifecycle, the sessionless 2026 lifecycle, revision selection,
+header validation, schema hydration, response encoding, and HTTP status.
```
justlevine:
```suggestion

Delegate MCP processing to `HttpRequestHandler`. It owns the exact 2025 session lifecycle, the sessionless 2026 lifecycle, revision selection, header validation, schema hydration, response encoding, and HTTP status.
```

docs/guides/custom-transports.md
```diff
@@ -1,186 +1,121 @@
...
-See [Transport Permissions](transport-permissions.md) for simpler authentication solutions.
+## Non-HTTP transports
 
-### Use Custom Transports For:
-- ✅ **Custom routing patterns**
-- ✅ **Message queue integration**
-- ✅ **Request signing/encryption**
-- ✅ **Specialized protocols**
+Use `McpWireOrchestrator` for a message queue, socket, or another transport that
+already supplies raw JSON:
```
justlevine, line 86:
```suggestion

Use `McpWireOrchestrator` for a message queue, socket, or another transport that already supplies raw JSON:

```

docs/guides/custom-transports.md
```diff
@@ -1,186 +1,121 @@
...
 
-## Next Steps
+The transport owns authentication, delivery, and connection state. For a 2025
+connection, retain the successful initialize parameters and pass them as the
+fourth argument on later requests. A 2026 request carries its revision and
+client capabilities in `_meta`, so no connection-scoped MCP session is used.
+
+Do not invoke `RequestRouter` with raw arrays. The orchestrator must select and
+validate the exact request record before dispatch.
+
```
justlevine, line 105:
```suggestion

The transport owns authentication, delivery, and connection state. For a 2025 connection, retain the successful initialize parameters and pass them as the fourth argument on later requests. A 2026 request carries its revision and client capabilities in `_meta`, so no connection-scoped MCP session is used.

Do not invoke `RequestRouter` with raw arrays. The orchestrator must select and validate the exact request record before dispatch.

```

includes/Cli/McpCommand.php
```diff
@@ -132,18 +133,43 @@ public function list( array $args, array $assoc_args ): void {
...
+				'Tools'          => $server->count_tools(),
+				'Resources'      => $server->count_resources(),
+				'Prompts'        => $server->count_prompts(),
+				'2025 Tools'     => count( $server->get_tools( $schema_2025 ) ),
+				'2025 Resources' => count( $server->get_resources( $schema_2025 ) ),
+				'2025 Prompts'   => count( $server->get_prompts( $schema_2025 ) ),
+				'2026 Tools'     => count( $server->get_tools( $schema_2026 ) ),
+				'2026 Resources' => count( $server->get_resources( $schema_2026 ) ),
+				'2026 Prompts'   => count( $server->get_prompts( $schema_2026 ) ),
+				'Description'    => $server->get_server_description(),
```
justlevine: IMO this feels like the wrong shape. If there's a reason why we need the specific breakdown per-schema prompts (do we?) then they should be fed via a CLI argument (--supports-protocol=2025-11-25,2026-07-28), and keyed to the specific protocol number, not a year.

includes/Cli/StdioServerBridge.php
justlevine, line 1: Looks like AI destroyed all the doc blocks in this file.

includes/Core/McpServer.php
```diff
@@ -94,12 +96,8 @@ class McpServer {
 	 */
 	private McpTransportFactory $transport_factory;
 
-	/**
-	 * Whether MCP validation is enabled.
-	 *
-	 * @var bool
-	 */
-	private bool $mcp_validation_enabled;
+	/** @var \WP\McpSchema\Schemas */
```
justlevine: Let's give this a full doc-block

includes/Core/McpAdapter.php
```diff
@@ -191,9 +191,9 @@ private function register_wp_cli_commands(): void {
 	 * @param array<class-string<\WP\MCP\Transport\Contracts\McpTransportInterface>> $mcp_transports Array of MCP transport class names to initialize.
 	 * @param class-string<\WP\MCP\Infrastructure\ErrorHandling\Contracts\McpErrorHandlerInterface>|null $error_handler The error handler class name. If null, NullMcpErrorHandler will be used.
 	 * @param class-string<\WP\MCP\Infrastructure\Observability\Contracts\McpObservabilityHandlerInterface>|null $observability_handler The observability handler class name. If null, NullMcpObservabilityHandler will be used.
-	 * @param list<string> $tools Ability names to register as tools.
-	 * @param list<string> $resources Resources to register.
-	 * @param list<string> $prompts Prompts to register.
+	 * @param list<string|\WP\MCP\Domain\Tools\McpTool> $tools Ability names or MCP tools to register.
+	 * @param list<string|\WP\MCP\Domain\Resources\McpResource> $resources Ability names or MCP resources to register.
+	 * @param list<string|\WP\MCP\Domain\Prompts\McpPrompt|\WP\MCP\Domain\Prompts\Contracts\McpPromptBuilderInterface> $prompts Ability names, MCP prompts, or prompt builders to register.
```
justlevine, line 196: I understand a `McpComponentInterface` technically _can_ be passed down without throwing an error, but does it _need to_? Adding it to the public params means it'l be a breaking change if we ever want to iterate here, so if there isn't a concrete end user case for it, i'd rather we leave it off until we're dogfooding it ourselves or have a case for it.

includes/Core/McpServer.php
```diff
@@ -121,9 +119,9 @@ class McpServer {
...
-	 * @param list<string> $tools Optional ability names to register as tools during construction.
-	 * @param list<string> $resources Optional resources to register during construction.
-	 * @param list<string> $prompts Optional prompts to register during construction.
+	 * @param list<string|\WP\MCP\Domain\Tools\McpTool> $tools Optional ability names or MCP tools to register during construction.
+	 * @param list<string|\WP\MCP\Domain\Resources\McpResource> $resources Optional ability names or MCP resources to register during construction.
+	 * @param list<string|\WP\MCP\Domain\Prompts\McpPrompt|\WP\MCP\Domain\Prompts\Contracts\McpPromptBuilderInterface> $prompts Optional ability names, MCP prompts, or prompt builders to register during construction.
 	 * @param callable|null $transport_permission_callback Optional custom permission callback for transport-level authentication. If null, defaults to is_user_logged_in().
```
justlevine, line 137: https://github.com/WordPress/mcp-adapter/pull/310#discussion_r4020432961

includes/Domain/Resources/RegisterAbilityAsMcpResource.php
```diff
@@ -13,13 +13,12 @@
...
 use WP\MCP\Domain\Utils\McpValidator;
 use WP\MCP\Infrastructure\ErrorHandling\Contracts\McpErrorHandlerInterface;
-use WP\McpSchema\Server\Resources\DTO\Resource as ResourceDto;
 use WP_Error;
 
 /**
  * Converts WordPress abilities to MCP Resource metadata.
  *
- * This class builds Resource DTOs for resources/list responses.
+	 * This class builds neutral Resource metadata for revision projection.
```
justlevine:
```suggestion
	* This class builds neutral Resource metadata for revision projection.
```

includes/Domain/Utils/ContentBlockHelper.php
```diff
@@ -235,20 +120,29 @@ public static function json_text( $data, int $flags = 0, ?Annotations $annotatio
...
-		return array_map(
-			static function ( ContentBlockInterface $block ): array {
-				return $block->toArray();
-			},
-			$blocks
+		return array_values( $blocks );
+	}
+
+	/** Build the embedded wrapper. */
+	private static function embedded_resource( array $resource_data, ?array $annotations, ?array $_meta ): array {
```
justlevine: Give this a real doc-block please.

includes/Domain/Utils/ContentBlockHelper.php
```diff
@@ -235,20 +120,29 @@ public static function json_text( $data, int $flags = 0, ?Annotations $annotatio
...
+				'type'        => 'resource',
+				'resource'    => $resource_data,
+				'annotations' => $annotations,
+				'_meta'       => McpValidator::normalize_meta( $_meta ),
+			)
 		);
 	}
+
+	/** Remove optional null fields while preserving false, zero, and empty strings. */
+	private static function without_nulls( array $data ): array {
```
justlevine: Give this a real doc-block please

includes/Handlers/Prompts/PromptsHandler.php
```diff
@@ -9,155 +9,102 @@
...
-			 * or return a WP_Error to block execution and return an error to the client.
+			 * Filters prompt arguments before execution.
 			 *
 			 * @since 0.5.0
-			 *
-			 * @param array                              $arguments   The prompt arguments.
-			 * @param string                             $prompt_name The prompt name being retrieved.
-			 * @param \WP\MCP\Domain\Prompts\McpPrompt   $mcp_prompt  The MCP prompt instance.
-			 * @param \WP\MCP\Core\McpServer             $server      The MCP server instance.
 			 */
```
justlevine, line 135: Please restore the filter params.

includes/Handlers/Prompts/PromptsHandler.php
```diff
@@ -167,19 +114,9 @@ public function get_prompt( array $params, $request_id = 0 ) {
...
-			 * content enrichment, or audit logging.
-			 *
 			 * @since 0.5.0
-			 *
-			 * @param mixed|\WP_Error                    $result      The raw execution result (may be WP_Error).
-			 * @param array                              $arguments   The prompt arguments used.
-			 * @param string                             $prompt_name The prompt name.
-			 * @param \WP\MCP\Domain\Prompts\McpPrompt   $mcp_prompt  The MCP prompt instance.
-			 * @param \WP\MCP\Core\McpServer             $server      The MCP server instance.
 			 */
```
justlevine, line 153: Please restore the filter params

includes/Infrastructure/ErrorHandling/McpErrorFactory.php
```diff
@@ -9,472 +9,194 @@
...
-	 * @param int $code The error code.
-	 * @param string $message The error message.
-	 * @param mixed|null $data Optional additional error data.
-	 *
-	 * @return \WP\McpSchema\Common\JsonRpc\DTO\JSONRPCErrorResponse
-	 */
-	public static function create_error_response( $id, int $code, string $message, $data = null ): JSONRPCErrorResponse {
-		return JSONRPCErrorResponse::fromArray(
-			array(
-				'jsonrpc' => McpConstants::JSONRPC_VERSION,
```
justlevine, line 77: So I understand why stripping out `McpConstants` from the library, but would a `WordPress/Mcp/Core/Constants` be better than hardcoding this numbers everywhere?

includes/Transport/Infrastructure/HttpRequestContext.php
```diff
@@ -70,6 +73,23 @@ public function __construct( \WP_REST_Request $request ) {
...
+			}
+
+			$value = is_array( $values ) ? reset( $values ) : $values;
+			if ( ! is_string( $value ) ) {
+				continue;
+			}
+
+			$this->headers[ $key ] = $value;
+		}
 	}
```
justlevine: Can we move this into a private function and keep this single-responsibility?

```suggestion
		$this->headers          =  $this->prepare_headers( $request->get_headers() );
```

justlevine (changes requested):
Mainly code quality and doc nits

README.md
galatanovidiu: Done

README.md
galatanovidiu: Folded the schema-runtime migration into `vx.y.z.md` and updated the links. Also adjusted the introduction so direct API consumers know the migration applies even when using the standalone plugin. Done in c3e0fa2.

tests/phpunit/Unit/ComponentCompatibilityTest.php
galatanovidiu: This already uses the Yoast polyfill through `WP_UnitTestCase`. Its inherited `setUp()` calls `set_up()`, so I think we should keep the existing method here.

docs/architecture/overview.md
galatanovidiu: Good call asking for another review, a lot has changed in these docs. I’ve reviewed the overview against the code and applied corrections in 58b6027. Next, I’ll review the other documentation files, both touched and untouched by this PR, for accuracy and consistency.

docs/architecture/overview.md
galatanovidiu: Removed the duplicated verification section and linked to CONTRIBUTING and the testing guide instead. Done in 58b6027.

docs/guides/error-handling.md
galatanovidiu: Removed the hard wrapping here and checked all Markdown files in the repo for the same issue. Cleaned up eight files while preserving code blocks, intentional line breaks, and rendered content. Done in 877ec12.

docs/guides/custom-transports.md
galatanovidiu: Applied the paragraph unwrapping throughout this guide in 877ec12.

includes/Cli/StdioServerBridge.php
galatanovidiu: Good catch. Restored the docblocks here in 9470c0f. I also reviewed the other PHP files changed in this PR for missing or incomplete docblocks and repaired them. This restores parameter, return, exception, and hook documentation without changing runtime behavior.

includes/Core/McpServer.php
galatanovidiu: Good catch. Expanded this into a full docblock explaining the schema provider, its contents, and how to access a revision, with a forVersion() reference. Done in 9470c0f as part of the wider docblock review.

includes/Domain/Resources/RegisterAbilityAsMcpResource.php
galatanovidiu: Good catch. Fixed the indentation in 9470c0f. I also reviewed the other changed PHP files and repaired missing or incomplete docblocks.

includes/Domain/Utils/ContentBlockHelper.php
galatanovidiu: Good catch. Added the parameter and return documentation for the embedded resource wrapper in 9470c0f. I also repaired the other incomplete docblocks in this helper and the other changed PHP files.

includes/Domain/Utils/ContentBlockHelper.php
galatanovidiu: Good catch. Documented the input and return types, preserved keys, and null-only filtering in 9470c0f. This was included in the wider docblock review.

includes/Handlers/Prompts/PromptsHandler.php
galatanovidiu: Good catch. Restored all four filter parameter descriptions and the WP_Error short-circuit behavior in 9470c0f. The runtime arguments are unchanged.

includes/Handlers/Prompts/PromptsHandler.php
galatanovidiu: Good catch. Restored all five filter parameter descriptions, including the raw result or WP_Error, in 9470c0f. I checked the other changed files for the same problem too.

includes/Infrastructure/ErrorHandling/McpErrorFactory.php
galatanovidiu: The error codes are already centralized as constants on `McpErrorFactory`, and callers use those names. I think keeping them alongside error construction is clearer than adding a general `Core\Constants` class.

includes/Transport/Infrastructure/HttpRequestContext.php
galatanovidiu: Done in a7b0a76. Extracted the header preparation into a private method, preserving the existing filtering and normalization.

includes/Core/McpAdapter.php
galatanovidiu: To clarify, would you prefer we remove the concrete component types from these parameter docblocks, or remove support for registering them directly? The registry already accepted `McpTool`, `McpResource`, and `McpPrompt` instances before this PR, so I want to distinguish documenting existing behavior from changing the supported API.

includes/Transport/Infrastructure/HttpRequestHandler.php
```diff
@@ -11,342 +11,154 @@
...
-	private function validate_protocol_version_header( HttpRequestContext $context ): ?array {
-		if ( null === $context->protocol_version ) {
-			return null;
-		}
-
-		if ( McpVersionNegotiator::is_supported( $context->protocol_version ) ) {
-			return null;
+		if ( null === $processed['response'] ) {
+			return new \WP_REST_Response( null, 202 );
 		}
```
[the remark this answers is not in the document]
galatanovidiu: Fixed in 21e2bc9. Accepted notifications still return an empty 202. Rejected notifications now return an HTTP error with no body, including the early session-validation paths. This also avoids the schema exception caused by trying to build a response with no request ID. Added regression tests for these cases and preserved STDIO silence.

includes/Core/McpAdapter.php
justlevine: Personally, I'd prefer to _leave them off the doc-block_ so we can decide in the future if we need to support direct class instances in the future or if it's something else we can cleanup. But I defer to you.

justlevine (approved):
No other blocking notes, looks good 🙌

includes/Core/McpAdapter.php
galatanovidiu: I’d keep them in the docblock. The MRTR guide explicitly shows creating an `McpTool` and passing it when registering a server, so the parameter documentation should reflect that supported usage.

I understand the concern about committing to these classes, but leaving them undocumented would not prevent integrations from depending on them. If we decide to remove that support later, we would still need to handle the compatibility impact.

Issue #270, opened 2026-08-02
HttpTransport: MCP endpoint registered as GET-cacheable without cache-control defense; risks full-page-cache poisoning under LSC/W3TC/WP Rocket


raftaar1191:
## Summary

`HttpTransport::register_routes()` registers the MCP endpoint with `methods => array( 'POST', 'GET', 'DELETE' )` but does not emit `Cache-Control` headers or `define( 'DONOTCACHEPAGE', true )`. Under LiteSpeed Cache, WP Rocket, W3 Total Cache, WP Super Cache, host-level FastCGI cache, or a similar full-page cache with defaults, this means the GET method can be cached and served to a subsequent request from a different user.

Today GET returns HTTP 405 (SSE not yet implemented, per the class-level comment), so the immediate impact is limited — but full-page caches _do_ cache 4xx responses in many configurations, and the moment SSE lands (or any GET handler is added) the exposure grows immediately. Fixing at the transport layer now is cheaper than hunting per-handler later.

## Reproduction

1. Install any WordPress site with the `wordpress/mcp-adapter` package active and LiteSpeed Cache 7.8.x enabled with default settings.
2. Issue an unauthenticated `GET /wp-json/{namespace}/{route}` (whatever the MCP endpoint resolves to).
3. Inspect `X-LiteSpeed-Cache` header on subsequent requests to the same URL — the cached 405 is served without invoking PHP.

LSC's cache-decision code (`src/control.cls.php` `_setting_cacheable()` L833-918) treats GET as cacheable by default unless one of these signals is present:

- `define( 'DONOTCACHEPAGE', true )` — checked at L638-641; the only response-level signal LSC honors from the callee.
- URL in admin-configured `cache-exc` regex list.

`Cache-Control: no-store` response headers ALONE are not consulted by LSC's decision — verified against LSC 7.8.1 source.

## Impact

- **Today (GET returns 405)**: minor. A stale 405 served to a subsequent request would break an SSE-attempting client but the response body carries no data.
- **Once SSE (or any other GET handler) lands**: severe. A GET response body containing per-session tool listings, event data, or session tokens would be cached and served to unrelated requests — data-leak class bug.

Also applies to POST responses that carry per-user tool-call results. LSC skips POST by default, but not all full-page caches do (some FastCGI configurations cache POST responses keyed by URL alone).

## Suggested fix

Add cache-defense headers at the transport layer, so every response emitted through `HttpTransport::handle_request()` is safe by default:

```php
[...]

PR #314, opened 2026-09-07
fix: set cache-defense headers on MCP HTTP endpoint responses

sarthak-19:
## Summary
- `HttpTransport::register_routes()` exposes POST/GET/DELETE on the MCP endpoint, but `handle_request()` never emitted `Cache-Control`/`Pragma` headers or defined `DONOTCACHEPAGE`. Under a full-page cache with default settings (LiteSpeed, WP Rocket, W3 Total Cache, WP Super Cache, host-level FastCGI cache), a response could be cached and served to a different user on a subsequent request.
- `handle_request()` now defines `DONOTCACHEPAGE` (guarded with `defined()`) and sets `Cache-Control: no-store, no-cache, must-revalidate, private` plus `Pragma: no-cache` on every response, applied uniformly at the transport layer so POST, GET, and DELETE are all covered without relying on individual handlers to opt in.

## Test plan
- [x] Added `test_post_response_has_cache_defense_headers`, `test_get_response_has_cache_defense_headers`, and `test_delete_response_has_cache_defense_headers` in `tests/phpunit/Integration/HttpTransportTest.php`, asserting the headers/constant on each method.
- [x] `composer lint` (phpcs) — clean, no errors.
- [x] `composer phpstan` — no errors.
- [x] `npm run test:php -- --filter HttpTransportTest` — 29 tests, 115 assertions, all passing.
- [x] `npm run test:php` (full suite) — 1037 tests, 3959 assertions, 1 pre-existing skip, no failures.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fplugins.php%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.5%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fmcp-adapter%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-314-3abbf70925e7afa91b2fd6ab1c21dbc899e4d5fa-mcp-adapter.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ All notable changes to this project will be documented in this file, per [the Ke
 ### Changed
 - Usage of MCP Adapter as a bundled library has been deprecated in favor of using the canonical MCP Adapter plugin. See the [vx.y.z migration guide](migration/vx.y.z.md) for instructions on how to migrate away from a bundled copy of MCP Adapter.
 
+### Fixed
+- The MCP HTTP endpoint now sends `Cache-Control: no-store, no-cache, must-revalidate, private` and `Pragma: no-cache` headers, and defines `DONOTCACHEPAGE`, on every response. Previously, full-page caches (LiteSpeed, WP Rocket, W3 Total Cache, WP Super Cache, host-level FastCGI cache) could cache and replay a response to a different user, since the endpoint carried no cache-defense signal.
+
 ## [0.6.1] - 2026-08-13
 
 ### Fixed
--- a/includes/Transport/HttpTransport.php
+++ b/includes/Transport/HttpTransport.php
@@ -147,8 +147,18 @@ class HttpTransport implements McpRestTransportInterface {
 	 * @return \WP_REST_Response
 	 */
 	public function handle_request( \WP_REST_Request $request ): \WP_REST_Response {
-		$context = new HttpRequestContext( $request );
+		// Prevent full-page caches (LiteSpeed, WP Rocket, W3TC, WP Super Cache, etc.)
+		// from caching and replaying MCP responses to other users.
+		if ( ! defined( 'DONOTCACHEPAGE' ) ) {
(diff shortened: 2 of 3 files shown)


galatanovidiu:
I don't think there's a bug here, so I'm closing this. I checked it on a live wp-env (WP 7.1) with curl, on trunk and with this patch applied.

- Core already sends `Cache-Control: no-cache, must-revalidate, max-age=0, no-store, private` and `Expires` on every request that reaches `HttpTransport::handle_request()`. `WP_REST_Server::serve_request()` does that for logged-in users, and `check_permission()` only lets authenticated users through.
- Unauthenticated requests get a 401 from the permission callback before `handle_request()` runs, so this patch does not touch the GET scenario from #270.
- The `Cache-Control` set here never reaches the wire. Core sends its no-cache headers after `$result->get_headers()`, and `header()` replaces same-name headers. The only new bytes are `Pragma: no-cache` and `DONOTCACHEPAGE`. The tests skip `WP_REST_Server`, so they can't see this.
- In LiteSpeed 7.9.1, `O_CACHE_REST` defaults to `false` and non-GET requests are always no-cache, so "default settings serve a cached 405" from #270 does not hold. The SSE argument is gone too, the 2026-07-28 revision removes the GET stream (#300).

The changelog claims a response could be replayed to another user. I can't find that leak, and I don't want a security fix in the release notes for a "might happen". If this is a real problem, I need a reproduction: the plugin and version, the request sent, and the response headers showing the cached reply. That would reopen this.

PR #319, opened 2026-09-09
Let the schema reject malformed component data instead of repairing it

galatanovidiu:
## What?

Follow up to #310.

Removes the adapter-side repair and type-filter layer in front of the `php-mcp-schema` projection. Component and handler fields reach the schema without the removed repair and type filters, and the schema decides whether they fit the selected revision. The changes also add rejection diagnostics, regression coverage, and migration notes.

## Why?

The dual-revision runtime already projects every component and handler result through the schema package for the negotiated revision. The older repair code in front of it silently changed data before the schema saw it: it coerced annotation values, dropped malformed `_meta` and `mimeType`, rewrote prompt roles and content types, injected `type: object` into input schemas, trimmed labels, and rejected values the protocol accepts (a 2048-byte URI cap, a bare `scheme:` URI, an empty resource name). A component that failed the schema was often repaired into something that passed, or dropped without a visible error. Two layers deciding what is valid also meant they could disagree, and one such disagreement made a tool unavailable in both revisions while passing the adapter check.

The official TypeScript SDK server stores definitions as given and validates only `tools/call`; its client strips unknown keys on receipt. Filtering adapter-side enforced rules the protocol does not have.

## How?

**Fail-fast.** The schema package is the contract. A component that fails projection in every supported revision is rejected at registration. Each failed revision is logged through the error handler; a completely rejected component also raises one `_doing_it_wrong` notice containing the component identifier, revisions, and schema JSON-pointer messages. Components valid in one revision remain registered and available in that revision. A handler result that fails final projection returns `-32603 Internal error: Invalid handler result`.

**Three adapter checks stay** because the schema package cannot express them and the adapter depends on them: the tool and prompt name rule (registry lookup key), the resource URI scheme (scheme folding), and `lastModified` as an ISO 8601 timestamp with a time zone. The timestamp rule is the one documented exception to "rules the protocol does not have are out of scope": the official TypeScript SDK client rejects a whole `resources/list` result when one member's `lastModified` lacks a zone. The check mirrors the SDK's `z.iso.datetime({ offset: true })`, accepts any fractional-second precision, anchors with `\z`, and parses the calendar date in UTC. It rejects the resource rather than dropping the annotation.

**Carried as given** on every registration path: icons, `_meta`, `mimeType`, `size`, `annotations`, `outputSchema`, `execution`, explicit `mcp.arguments` entries, label and description, resource `name`, and a set `inputSchema`. An explicit `null` counts as absent, as everywhere else in the adapter. Content-block helpers take `_meta` and `mimeType` untyped so a decoded JSON object reaches the schema uncast; casting turns numeric-string keys into a list the schema rejects.

**Modern result metadata.** Validate result `_meta` through the selected schema's `ResultMetaObject` before adding `io.modelcontextprotocol/serverInfo`. This rejects lists and scalars instead of turning them into valid output, and preserves valid `stdClass` fields, including numeric object keys, without mutating provider metadata. PHP arrays use `fromArray()` and other values use `fromValue()` so schema validation retains their original shape. Explicit `null` remains absent. Supplied `serverInfo` must be schema-valid; the response uses the Adapter server's name and version.

**Annotations.** Top-level `meta.annotations` is the location WordPress core defines and fills with null defaults, so it is read without a deprecation notice; `mcp.annotations` overrides it, and an empty `mcp.annotations` array now means "no annotations" instead of falling back. The mapper renames keys and drops nulls only.

**Prompt results.** The shape conveniences (`messages`, `text`, `role` with `content`, `texts`, JSON fallback) stay, chosen by key presence. The value under the key is carried as given, so `messages => 'bad'` reaches the schema instead of the JSON fallback. Message entries keep every key. An empty `messages` list is emitted as given; the `(No messages returned)` placeholder is gone. A JSON-encode failure in the fallback throws to the handler boundary.

**Errors that changed.** Added `mcp_prompt_invalid_arguments` (non-array `mcp.arguments`), `mcp_ability_invalid_meta` (non-array `meta.mcp`, matching the fail-closed rule in `McpAbilityExposure`; before this an object value threw out of the `McpServer` constructor during `mcp_adapter_init`), `mcp_resource_invalid_annotations` and `resource_annotations_invalid` (bad `lastModified`), `mcp_resource_name_filter_invalid` (non-string filter result). Removed `mcp_prompt_invalid_argument`, `mcp_prompt_argument_missing_name`, `mcp_resource_missing_name`. `mcp_resource_missing_uri` now covers a missing or null key only; a wrong-typed URI is `mcp_resource_invalid_uri` / `resource_uri_invalid`.

**Structure.** A private `ThrowableGuardTrait::guard()` replaces fifteen try/catch-to-`WP_Error` blocks with the same codes, messages, and `error_type` context. `RegisterAbilityAsMcpResource` and `RegisterAbilityAsMcpPrompt` lose `make()` and their pass-through getters; `build()` is the single entry point, and both are `@internal` like the tool converter. `McpAbilityMeta::mcp()` is the one reader of `meta.mcp`. Dead helpers are deleted: `ContentBlockHelper::error_text`, `json_text`, `to_array_list`, `audio`, `McpValidator::validate_base64`.

**Deliberately unchanged.** `_meta` key-name validation is prose-only in the spec and no SDK enforces it. The 2026-07-28 removal of `execution` is a revision projection rule, not repair. `meta.mcp` never reaches a list response; only the keys the adapter maps to schema fields do.

**Migration and regression coverage.** The changelog and migration guide document the removed methods and error codes, annotation overrides, and reversal of the 0.6.0 "malformed `_meta` is omitted" behavior. Modern result metadata is validated before server identity is added. Regression tests cover rejection diagnostics, valid peers and partial revision availability, malformed metadata, prompt shapes and arguments, labels, resource names, timestamps, and input-schema defaults. Twelve additional metadata datasets cover valid arrays and objects, numeric object keys, null-as-absent, malformed values, server identity, and provider immutability. The existing malformed prompt test now checks result metadata in both revisions. Three malformed prompt fixtures opt out of default discovery and are registered explicitly in rejection coverage.

**External follow-up:** the WPCOM isolated test that calls `RegisterAbilityAsMcpResource::make()` must switch to `build()['resource_data']`; that integration test is outside this repository and has not been changed here.

## Testing Instructions

Run on the test environment:

```bash
npx wp-env --config=.wp-env.test.json run cli --env-cwd=wp-content/plugins/mcp-adapter/ vendor/bin/phpunit -c phpunit.xml.dist
npx wp-env --config=.wp-env.test.json run cli --env-cwd=wp-content/plugins/mcp-adapter/ composer run-script lint
npx wp-env --config=.wp-env.test.json run cli --env-cwd=wp-content/plugins/mcp-adapter/ composer run-script phpstan
```

Validated for `0ab937d`: PHPUnit 440 tests, 2065 assertions, 1 pre-existing skip, 0 failures; PHPCS clean; PHPStan level 8 no errors. `git diff --check` and the new documentation link/anchor checks also passed. No coverage driver was available.

Behaviour checks run against the test environment with throwaway PHPUnit probes, then deleted:

- `tools/list`, `resources/list`, `prompts/list` for abilities whose `meta.mcp` carried unknown keys emitted only schema fields.
- `meta.mcp => new stdClass()` returned `mcp_ability_invalid_meta` from all three converters instead of throwing.
- `inputSchema => ['properties' => []]` and `inputSchema => 'bad'` in `McpTool::fromArray` were rejected at projection with the schema's JSON-pointer message instead of being repaired or throwing a `TypeError`.
- A prompt result of `['messages' => 'bad']` was emitted as given and rejected at the wire, instead of becoming valid text content.
- A boolean property schema produced a name-only prompt argument; an empty argument title passed.

Converter output equivalence for the structural commit: 17 inputs, old versus new `build()` JSON identical.

## Changelog Entry

> Changed - Remove component and handler field repair, coercion, and type filtering. Components rejected in every revision now raise `_doing_it_wrong` with schema paths alongside error-handler logs. Modern result `_meta` is schema-validated before server identity is added, rejecting malformed values and preserving valid object fields.
> Removed - `RegisterAbilityAsMcpResource::make()`, `RegisterAbilityAsMcpPrompt::make()`, `ContentBlockHelper::error_text()`, `json_text()`, `to_array_list()`, `audio()`, `McpValidator::validate_base64()`; error codes `mcp_prompt_invalid_argument`, `mcp_prompt_argument_missing_name`, `mcp_resource_missing_name`.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fplugins.php%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.5%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fmcp-adapter%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-319-2233fb0b7e67dc77b9c9c0c64178fc78a4de5b36-mcp-adapter.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Domain/Utils/ContentBlockHelper.php
```diff
@@ -12,131 +12,137 @@
 /**
  * Builds neutral arrays that are validated with the selected result schema.
  *
+ * Every field is carried as given. The schema package decides whether the
+ * block fits the selected revision. `mimeType` and `_meta` values are therefore
+ * not typed here: a decoded JSON object arrives as stdClass and must reach the
+ * schema unchanged, because casting it to an array turns numeric-string keys
+ * into a list.
+ *
```
justlevine: This reads like AI Slop (decision log)
```suggestion
 * Builds neutral arrays that are validated with the selected result schema.
 *
```

includes/Domain/Utils/McpAbilityMeta.php
```diff
@@ -0,0 +1,55 @@
...
+ * component data; the adapter reads it and never emits it. Core validates only
+ * `meta` itself as an array, so the converters check the block here before they
+ * index it. A block that is set and not an array is rejected, matching the
+ * fail-closed rule in McpAbilityExposure.
+ *
+ * @internal
+ *
+ * @since n.e.x.t
+ */
+final class McpAbilityMeta {
```
justlevine, line 27: I'm wondering if this is the better pattern versus an `AbstractRegisterAbilityAsMcpComponent` that has this helper but also DRYs out some of the other stuff. NBD because you're marking this `@internal`, but still thought it was worth bringing up.

justlevine (commented):
Functionally lgtm, not entirely sold on the whole `::guard()` helper, but def not against it 🚀

includes/Handlers/Tools/ToolsHandler.php
```diff
@@ -347,13 +346,23 @@ public function call_tool( CallToolRequest $request, McpRequestContext $request_
...
 	 */
-	private function content_metadata( $meta ): ?array {
-		return McpValidator::normalize_meta( $meta instanceof \stdClass ? (array) $meta : $meta );
+	private function content_metadata( $meta ) {
+		if ( null === $meta || is_array( $meta ) || $meta instanceof \stdClass ) {
+			return $meta;
+		}
+
+		throw new \UnexpectedValueException( 'Tool result _meta must be a JSON object.' );
 	}
```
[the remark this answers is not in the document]
galatanovidiu: Updated this in e537611. Content `_meta` now reaches the selected PHP MCP schema without the separate handler check. Invalid output consistently returns `-32603: Internal error: The server produced an invalid result.`

Request observability now includes final response validation, so an invalid response produces one failed `mcp.request` event instead of being recorded as successful. The configured error logger receives a correlated diagnostic with the schema pointer and request/component context. Request argument values and result payloads are not attached.

Custom router behavior is preserved, including discovery bypassing `route_request()` overrides. I also updated the migration, error-handling, and observability docs, with examples enabling both logging handlers.

For E2E validation, I tested the changes through an isolated Abilities Catalog installation on top of #316:

- **408 requests in the final run:** 204 each on WordPress 6.9 and 7.1, using MCP `2025-11-25` and `2026-07-28` over HTTP and persistent STDIO.
- Seven metadata positions covering image and embedded text/blob resources, with valid objects, null, scalars, and lists. Checked valid field preservation, numeric object keys, null omission, and malformed-result errors.
- Ordinary tool execution errors and the catalog’s consent/MRTR request-and-retry flow.
- **224 intentional validation failures:** each produced the expected client error, exactly one failed request event, and one diagnostic. With the built-in error-log handlers enabled, I also checked the actual PHP logs for correlation, schema pointers, and absence of the controlled private-result value.

The broader catalog suite passed on WordPress 7.1, including search, curated and default servers, core reads, resources, prompts, HTTP validation, consent flows, and mixed-revision STDIO. This used a disposable-runner adjustment for the Origin behavior already changed in #310.

PHPUnit passes with 480 tests and 4,212 assertions, with one existing skip. PHPStan passes. PHPCS reports no errors and one existing CLI inheritance warning.

includes/Domain/Utils/ContentBlockHelper.php
galatanovidiu: Makes sense. I removed the decision log and kept the one-line class summary.

includes/Domain/Utils/McpAbilityMeta.php
galatanovidiu: Fair point. I think the small helper is the better fit here. It centralizes the one thing these converters share: reading and validating `meta.mcp`. Tools, resources, and prompts are different MCP entities with different construction and validation paths, so an abstract base would couple unrelated converters without giving us a useful shared abstraction.

galatanovidiu:
Fair point. I kept `guard()` because it replaces fifteen identical `try/catch` blocks across tool, resource, and prompt callbacks while preserving the same `WP_Error` code, message, and `error_type`. Since it is a private helper used only at callback boundaries, I think the small abstraction is worth it here. If you’d still prefer the explicit `try/catch` blocks, I’m happy to remove the helper and put them back.

justlevine (approved):
> Since it is a private helper used only at callback boundaries, I think the small abstraction is worth it here. 

Good enough for me. Happy to defer to you 🙇

PR #327, opened 2026-09-12
fix: pass three arguments to mcp_adapter_validation_enabled at every call site

### LINKED ISSUE #305: mcp_adapter_validation_enabled is applied with 1 argument in 7 places and 3 in McpServer
### What happens

`mcp_adapter_validation_enabled` is applied with a different number of arguments depending on where it fires. `McpServer` passes three:

```php
// includes/Core/McpServer.php:168
$this->mcp_validation_enabled = apply_filters( 'mcp_adapter_validation_enabled', false, $this->server_id, $this );
```

Seven other call sites pass one:

```
includes/Domain/Tools/RegisterAbilityAsMcpTool.php:84
includes/Domain/Tools/McpTool.php:192
includes/Domain/Resources/RegisterAbilityAsMcpResource.php:465
includes/Domain/Resources/McpResource.php:199
includes/Domain/Prompts/RegisterAbilityAsMcpPrompt.php:513
includes/Domain/Prompts/McpPrompt.php:206
includes/Domain/Prompts/McpPrompt.php:276
```

An integrator who registers a callback against the documented three-argument signature — `add_filter( 'mcp_adapter_validation_enabled', $cb, 10, 3 )` with three required parameters — gets the three arguments from `McpServer` and one from everywhere else. The one-argument invocations then raise `ArgumentCountError` during ability-to-tool conversion, which can leave the server without its tools.

Verified against the `mcp-adapter.zip` asset of the v0.6.1 release.

### Suggested fix

Either pass `$server_id` and the server (or `null` where no server is in scope) at every call site, or have the domain classes read the server's already-filtered `mcp_validation_enabled` value instead of re-applying the filter themselves. The second reads better from here — it makes the filter fire once per server rather than once per converted object, and removes the possibility of the two disagreeing.

### How this was found

While reviewing a plugin that bundles the adapter. We are not patching our vendored copy — keeping it byte-identical to the release is what makes "which adapter version is actually running" answerable — so reporting it instead.

Happy to open a PR if the maintainers would prefer one; wanted to check which of the two shapes you'd want first.

cpruijsen:
## What?
Closes #305

Pass the documented extra arguments (`$server_id`, `$server`) at every `mcp_adapter_validation_enabled` application. Domain factories that have no server in scope pass `null, null`.

## Why?
`McpServer` applies this filter with three arguments. Seven domain factories applied it with one. An integrator who registers a three-argument callback (`add_filter( 'mcp_adapter_validation_enabled', $cb, 10, 3 )`) gets `ArgumentCountError` during ability-to-tool conversion, and the server can come up without its tools.

Verified against the v0.6.1 release zip by the reporter, and still reproduced on current trunk: `RegisterAbilityAsMcpTool::build()` passed one argument into a three-parameter callback while constructing a server that registers `test/always-allowed`.

The extra arguments were added on `McpServer` in ee97837 so multi-server hooks could tell servers apart, and were never mirrored in the domain factories.

## How?
Every remaining `apply_filters( 'mcp_adapter_validation_enabled', false )` now passes `null, null` for the documented extra args. `McpServer` is unchanged at the call site (it already passed `$this->server_id, $this`); its hook docs now state that those extra args are null when no server is in scope.

Alternative considered: have domain factories read `McpServer::is_mcp_validation_enabled()` and stop re-applying the filter (the reporter's preferred shape). Domain factories are public and used without a server (`McpTool::fromArray()` and the rest). Importing `McpServer` into Domain would reverse the existing Core to Domain direction.

A callback that enables validation only for a given `$server_id` still will not see that id during conversion.

A follow-up can pass an optional `?bool $mcp_validation_enabled` from `McpComponentRegistry` into the factories if you would rather the filter fire once per server on the registration path.

## Testing Instructions
1. `add_filter( 'mcp_adapter_validation_enabled', function( $enabled, $server_id, $server ) { return false; }, 10, 3 );`
2. Create an `McpServer` that registers at least one ability as a tool (and optionally a resource and a prompt).
3. Construction should complete and the components should be registered. Previously this raised `ArgumentCountError`.
4. `McpTool::fromArray()`, `McpResource::fromArray()`, `McpPrompt::fromArray()`, and `McpPrompt::fromBuilder()` with the same callback should also succeed.

Covered by `WordPressFiltersTest::test_validation_filter_three_argument_callback_does_not_break_component_conversion`.

## Changelog Entry

> Fixed - `mcp_adapter_validation_enabled` is applied with three arguments at every call site, so three-argument callbacks no longer raise `ArgumentCountError` during ability conversion.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22login%22%3Atrue%2C%22landingPage%22%3A%22%2Fwp-admin%2Fplugins.php%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.5%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fmcp-adapter%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-327-b116400bccd228061eb771d11f6f16f458bce705-mcp-adapter.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Core/McpServer.php
+++ b/includes/Core/McpServer.php
@@ -161,9 +161,9 @@ class McpServer {
 		 *
 		 * @since 0.3.0
 		 *
-		 * @param bool      $enabled   Whether validation is enabled. Default false.
-		 * @param string    $server_id The server ID being configured.
-		 * @param \WP\MCP\Core\McpServer $server    The McpServer instance being constructed.
+		 * @param bool                       $enabled   Whether validation is enabled. Default false.
+		 * @param string|null               $server_id The server ID being configured, or null when no server is in scope.
+		 * @param \WP\MCP\Core\McpServer|null $server    The McpServer instance, or null when no server is in scope.
 		 */
 		$this->mcp_validation_enabled = apply_filters( 'mcp_adapter_validation_enabled', false, $this->server_id, $this );
 
--- a/includes/Domain/Prompts/McpPrompt.php
+++ b/includes/Domain/Prompts/McpPrompt.php
@@ -203,7 +203,7 @@ final class McpPrompt implements McpComponentInterface {
 		}
 
 		// Optional deep validation if enabled.
-		$mcp_validation_enabled = apply_filters( 'mcp_adapter_validation_enabled', false );
+		$mcp_validation_enabled = apply_filters( 'mcp_adapter_validation_enabled', false, null, null );
 		if ( $mcp_validation_enabled ) {
 			$validation_result = McpPromptValidator::validate_prompt_dto( $prompt );
 			if ( is_wp_error( $validation_result ) ) {
@@ -273,7 +273,7 @@ final class McpPrompt implements McpComponentInterface {
 		}
 
 		// Optional deep validation if enabled.
-		$mcp_validation_enabled = apply_filters( 'mcp_adapter_validation_enabled', false );
+		$mcp_validation_enabled = apply_filters( 'mcp_adapter_validation_enabled', false, null, null );
 		if ( $mcp_validation_enabled ) {
 			$validation_result = McpPromptValidator::validate_prompt_dto( $prompt );
 			if ( is_wp_error( $validation_result ) ) {
--- a/includes/Domain/Prompts/RegisterAbilityAsMcpPrompt.php
+++ b/includes/Domain/Prompts/RegisterAbilityAsMcpPrompt.php
@@ -510,7 +510,7 @@ class RegisterAbilityAsMcpPrompt {
 		}
 
 		// Optional deep validation if enabled.
-		$mcp_validation_enabled = apply_filters( 'mcp_adapter_validation_enabled', false );
+		$mcp_validation_enabled = apply_filters( 'mcp_adapter_validation_enabled', false, null, null );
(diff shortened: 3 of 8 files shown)


cpruijsen:
Closing this: #310 removed the `mcp_adapter_validation_enabled` filter outright, so the argument-count mismatch this patched no longer exists.

That change dropped the filter along with `McpServer::is_mcp_validation_enabled`, the validator classes and the optional deep-validation branch at every call site. Wire validation now always runs through the selected schema, per the CHANGELOG entry.

#305 still describes the old filter, so it looks stale too, though that is a maintainer's call.

